PR #2, opened 2015-07-16
Add multithreading support for libbacktrace

sujayakar:
Our previous usage of libbacktrace wasn't thread-safe since it passed `threaded = 0` to `backtrace_create_state`, which then requires all library calls to be serialized.  It's pretty easy to get a segfault by calling into the `backtrace` library across multiple threads:

```rust
#![feature(drain)]
extern crate backtrace;

use std::str;
use std::thread;

fn iter() {
    for _ in (0..16) {
        backtrace::trace(&mut |frame| {
            backtrace::resolve(frame.ip(), &mut |symbol| {
                symbol.name().and_then(|s| str::from_utf8(s).ok()).map(|name| {
                    let mut demangled = String::new();
                    backtrace::demangle(&mut demangled, name).unwrap();
                    println!("{}", demangled);
                });
            });
            true
        });
    }
}

fn main() {
    let mut threads: Vec<_> = (0..16)
        .map(|_| thread::spawn(iter))
        .collect();

    for t in threads.drain(..) {
        t.join().unwrap()
    }
}
```

This patch adds a `StaticMutex` fallback for when we detect that libbacktrace isn't thread-safe.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -68,6 +68,7 @@
 #![doc(html_root_url = "http://alexcrichton.com/backtrace-rs")]
 #![deny(missing_docs)]
 #![cfg_attr(test, deny(warnings))]
+#![feature(static_mutex)]
 
 extern crate libc;
 extern crate debug_builders;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -68,7 +68,6 @@
 #![doc(html_root_url = "http://alexcrichton.com/backtrace-rs")]
 #![deny(missing_docs)]
 #![cfg_attr(test, deny(warnings))]
-#![feature(static_mutex)]
 
 extern crate libc;
 extern crate debug_builders;
--- a/src/symbolize/libbacktrace.rs
+++ b/src/symbolize/libbacktrace.rs
@@ -18,15 +18,25 @@ use std::env;
 use std::ffi::CStr;
 use std::path::Path;
 use std::ptr;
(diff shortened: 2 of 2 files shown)


sujayakar:
hmm, @alexcrichton what's the recommended set of feature flags if we want this to build on stable, release, & nightly?

alexcrichton:
Oh wow this totally slipped my mind! Manually synchronizing has the downside of possibly deadlocking if it's called recursively, but I feel like that's fine

alexcrichton:
Er oops, commented too soon.

Anyway, I feel like adding a mutex is fine. I think it's fine to just always request the not-thread-safe state and don't test for support built in (e.g. always grab the lock). Other than that to build on stable it basically needs to avoid `StaticMutex`, and I've done that in the past with a static `*mut Mutex<()>` initialized to `0 as *mut _` which is then dynamically initialized in a `Once` with a `Box<Mutex<()>>`. This leaks the `Box` but that's pretty minor so it probably doesn't matter too much.

Thanks for this! I can get around to these changes later tonight as well if you don't beat me to them :)

alexcrichton:
Oh wow I totally forgot, but to acquire a lock this lib [actually already has the support](https://github.com/alexcrichton/backtrace-rs/blob/master/src/lib.rs#L104-L137), so you should just need:

```rust
let _guard = ::lock::lock();
```

at the top of the function. It'd also be great to have a test for this, but I can add that later as well :)

PR #12, opened 2016-04-27
Add impl of Default for Backtrace::new

vi:
There is even https://github.com/Manishearth/rust-clippy/issues/705 for such things


alexcrichton:
Thanks!

PR #14, opened 2016-07-17
Adjust Backtrace's Debug impl to match libstd's

sfackler:
Differences from libstd:

* No cut off after 100 frames
* `{:4}` instead of <code>  {:2}</code> to display the frame number because of above.


alexcrichton:
Thanks!

PR #17, opened 2016-10-06
Mutable API for Backtrace

jchlapinski:
Hi,
I would like to be able to skip the first two captured frames, as those are rather uninformative, and possibly also the last, which in my case always gets displayed as `0x0 - <unknown>`.

In order to do so without re-implementing Backtrace struct, there needs to be a public API to either get a mutable reference for the frames vector, like:
```
impl Backtrace { 
  //...
  pub fn frames_mut(&mut self) -> &mut Vec<BacktraceFrame> {
    &mut self.frames
  }
}
``` 
or at least a custom constructor to produce Backtrace with frames, like: 
```
impl Backtrace {
  // ...
  pub fn with_frames(frames: Vec<BacktraceFrame>) -> Backtrace {
    Backtrace { frames: frames }
  }
}
```

In this PR I have prepared both variants, please consider adding at least one of them.

Regards,
Jakub

### DIFF
--- a/src/capture.rs
+++ b/src/capture.rs
@@ -2,6 +2,7 @@ use std::fmt;
 use std::mem;
 use std::os::raw::c_void;
 use std::path::{Path, PathBuf};
+use std::convert::{From, Into};
 
 use {trace, resolve, Frame, Symbol, SymbolName};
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! I'd personally prefer to avoid the `&mut Vec<BacktraceFrame>` but perhaps the `with_frames` could be used (perhaps named `new_frames`?)

jchlapinski:
@alexcrichton I will prepare PR with just the constructor then, as for the name I think `with_frames` is more by the official [convention](https://github.com/rust-lang/rfcs/blob/master/text/0430-finalizing-naming-conventions.md#general-naming-conventions) - it is with_more_detailes variant IMO. It could also be considered a conversion constructor, in which case the name would be `from_frames` but most of those conversion constructors I have seen seem to rather make a copy of the data when converting...

It is all the same to me how the constructor would be called, however I've always found conventions to be important to adhere to. Let me know which name do You prefer, and I'll prep the PR accordingly

Regards,
Jakub

alexcrichton:
Actually yeah how about:

```rust
impl From<Vec<BacktraceFrame>> for Backtrace {
    // ...
}
```

jchlapinski:
Ok, I have reworked this as @alexcrichton suggested.

alexcrichton:
Thanks!

PR #27, opened 2017-02-12
Attempt to discover a cross-compile toolchain version of ar, just like objcopy

jnicholls:
This pattern may warrant something hoisted into the gcc crate, in my opinion.  This is becoming very common-place.

### DIFF
--- a/backtrace-sys/build.rs
+++ b/backtrace-sys/build.rs
@@ -37,6 +37,7 @@ fn main() {
 
     let cfg = gcc::Config::new();
     let compiler = cfg.get_compiler();
+    let cc = compiler.path().file_name().unwrap().to_str().unwrap();
     let mut flags = OsString::new();
     for (i, flag) in compiler.args().iter().enumerate() {
         if i > 0 {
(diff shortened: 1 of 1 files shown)


backtrace-sys/build.rs
```diff
@@ -72,11 +84,10 @@ fn main() {
     let tmpdir = dst.join("__tmp");
     drop(fs::remove_dir_all(&tmpdir));
     t!(fs::create_dir_all(&tmpdir));
-    run(Command::new("ar").arg("x").arg(&lib).current_dir(&tmpdir), "ar");
+    run(Command::new(&ar).arg("x").arg(&lib).current_dir(&tmpdir), "ar");
```
alexcrichton: Could you change "ar" here at the end to &ar as well?

backtrace-sys/build.rs
```diff
@@ -97,7 +108,7 @@ fn main() {
         objs.push(obj.path());
     }
 
-    run(Command::new("ar").arg("crus").arg(&lib).args(&objs), "ar");
+    run(Command::new(&ar).arg("crus").arg(&lib).args(&objs), "ar");
```
alexcrichton: (same as above)

alexcrichton:
Looks great, thanks!

jnicholls:
Updated the commit per your request.  Thanks.

PR #28, opened 2017-02-14
Try harder to find cross-compile build tools

Keruspe:
On exherbo, eveything including the native host is cross-compiled and "unprefixed" toolchain is banned by the package manager, so ar and objcopy are not available.
Furthermore, we allow people to chose their system compiler and thus we don't set e.g. CC=x86_64-pc-linux-gnu-gcc but CC=x86_64-pc-linux-gnu-cc
cc is a pretty standard name for the system C compiler, thus I think this would be a good idea to check using "cc" too, and not just "gcc"


Keruspe:
This currently breaks several packages such as cross and cargo-shell, so if this gets in a release would be greatly appreciated.

alexcrichton:
Thanks!

PR #29, opened 2017-02-16
Use $OBJCOPY environment var if present

zzeroo:
Quick n dirty fix to build the crate in yocto/ openembedded environments.

See here https://github.com/rust-embedded/meta-rust-bin/issues/16 and https://github.com/alexcrichton/backtrace-rs/issues/25#issuecomment-279742551


alexcrichton:
Thanks!

PR #31, opened 2017-02-17
Add support for demangling C++ frames' symbols

fitzgen:
This adds a new optional (but on-by-default) dependency on the `cpp_demangle` crate. When this feature is enabled, if demangling a frame's symbol as a Rust symbol fails, then we attempt to demangle it as a C++ symbol.

r? @alexcrichton 

I didn't add any tests because I am lazy. Don't hesitate to push back on this if you feel it needs testing.

### DIFF
--- a/src/symbolize/mod.rs
+++ b/src/symbolize/mod.rs
@@ -1,8 +1,10 @@
 use std::fmt;
+#[cfg(not(feature = "cpp_demangle"))]
+use std::marker::PhantomData;
 use std::os::raw::c_void;
 use std::path::Path;
 use std::str;
-use rustc_demangle::{demangle, Demangle};
+use rustc_demangle::{try_demangle, Demangle};
 
 /// Resolve an address to a symbol, passing the symbol to the specified
 /// closure.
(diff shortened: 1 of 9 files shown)


fitzgen:
Woops, ok force pushed a fix for building without the `cpp_demangle` feature. Should be all good now.

src/symbolize/mod.rs
```diff
@@ -108,26 +111,82 @@ impl fmt::Debug for Symbol {
...
             demangled: demangled,
+            cpp_demangled: cpp,
         }
     }
 
-    /// Returns the raw symbol name as `&str` if the symbols is valid utf-8.
-    pub fn as_str(&self) -> Option<&'a str> {
-        self.demangled.as_ref().map(|s| s.as_str())
+    /// Returns the raw symbol name as a `str` if the symbols is valid utf-8.
+    pub fn as_str(&self) -> Option<Cow<'a, str>> {
```
alexcrichton: Could this stay as it was but leave a TODO to fix this on the next major version change?

alexcrichton:
Looks great to me, thanks! Perhaps though this could be off by default to start off with? Also, could you add just a smoke test or two that it works?

fitzgen:
Ran into some issues:

* `rustc_demangle::demangle` does not return a result or option, so we can't know whether demangling suceeded or not, and therefore can't decide to only try and C++ demangle based on that information.

* Some Rust symbols parse as C++ symbols, but format differently, so we can't try C++ first and then fall back to Rust symbols either.

Because of those two things, I don't think we can proceed. I think we need to fix the `rustc_demangle` API first.

alexcrichton:
@fitzgen it looks like there may be some `println!` statements in the published version of cpp_demangle?

alexcrichton:
Oh wait that's int he crate, carry on!

alexcrichton:
It looks like the tests may also be failing on Linux because there's some inlined functions in libstd?

fitzgen:
Yeah I realized this wasn't ready and I need to rewrite parts of it to leverage `rustc_demangle::try_demangle`.

fitzgen:
I'll ping you again once I've done that. Thanks!

fitzgen:
@alexcrichton OK, assuming CI passes, I think this is ready.

build.rs
```diff
@@ -49,3 +52,23 @@ fn expand_serde(out_dir: &Path) {
         serde_codegen::expand(&out_dir.join("tmp.rs"), &dst).unwrap();
     }).unwrap().join().unwrap();
 }
+
+// TODO: how to only do this for tests? I tried to do:
+//
+// > #[cfg(all(test, feature = "demangle_cpp"))]
+//
+// but it did not work...
+#[cfg(feature = "demangle_cpp")]
```
fitzgen: This is unfortunate. Perhaps you know how we can only link the test C++ code when building for tests?

alexcrichton: Ah yeah unfortunately this isn't possible right now, to work around it we'll have to move it to a whole separate crate which is cargo test'd.

fitzgen:
Ok, moved the C++ smoke test to a sub-crate. Hopefully this did not break CI...

alexcrichton:
Thanks!

fitzgen:
Thank *you*!

PR #32, opened 2017-02-18
Add possibility to create a lazy backtrace.

Yamakaky:
#[bench]
    fn b(b: &mut Bencher) {
        b.iter(|| backtrace::Backtrace::new());
    }

    #[bench]
    fn c(b: &mut Bencher) {
        b.iter(|| backtrace::LazyBacktrace::new());
    }

`c()` is three times as fast as `b()`.

### DIFF
--- a/src/backtrace/dbghelp.rs
+++ b/src/backtrace/dbghelp.rs
@@ -14,6 +14,7 @@ use std::mem;
 use winapi::*;
 use kernel32;
 
+#[derive(Clone)]
 pub struct Frame {
     inner: STACKFRAME64,
 }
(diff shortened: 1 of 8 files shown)


Yamakaky:
Comments welcome on the comments ;)

alexcrichton:
Perhaps instead of adding a new type we could reuse the existing `Backtrace` type? Something like:

```rust
let mut b = Backtrace::new_unresolved();
println!("{}", b); // no symbols
b.resolve();
println!("{}", b); // symbols displayed
```

Yamakaky:
You mean with an enum? Could be, but it adds 1 byte for each backtrace.

alexcrichton:
So long as it's the same type, fine by me.

Yamakaky:
Quick update: we could use https://github.com/reem/rust-lazy

oli-obk:
Any update on this?

PR #35, opened 2017-03-05
Remove backtrace-sys dependency on libc

nox:


nox:
This is not a breaking change.

alexcrichton:
Thanks for the PR, but this crate needs libc because it compiles C code which uses libc.

Could you explain the rationale for this PR?

nox:
Oh dang I'm blind. This patch makes it crystal clear that any libc bump is non-breaking, though.

The rationale is mentioned in the issue linked in this PR, https://github.com/servo/servo/issues/15824, trying to measure and minimise the impact of a libc bump.

alexcrichton:
Ok, then in that case I'm going to hold off on merging it as there's no immediate benefit, I don't think we'll need it in the long run, and I believe it's more correct to leave it.

nox:
Ok.

nox:
> I believe it's more correct to leave it.

I'm not sure I understand this part of your comment, what's "it"?

When Cargo is taught about the concept of public dependency, will you be ok with making `libc` a private dependency of both `backtrace-sys` and `backtrace`?

nox:
I'm also not sure what is "more correct".

alexcrichton:
"it" being `extern crate libc` and "more correct" meaning that this is how the crate declares a dependency on libc, through the `libc` crate.

Cargo is not teaching us to not have public dependencies, that's the wrong lesson to learn. Cargo shows the importance of carefully selecting public dependencies.

nox:
@alexcrichton That is indeed true, but would you accept a PR not removing the `extern crate libc` statement, but keeping the removal of references to `libc::uintptr_t`?

nox:
@alexcrichton Or do you think `libc` should be a public dependency? If so, why?

alexcrichton:
I think that's missing the point, these types are `uintptr_t` from libc, so we need `libc::uintptr_t`

nox:
But I've been told on `#rust-internals` that this type is guaranteed to be `usize` all the time. Is that wrong?

nox:
I understand that `uintptr_t` and `usize` don't spell the same, but how is that different from mixing `libc` and `std::os::raw` types in `backtrace-sys`?

```rust
use libc::uintptr_t;
use std::os::raw::{c_void, c_char, c_int};
```

alexcrichton:
I feel like at this point you're trying to belligerently merge PRs dropping libc for the sole purpose of removing libc itself, no other reason.

The *worry* here, if I understand it correctly, is that if libc upgrades across a major version then if two libc types are shared they may not be the same. This is correct for types like `enum` and `struct`, but not correct for types like `type foo = bar`. All of the types in question here are just `type foo = bar`, and by definition of libc their definition cannot change across major versions.

At this point this is really sucking up more of my time than I intended. If we're to continue this please lay out concrete reasons why such a change is necessary, why it's beneficial, and why we should do it **now**.

nox:
I'm sad you take it this way, let's drop this discussion.

SimonSapin:
@alexcrichton I don’t think there is any belligerence here. The branch’s name shouldn’t be taken too seriously, and certainly not personally. Dropping libc dependencies is valuable to reduce the pain when a breaking change is eventually made to libc. In particular, `c_void` is an enum and so incompatible across libc versions. We’ve lived through this pain before. The name libcpocalypse was not given lightly.

PR #37, opened 2017-03-15
Support unresolved backtraces.

ncalexan:
This is a reworking of @Yamakaky's https://github.com/alexcrichton/backtrace-rs/pull/32.  It's a little awkward (`.frames()` doesn't make sense for `Backtrace::Unresolved`) and the tests fail for reasons I don't understand, but I want to see progress on this ticket so I'm posting my attempt.  Perhaps others see a better path forward?

### DIFF
--- a/src/capture.rs
+++ b/src/capture.rs
@@ -27,8 +27,17 @@ use {trace, resolve, SymbolName};
 #[derive(Clone)]
 //~ HACK1 #[derive(RustcDecodable, RustcEncodable)]
 //~ HACK2 #[derive(Deserialize, Serialize)]
-pub struct Backtrace {
-    frames: Vec<BacktraceFrame>,
+pub enum Backtrace {
+    /// A resolved backtrace,
+    Resolved {
+        /// The resolved backtrace frames.
+        frames: Vec<BacktraceFrame>,
+    },
+    /// An unresolved backtrace.
+    Unresolved {
+        /// The unresolved frame captures to resolve in the future.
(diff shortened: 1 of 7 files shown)


Yamakaky:
`.frames()` could resolve the backtrace.

ncalexan:
> .frames() could resolve the backtrace.

I tried this, but result of `resolve()` is dropped before the reference can be returned.  This could be achieved by making `resolve(&mut self)`; or potentially by making `frames()` return something that implemented `Deref<Target=&[BacktraceFrame]>`.

src/capture.rs
```diff
@@ -27,8 +27,17 @@ use {trace, resolve, SymbolName};
 #[derive(Clone)]
 //~ HACK1 #[derive(RustcDecodable, RustcEncodable)]
 //~ HACK2 #[derive(Deserialize, Serialize)]
-pub struct Backtrace {
-    frames: Vec<BacktraceFrame>,
+pub enum Backtrace {
```
alexcrichton: Unfortuntely the way visibility works means that the internal details of this are exposed, but perhaps a representation like:

```rust
pub struct Backtracke {
    inner: Inner,
}

enum Inner {
    Resolved { ... }
    Unresolved { ... }
}
```

alexcrichton:
Thanks! I don't think we want `frames()` to panic in the case of an unresolved backtrace though. Can we always store a valid list of frames and then just lazily fill them in when `resolve()` is called?

PR #43, opened 2017-04-19
explicit convert dlopen parameter filename to *const c_char

### LINKED ISSUE #42: mismatched types expected i8, found u8 targeting i686-linux-android
```
  --> /home/jarod/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.0/src/dylib.rs:30:32
   |
30 |         let ptr = libc::dlopen(name.as_ptr(), libc::RTLD_LAZY);
   |                                ^^^^^^^^^^^^^ expected i8, found u8
   |
   = help: here are some functions which might fulfill your needs:
           - .offset(...)
           - .wrapping_offset(...)

```

jarod:
to fix #42

### DIFF
--- a/src/dylib.rs
+++ b/src/dylib.rs
@@ -3,7 +3,7 @@ use std::marker;
 use std::mem;
 use std::sync::atomic::{AtomicUsize, Ordering};
 
-use libc::{self, c_void};
+use libc::{self, c_char, c_void};
 
 pub struct Dylib {
     pub init: AtomicUsize,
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! I think this is fixed in more recent versions of Rust though? The `c_char` type in libstd was corrected and should reflect the libc crate now I believe.

What version of Rust are you using?

jarod:
I'm using the latest stable rust. 
```
$ rustc --version
rustc 1.16.0 (30cf806ef 2017-03-10)
```

alexcrichton:
Ok I think this is fixed on nightly so let's merge in the meantime while we're waiting for that to propagate to stable.

PR #48, opened 2017-05-20
Utilize cfg_attr instead of a build script.

Mark-Simulacrum:
This also updates serde to a more recent version (1.0).

`cfg_attr` is now expanded with `include!` (though we don't use it anymore) since https://github.com/rust-lang/rust/pull/33706. 

`serde_derive` also supports `cfg_attr`, so we should be fine in that respect as well.

This is, I believe, a *breaking change* due to the bump in the serde dependency from 0.8 to 1.0.


Mark-Simulacrum:
Odd that AppVeyor succeeded, looking into the failure.

alexcrichton:
👍 Nice!

PR #49, opened 2017-05-20
Fix nightly error due to unused_macros lint

Mark-Simulacrum:
We could annotate `macro_rules! dlsym` itself with the appropriate cfg attributes, but I think this is cleaner.


alexcrichton:
Thanks!

PR #50, opened 2017-05-20
no_std support

Mark-Simulacrum:
This is likely broken and unfinished, but putting it up so that I can get some form of testing on Windows and for review, if we want.


Mark-Simulacrum:
Discussed some on IRC and it looks like while this is something we'd be interested in for the future, right now it's not feasible to utilize this inside std, and I don't want to take the time now. I won't delete the branch though so if anyone wants to push this forward and get it done I'm sure people will be happy!

PR #51, opened 2017-05-21
Do not use `cpp_demangle`'s default features

fitzgen:
This avoids building its `c++filt` clone, and the dependencies it has, such as `clap`.


fitzgen:
Had `"false"` instead of `false`... grumble grumble `bash` got things right .. *ducks*

alexcrichton:
👍

PR #53, opened 2017-06-18
On FreeBSD use gmake

kirillrdy:
currently I can't build backtrace-rs on FreeBSD.
using gmake backtrace-rs can be built on FreeBSD.

I am not a direct consumer of backtrace but via error-chain crate, but I tried your simple example from readme and it seems to be working.
One of tests is failing but i did't have a close look into why.


alexcrichton:
Thanks!

PR #54, opened 2017-06-19
use GNU-make on BSDs

semarie:
"make" is BSD make on all BSDs. Use "gmake" if the host (and not the target) is a BSD.

this commit would be required to build nightly rustc on BSD.


semarie:
Cc @kirillrdy, as you provided a PR for FreeBSD previously and I slightly changed the code...

Issue #44, opened 2017-05-04
Cross-compile for androideabi on -msvc target fails


vvanders:
```
error: failed to run custom build command for `backtrace-sys v0.1.10`
process didn't exit successfully: `...target\debug\build\backtrace-sys-0c4d13204b0e1510\build
-script-build` (exit code: 101)
--- stdout
OPT_LEVEL = Some("0")
TARGET = Some("armv7-linux-androideabi")
HOST = Some("x86_64-pc-windows-msvc")
TARGET = Some("armv7-linux-androideabi")
TARGET = Some("armv7-linux-androideabi")
HOST = Some("x86_64-pc-windows-msvc")
CC_armv7-linux-androideabi = None
CC_armv7_linux_androideabi = None
TARGET_CC = None
CC = None
TARGET = Some("armv7-linux-androideabi")
HOST = Some("x86_64-pc-windows-msvc")
CFLAGS_armv7-linux-androideabi = None
CFLAGS_armv7_linux_androideabi = None
TARGET_CFLAGS = None
CFLAGS = None
PROFILE = Some("debug")
running: "...\\.cargo\\registry\\src\\github.com-1ecc6299db9ec823\\backtrace-sys-0.1.10\\src/libbacktrace/c
onfigure" "--with-pic" "--disable-multilib" "--disable-shared" "--disable-host-shared" "--target=armv7-linux-androideabi
" "--host=x86_64-pc-windows-msvc"

--- stderr
thread 'main' panicked at 'failed to get status: %1 is not a valid Win32 application. (os error 193)', ...\.c
argo\registry\src\github.com-1ecc6299db9ec823\backtrace-sys-0.1.10\build.rs:131
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```
Hitting this with some JNI libraries, not quite sure what the fix is here but happy to take a shot at it if I could get some pointers.

vvanders:
Also happens on GNU, this looks like the offending SHA1: 390ea2a3b630d5bcb83b02c8c6ce814b9d5df7d8

Also the dependency that specified this(jni-rs) has 0.1.2 in the dependency chain but 0.1.10 somehow is getting pulled in.

vvanders:
Looking at it a bit more jni-rs declares error-chain 0.5.0 which references backtrace 0.2.1 but I'm getting backtrace 0.2.3 in the Cargo.lock. Any idea why the versioning isn't working right here?

alexcrichton:
[...]

PR #57, opened 2017-07-14
Allow cross-compilation to Linux from Windows

alusch:
Addresses at least some of the problems with cross-compiling from Windows in #44 (verified that `arm-unknown-linux-gnueabihf` works).

1. Remove `--build` parameter to `configure`
2. Replace backslashes in the path to `configure` with forward slashes
3. Explicitly set the `AR` environment variable when calling `configure`

The second item requires a bit of gymnastics to do the replacement without converting to `String` as the path may not be representable as UTF-8.

Tested native and `arm-unknown-linux-gnueabihf` builds on Windows and Linux.

### DIFF
--- a/backtrace-sys/build.rs
+++ b/backtrace-sys/build.rs
@@ -4,6 +4,8 @@ use std::env;
 use std::ffi::OsString;
 use std::fs;
 use std::io;
+#[cfg(windows)]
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! I'll touch this up a bit after merging but looks good to me!

PR #60, opened 2017-07-23
Don't read past the Name buffer if the symbol name exceeds the max length

### LINKED ISSUE #59: Prints garbage for function name if it exceeds MAX_SYM_NAME (2000) characters
image

(Sorry for screenshot, I can't copy the garbage out of the terminal window.)

I had it AV once as well in the `OsString::from_wide` call.

The reason is that [`SymFromAddrW`](https://github.com/alexcrichton/backtrace-rs/blob/d816fff4e4ca5db847a2e0245732494833c95863/src/symbolize/dbghelp.rs#L66) will set `info.NameLen` to the real length, which can be greater than `MAX_SYM_NAME`. So when backtrace [converts the buffer to a widestring,](https://github.com/alexcrichton/backtrace-rs/blob/d816fff4e4ca5db847a2e0245732494833c95863/src/symbolize/dbghelp.rs#L74) it should use min of `info.NameLen` and `info.MaxNameLen` rather than just `info.NameLen` - the latter causes it to read past the buffer into garbage. This behavior doesn't seem to be documented as far as I can tell, but it's what I observed in the debugger.

(I don't apologize for not boxing my futures :P )

Arnavion:
SymFromAddrW sets NameLen to the real length of the symbol, which can be
greater than MaxNameLen. Before this change, such a symbol would cause
the code to read past the end of the Name buffer into garbage. With this
change, the code reads only till the min of NameLen and (MaxNameLen - 1).

Fixes #59

### DIFF
--- a/src/symbolize/dbghelp.rs
+++ b/src/symbolize/dbghelp.rs
@@ -70,8 +70,15 @@ pub fn resolve(addr: *mut c_void, cb: &mut FnMut(&super::Symbol)) {
         if ret != TRUE {
             return
         }
+
+        // If the symbol name is greater than MaxNameLen, SymFromAddrW will
+        // give a buffer of (MaxNameLen - 1) characters and set NameLen to
+        // the real value.
+        let name_len = ::std::cmp::min(info.NameLen as usize,
+                                       info.MaxNameLen as usize - 1);
+
         let name = slice::from_raw_parts(info.Name.as_ptr() as *const u16,
-                                         info.NameLen as usize);
+                                         name_len);
         let name = OsString::from_wide(name);
 
         let mut line = mem::zeroed::<IMAGEHLP_LINEW64>();
--- a/tests/long_fn_name.rs
+++ b/tests/long_fn_name.rs
@@ -1,0 +1,56 @@
+extern crate backtrace;
+
+#[cfg(all(windows, feature = "dbghelp"))]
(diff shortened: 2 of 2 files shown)


Arnavion:
I can add a test case based on this if you want:

```rust
extern crate backtrace;
use backtrace::Backtrace;

mod _234567890_234567890_234567890_234567890_234567890 {
    pub struct _234567890_234567890_234567890_234567890_234567890<T>(T);
    impl<T> _234567890_234567890_234567890_234567890_234567890<T> {
        pub fn new() -> ::Backtrace {
			::Backtrace::new()
        }
    }
}

fn main() {
    ::std::env::set_var("RUST_BACKTRACE", "1");

    use _234567890_234567890_234567890_234567890_234567890::
        _234567890_234567890_234567890_234567890_234567890 as S;

    let bt = S::<S<S<S<S<S<S<S<S<S<i32>>>>>>>>>>::new();
    println!("{:?}", bt); // Verify that frame[2]'s name is <= 2000
}
```

alexcrichton:
Unusual! 

Would it be possible to include that test though? I think we'd just only activate it for MSVC targets, right?

Arnavion:
How do I do that? Is there a feature for -gnu vs -msvc?

alexcrichton:
Thanks!

PR #62, opened 2017-08-10
Fix trace for iOS arm builds

king6cong:


alexcrichton:
Thanks for the PR! I don't think this is quite the right strategy, though. Does libunwind not work at all on iOS? Could you provide more of a description in this PR?

If this really must be disabled with no replacement then [this `cfg_if!`](https://github.com/alexcrichton/backtrace-rs/blob/3d96a9242ed2096984d15d177f4762b699bee6d4/src/backtrace/mod.rs#L90-L112) should be modified rather than the implementaiton here specifically.

king6cong:
Yeah, this can be reproduced by:

cargo build --target armv7-apple-ios
cargo build --target armv7s-apple-ios
```
error[E0425]: cannot find function `_Unwind_Backtrace` in module `uw`
  --> src/backtrace/libunwind.rs:53:13
   |
53 |         uw::_Unwind_Backtrace(trace_fn, &mut cb as *mut _ as *mut _);
   |             ^^^^^^^^^^^^^^^^^ not found in `uw`
```

but `cargo build --target aarch64-apple-ios` works well, have not looked into this yet ;)

king6cong:
Updated :)

PR #65, opened 2017-08-16
Add a feature to symbolicate with `gimli-rs/addr2line`

fitzgen:
This commit adds a new feature, "gimli", which enables symbolicating addresses into file/line/function information using the `gimli-rs/addr2line` crate. Unwinding is still performed with `libunwind`.

This is currently Linux only. We could enable this for macOS as well if we had a way to find the dSYM bundle for a given shared library or executable file. That involves looking for the LC_UUID load command in the mach-o file, and then using spotlight APIs to find the corresponding bundle containing the DWARF debug infos.

At least on Linux, this is one less C dependency. RIIR!!

------------------------------------------------

Note: we are waiting on one more bug fix before publishing a new release of `addr2line` to crates.io, so the Cargo.toml file has local path dependencies temporarily. Once we publish a new release, I'll fix that. I figure review can commence, nonetheless, as I don't expect any APIs to change.

r? @alexcrichton 

cc @philipc @jonhoo @tromey

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,28 @@
 language: rust
+
 rust:
   - stable
   - beta
   - nightly
+
 sudo: false
+
+dist: trusty
+
+addons:
+  sources:
+    # Provides clang-3.9
+    - llvm-toolchain-trusty-3.9
+  apt:
+    packages:
+      # Required for `bindgen`, which is required by `findshlibs`, which is
+      # required by the `gimli` feature.
+      - clang-3.9
+
 before_script:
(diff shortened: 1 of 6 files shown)


alexcrichton:
Looks great to me, thanks!

I'd be down for merging with a green CI

fitzgen:
@alexcrichton ok, just updated the commit. This should be ready to merge, assuming the CI comes back green, which I expect it to.

alexcrichton:
Nice!

PR #71, opened 2017-10-14
Set MAKEFLAGS to CARGO_MAKEFLAGS for the make command

Eijebong:


Eijebong:
Is this my fault or is it something that happens sometimes ?

alexcrichton:
Nah that is preexisting and I need to fix it, thanks!

PR #72, opened 2017-10-16
Support compilation with ccache

not-fl3:
That crate was the only dependency which fails to compile with ```CC="ccache gcc" CXX="ccache c++" cargo run``` for me, while with this fix it works fine. 

Maybe that should be done on ```cc-rs``` side, but as far as I understand - interaction with native "configure" script is not in ```cc-rs``` responsibility.

### DIFF
--- a/backtrace-sys/build.rs
+++ b/backtrace-sys/build.rs
@@ -40,6 +40,39 @@ fn find_tool(compiler: &cc::Tool, cc: &str, tool: &str) -> PathBuf {
         .unwrap_or_else(|| PathBuf::from(tool))
 }
 
+fn compiler_whitelist_prefix(compiler: &cc::Tool) -> Option<String> {
+    let whitelist = ["ccache", "distcc", "sccache"];
+    let cc = compiler.path().file_name().unwrap().to_str().unwrap();
+
+    whitelist.iter().find(|t| cc == **t).map(|t| t.to_string())
+}
+
+fn compiler_cc_variable(compiler: &cc::Tool) -> String {
+    let cc = compiler.path().file_name().unwrap().to_str().unwrap();
+
+    match compiler_whitelist_prefix(compiler) {
+        Some(prefix) => prefix + " " + compiler.args()[0].to_str().unwrap(),
+        None => cc.to_string()
+    }
+}
+
+fn compiler_flags(compiler: &cc::Tool) -> OsString {
+    let compiler_in_whitelist = compiler_whitelist_prefix(compiler).is_some();
+    let mut flags = OsString::new();
+    for (i, flag) in compiler
+        .args()
+        .iter()
+        .skip(if compiler_in_whitelist { 1 } else { 0 })
+        .enumerate()
+    {
+        if i > 0 {
+            flags.push(" ");
+        }
+        flags.push(flag);
+    }
+    flags
+}
+
 fn main() {
     let src = env::current_dir().unwrap();
     let dst = PathBuf::from(env::var_os("OUT_DIR").unwrap());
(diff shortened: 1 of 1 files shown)


alexcrichton:
What's the error you get with `CC="ccache gcc"` when compiling this crate? This seems somewhat invasive and I'm curious if there's perhaps a lighter touch that can fix this.

not-fl3:
```
CC="ccache gcc" CXX="ccache c++" cargo build                                                                 :(
   Compiling backtrace-sys v0.1.16 (file:///home/fl3/fun/backtrace-rs/backtrace-sys)
error: failed to run custom build command for `backtrace-sys v0.1.16 (file:///home/fl3/fun/backtrace-rs/backtrace-sys)`
process didn't exit successfully: `/home/fl3/fun/backtrace-rs/target/debug/build/backtrace-sys-f02e54dce75403d3/build-script-build` (exit code: 101)
--- stdout
OPT_LEVEL = Some("0")
TARGET = Some("x86_64-unknown-linux-gnu")
HOST = Some("x86_64-unknown-linux-gnu")
TARGET = Some("x86_64-unknown-linux-gnu")
TARGET = Some("x86_64-unknown-linux-gnu")
HOST = Some("x86_64-unknown-linux-gnu")
CC_x86_64-unknown-linux-gnu = None
CC_x86_64_unknown_linux_gnu = None
HOST_CC = None
CC = Some("ccache gcc")
TARGET = Some("x86_64-unknown-linux-gnu")
HOST = Some("x86_64-unknown-linux-gnu")
CFLAGS_x86_64-unknown-linux-gnu = None
CFLAGS_x86_64_unknown_linux_gnu = None
HOST_CFLAGS = None
CFLAGS = None
DEBUG = Some("true")
running: "sh" "/home/fl3/fun/backtrace-rs/backtrace-sys/src/libbacktrace/configure" "--with-pic" "--disable-multilib" "--disable-shared" "--disable-host-shared" "--host=x86_64-unknown-linux-gnu" "--build=x86_64-unknown-linux-gnu"
checking build system type... x86_64-unknown-linux-gnu
checking host system type... x86_64-unknown-linux-gnu
checking target system type... x86_64-unknown-linux-gnu
checking for x86_64-unknown-linux-gnu-gcc... ccache
checking for C compiler default output file name... a.out
checking whether the C compiler works... yes
checking whether we are cross compiling... no
checking for suffix of executables... 
--- stderr
configure: error: in `/home/fl3/fun/backtrace-rs/target/debug/build/backtrace-sys-d323e05ef3125578/out':
configure: error: cannot compute suffix of executables: cannot compile and link
See `config.log' for more details.
thread 'main' panicked at 'failed with: exit code: 1', backtrace-sys/build.rs:176:8
```


If I understand correctly - we call command "configure" and set environment variable CC from rust code. And we set it as a "ccache" instead of "ccache gcc". Than configure tries to check the compiler and call "$CC", which cant compile anything. 

I am unfamiliar with autotools configure things, so probably there is a more elegant solution.

backtrace-sys/build.rs
```diff
@@ -89,23 +122,17 @@ fn main() {
...
+    let ar = find_tool(&compiler, &cc, "ar");
+    let ranlib = find_tool(&compiler, &cc, "ranlib");
     let mut cmd = Command::new("sh");
 
     cmd.arg(configure)
        .current_dir(&dst)
        .env("AR", &ar)
        .env("RANLIB", &ranlib)
-       .env("CC", compiler.path())
+       .env("CC", cc.clone())
```
alexcrichton: I wonder if we can just fold all the flags into this env var? Something like setting `CC="sccache gcc"` for the configure script as well?

not-fl3: You mean something like 

```rust
let mut cc_with_flags = compiler.path().to_path_buf().into_os_string();
cc_with_flags.push(" ");
cc_with_flags.push(flags);
```
And then
```rust
    .env("CC", &cc_with_flags)
    .env("CFLAGS", "")
```
?

It works and its definitely less code! But not sure about different platforms or something. 

Just as idea - maybe make function to get CC and CFLAGS from cc-rs's ```Tool``` and add it to ```cc-rs```? It may be usefull for such cases.

alexcrichton: Yeah that's what I was thinking, although this may or may not be one of those build systems that rejects such a pattern.

Although along the lines of modifying the `cc` crate I wonder if we could perhaps fix this upstream? The `cc` crate knows when it recognizes ccache/sccache and whatnot, so I wonder if we could make `Tool::path` return the actual compiler, not the ccache path? That way all tools looking at paths and such would "just work", but compilations using `cc` would also use sccache?

not-fl3: As for me - it would be a bit confusing for a user - some crates will use ccache and some will not.

I think the root of the problem - that some of build systems ignore compiler arguments sometimes, so we cant use compiler path as a compilation command. 

Maybe something like this will work? 

when compiling with CC="ccache gcc"  ```cc``` would return "gcc" as compiler path and flags vec without "gcc" in the first element. But at the same time, it will have ```.cc()``` and ```.cflags()``` methods just for that case. And those methods may be used for other env-variables based build systems.

alexcrichton: Yeah that strategy sound plausible to me!

PR #74, opened 2017-11-12
Don't use libbacktrace on fuchsia

tkilbourn:
This unblocks us from using backtrace-rs on Fuchsia while we sort out how to get backtraces.


alexcrichton:
Thanks!

PR #79, opened 2017-12-21
migrate to winapi 0.3

steffengy:
This should be ready to land as is once CI passes and after another winapi release....

Also the kernel32 feature seems to be unused?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,9 +30,7 @@ findshlibs = { version = "0.3.3", optional = true }
 libc = "0.2"
 
 [target.'cfg(windows)'.dependencies]
-dbghelp-sys = { version = "0.2", optional = true }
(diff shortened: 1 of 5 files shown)


alexcrichton:
Thanks! Yeah it's fine to drop the unused features.

Looks good to merge once CI is green!

steffengy:
Also blocked on another winapi release for https://github.com/retep998/winapi-rs/pull/538 for now.

steffengy:
So updated for CI:
- The linux builders fail with 'b.bind("::1:0") failed with: Cannot assign requested address (os error 99)' again (which they seem to do for quite some time)
- The mac builders now fail with `pip: command not found`, which seems to be quite recent

I can't align any of that with these changes, so I'll assume the breakage is caused by general CI updates/issues?

PR #87, opened 2018-01-23
Update gimli symbolizer to return inlined functions

philipc:
The addr2line crate can now return inlined functions. It also requires us to load the file ourselves now.

There doesn't appear to be any tests for inlined functions. I hacked the smoke test to try it out, but haven't included that in this PR since I wasn't sure how well other symbolizers would handle that.

cc @fitzgen

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -88,35 +88,15 @@ extern crate rustc_demangle;
 #[cfg(feature = "cpp_demangle")]
 extern crate cpp_demangle;
 
-#[cfg(all(feature = "gimli-symbolize",
-          unix,
-          target_os = "linux"))]
-extern crate addr2line;
-#[cfg(all(feature = "gimli-symbolize",
-          unix,
-          target_os = "linux"))]
-extern crate findshlibs;
-#[cfg(all(feature = "gimli-symbolize",
-          unix,
-          target_os = "linux"))]
-extern crate gimli;
-#[cfg(all(feature = "gimli-symbolize",
-          unix,
-          target_os = "linux"))]
-extern crate memmap;
-#[cfg(all(feature = "gimli-symbolize",
-          unix,
-          target_os = "linux"))]
-extern crate object;
-#[cfg(all(feature = "gimli-symbolize",
-          unix,
-          target_os = "linux"))]
-#[macro_use]
-extern crate rental;
-#[cfg(all(feature = "gimli-symbolize",
-          unix,
-          target_os = "linux"))]
-extern crate stable_deref_trait;
+cfg_if! {
+    if #[cfg(all(feature = "gimli-symbolize", unix, target_os = "linux"))] {
+        extern crate addr2line;
+        extern crate findshlibs;
+        extern crate gimli;
+        extern crate memmap;
+        extern crate object;
+    }
+}
 
 #[allow(dead_code)] // not used everywhere
 #[cfg(unix)]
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks for the PR! Would it be possible to trim down the deps a bit? I'd be ok with a little unsafety here, there's already plenty elsewhere in the crate!

src/symbolize/gimli.rs
```diff
@@ -109,33 +151,46 @@ pub fn resolve(addr: *mut c_void, cb: &mut FnMut(&super::Symbol)) {
...
+                    .and_then(|f| f.raw_name().ok().map(|f| f.to_string()));
+                let sym = super::Symbol {
+                    inner: Symbol::new(addr.0 as usize, file, line, name),
+                };
+                cb(&sym);
+                found_sym = true;
+            }
+        }
+
+        if !found_sym {
```
fitzgen: Nice. Maybe leave a comment about the symbol table fallback here?

src/symbolize/gimli.rs
```diff
@@ -109,33 +151,46 @@ pub fn resolve(addr: *mut c_void, cb: &mut FnMut(&super::Symbol)) {
...
+                cb(&sym);
+            }
+        }
     });
 }
 
 pub struct Symbol {
     addr: usize,
-    file: PathBuf,
+    file: Option<PathBuf>,
```
fitzgen: This is optional now because the fallback might not have this info, right?

src/lib.rs
```diff
@@ -96,6 +96,27 @@ extern crate addr2line;
...
+extern crate object;
+#[cfg(all(feature = "gimli-symbolize",
+          unix,
+          target_os = "linux"))]
+#[macro_use]
+extern crate rental;
+#[cfg(all(feature = "gimli-symbolize",
+          unix,
+          target_os = "linux"))]
+extern crate stable_deref_trait;
```
fitzgen: Does this crate already depend on `cfg-if`? If so, maybe we could do

```rust
cfg_if! {
    if #[cfg(all(feature = "gimli-symbolize", unix, target_os = "linux"))] {
        extern crate addr2line;
        extern crate findshlibs;
        extern crate gimli;
        extern crate memmap;
        extern crate object;
        extern crate rental;
        extern crate stable_deref_trait;
    } else {
        // nothing here, but cfg-if! requires an else
    }
}
```

Which is less `#[cfg(..)]` noise.

fitzgen:
> Thanks for the PR! Would it be possible to trim down the deps a bit? I'd be ok with a little unsafety here, there's already plenty elsewhere in the crate!

Are you specifically referring to `rental` and `stable_deref_trait`?

It isn't clear to me based on this comment what changes are necessary for merging this PR.

alexcrichton:
Er sorry yeah, if this could avoid rental/stable_deref_trait that'd be great (I already only have a passing understanding of what's going on). Beyond that though looks good!

src/lib.rs
philipc: Done, but without the else because cfg-if doesn't require it anymore.

src/symbolize/gimli.rs
philipc: Both that and because addr2line returns an option too, for the unlikely case where we found a function name, but not location info.

philipc:
I've removed the rental/stable_deref_trait deps and addressed the other comments. Subverting the borrow checker isn't something I'm used to, so please check the mem::transmute usage.

philipc:
The use of  '?' for Option broke the rustc 1.21 build. Should I avoid that or bump the version?

alexcrichton:
Oh oops! Nah I think it's fine to just bump the required version

PR #91, opened 2018-03-26
Remove indentation

king6cong:
I am not sure whether this is intended but it seems a little strange to me 😃


alexcrichton:
Thanks!

PR #108, opened 2018-05-31
travis ci for ios and android

kali:
* setup android toolchain
* builds tests but don't run them
* uses dinghy for toolchain discovery and configuration

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,6 @@
 language: rust
 sudo: false
 dist: trusty
-cache:
-  - cargo
-  - directories:
-      - cached
 
 addons:
   sources:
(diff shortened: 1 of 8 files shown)


alexcrichton:
Awesome, thanks for this! I've personally had bad luck with Cargo caches on Travis historically so I'd prefer to avoid those if we can. Additionally, I personally prefer to use Docker when we can because it helps shield projects from changes on Travis as well as allowing local developers to reproduce CI and test out changes without going through Travis. Finally, is `cargo dinghy` needed if we're not running tests? I think it'll be awhile before tests can be run so I'm curious if that's still needed (I haven't personally used it before).

I pushed a test commit to https://github.com/alexcrichton/backtrace-rs/commit/3e6115d5fcbed4792fc752791464c1f557372775 which moves all the Android builds out of .travis.yml into the `ci/docker` folder and am [waiting on CI](https://travis-ci.org/alexcrichton/backtrace-rs/builds/386222270) to see how it works, but I'm curious to hear what you think on these!

kali:
* OK about cargo cache. I mostly set up the cache to avoid compiling dinghy over and over.
* dinghy main contribution here is to setup the environment according to the toolchains. TARGET_CC, CARGO_xxx_LINKER, etc. It could be replaced by a few lines of shell, but going through the dance of writting them on my laptop and iterating them in travis is so cumbersome... Switching to docker would be a valid approach.
* Last time I checked, running tests in the Android emulator is relatively easy with dinghy on cargo. iOS dinghy tests in emulator have been broken by an XCode upgrade a few month ago, and I could not get it to work again.
* As for the huge compatibility matrix, many of the recent problems with libbacktrace where due to an old config.sub that could not make sense of rustc triples, so where very architectude dependent... There is so little test or test-build coverage in the ecosystem for mobile targets, I'm tempted to say we should just take it when we can...

kali:
All right, I got ios to work without dinghy. The library is small, so the environment is not that complicated to setup.

alexcrichton:
Awesome, thanks!

PR #114, opened 2018-07-05
Added mgwhelp support (Dr. MinGW)

matanui159:
Should fix https://github.com/alexcrichton/backtrace-rs/issues/66.

Currently its not working as gcc can't find `-lmgwhelp` for some reason even though I installed Dr. MinGW. Posted here anyway to see if I could get some help with it as this is the first time I have used Rust FFI.

Update: I'm able to find mgwhelp with gcc on its own, trying to find a way to pass `-Wl,--verbose` to gcc through rustc or cargo to get a better idea of whats going on.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,7 @@ backtrace-sys = { path = "backtrace-sys", version = "0.1.17", optional = true }
 # Note that not all features are available on all platforms, so even though a
 # feature is enabled some other feature may be used instead.
 [features]
-default = ["libunwind", "libbacktrace", "coresymbolication", "dladdr", "dbghelp"]
+default = ["libunwind", "libbacktrace", "coresymbolication", "dladdr", "dbghelp", "mgwhelp"]
 
 #=======================================
 # Methods of acquiring a backtrace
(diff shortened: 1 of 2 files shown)


alexcrichton:
Seems plausible to me! Once CI passes I'd be fine merging this

matanui159:
After figuring out the proper place to install Dr. MinGW (btw apparently rust installs its own mingw toolchain, I don't know why), it worked somewhat but only returned blank empty names, so I thought it was that I didn't have libdwarf installed, so I tried installing that, but I couldn't as it required libelf or something and at this point I feel like its just too much effort. For me personally, I use pc-windows-gnu because CLion doesn't support the MSVC debugger, but I'll probably use pc-windows-msvc for release making it:

 - pc-windows-gnu for development, where the debugger works and thus I don't need a backtrace.
 - pc-windows-msvc for release where I won't be able to use a debugger (unless they gave me access to their computer, which won't happen) so a backtrace will be more helpful.

If you want to try and get it working, I have it set to allow edits from maintainers, but for me at least its too much effort for something that won't be useful too often.

alexcrichton:
Ok no worries! I'm gonna close this in the meantime, but if it's figured out we can try to merge!

Issue #83, opened 2018-01-04
'assertion failed: sym - actual_fn_pointer < 1024'


ignatenkobrain:
```
	thread 'smoke_test_frames' panicked at 'assertion failed: sym - actual_fn_pointer < 1024', tests/smoke.rs:74:12
```

Not sure how to debug it, I'm running Fedora 28 x86_64 with rustc 1.22.1 and cargo 0.23.0.

alexcrichton:
I'm going to close this to clean up the issue tracker here. If the test is still failing it probably needs a code change to get it running here, but I don't want to track getting the test suite running on all platforms here in the issue tracker.

PR #119, opened 2018-07-13
Added Backtrace::from_here() and Backtrace::from_here_unresolved()

jchlapinski:
This PR adds two convenience constructors, which differ from `new` and `new_unresolved` respectively in that trace frames from backtrace crate itself are removed. Backtraces start directly with either constructor's call-site.


alexcrichton:
Thanks! I wonder though if this should just be the default behavior of `Backtrace::new`? I'm not actually sure why you'd *for sure want* per se the frames of the backtrace crate itself

alexcrichton:
Or an alternative would be to just not print them by default, but perhaps have an internal option which could be flipped to print them and/or iterate over them?

jchlapinski:
I think that this should be the default behavior of backtrace crate to omit its own frames, as those are just noise from the user perspective, I believe. If nobody objects I will rework this PR to replace original constructors and of course fix the checks.

alexcrichton:
Looking good! It looks like the submodule may be updated here accidentally though? Additionally, could the filtering happen when acquiring a slice or accessing the frames? That way the `Backtrace` instance could contain a lossless record of all frames, but by default when printing or accessing it'd only return a subset.

jchlapinski:
@alexcrichton honestly I do not know why the submodule seems to updated in this PR. I did not change anything there and definitely did not commit any files from `backtrace-sys`. After checking out origin/master in the submodule, git shows me no changes on main repo, as expected, so I do not know how to proceed to remove those changes from the PR?

jchlapinski:
@alexcrichton,
> Additionally, could the filtering happen when acquiring a slice or accessing the frames? That way the Backtrace instance could contain a lossless record of all frames, but by default when printing or accessing it'd only return a subset.

I think that is the best way to go. I have added additional `ext_index` field to `Backtrace`, where I store index of the first "outer" frame in the backtrace, however all frames are otherwise stored. Now `println!("{:?}", Backtrace::new())` will display only external part of the backtrace, but full backtrace can be displayed by `println!("{:#?}", Backtrace::new())`. I also added some public getters for convenience and completness.

This PR also fixes #83

jchlapinski:
@alexcrichton I left `frames()` as is, but I have added `ext_frames()` to get only external frames. We can of course revert how those getter work by default, I was however vary that someone might already depend on how `frames()` worked, perhaps performing some additional processing.

jchlapinski:
As a side note, the only check that fails on travis now is related to an error during `backtrace-sys` compilation on MacOS. I do not have a Mac nor MacOS at hand, so I can't really fix it.

alexcrichton:
Thanks for the PR again!

jchlapinski:
@alexcrichton sorry for the mess with libbacktrace submodule, definitely need to train myself in using git submodules. 

However the reason for this PR was so that `backtrace` crate itself tracked where the frames from the user code start. I would like to use this information as a baseline in my code, where I would like to skip some additional frames, coming from i.e. error conversion wrappers, etc. and generally polluting stacktraces after the error actually happened. In order to do that I need a way to get Backtrace::actual_start_index (quite an awful name IMO btw, perhaps we could shorted it to `start_index` ?). 

So, can we add some additional getter to Backtrace for `actual_start_index`, like `Backtrace::actual_start_index()`? 

Personally I would like best to do it "openly", by providing both getter and setter (`Backtrace::start_index(&self) -> usize` and `Backtrace::set_start_index(&mut self, ...)`, so that user can control what is displayed in the backtrace anyway he/she sees fit. 

Perhaps even `Backtrace::end_index()` and `Backtrace::set_end_index(...)` would also be usefull, as the last frames in the backtrace are also not very informative. But this can be generally achieved by converting `Backtrace` to `Vec<_>`, removing unwanted frames and converting back to `Backtrace`. 

The whole reason for this much messing around with backtrace for me is because I use them all the time for debugging, and I would much prefer browsing trough as little noisy traces as possible on a daily basis. But I do understand this might not be a priority for everyone, and simpler API is easier to maintain.  

Please let me know if additional changes to public API are possible and what would be Your preferred API to implement.

alexcrichton:
@jchlapinski oh no worries! It sounds like your use case though may be best served with the `From` constructor? It sounds like you manually want to filter the list of frames and then reconstruct a `Backtrace` object for your use case? I'm a bit wary to include too much state directly in `Backtrace` in favor of composition with vectors

jchlapinski:
@alexcrichton Sure, using `Into<Vec<_>>` for `Backtrace` suits my use case fine except that with current changes I cannot get `Backtrace::actual_start_index()` *before* converting `Backtrace` into `Vec<_>`. So in other words I can convert `Backtrace` into `Vec<BacktraceFrame>`, but I still have no easy way to know how many frames I can take from the top. Basically after convertion the value of 'actual_start_index` is lost, and it is not trivial to calculate from outside of this crate (it is platform dependent among other things)

> I'm a bit wary to include too much state directly in Backtrace in favor of composition with vectors

That is understandable and perfectly reasonable. But there is probably little harm in adding a simple getter for 'Backtrace::actual_start_index`, especially when it is useful from user perspective, as I explained above?

alexcrichton:
In your use case though don't you have extra information about what frames should be filtered? Do you *also* need the index of the first `Backtrace::new` frame?

jchlapinski:
Well, not really. I just want to skip a known number of backtrace frames - like one for converting my `MyErrorKind` into `MyError` somewhere (designed along the `std::io::Error`) and maybe another for a constructor `MyError::new()`, since those stack frames are present in the backtrace, but I just do not need them to see there (I know that code works). So basically I was just thinking of nothing more that just skipping a given number of frames at the front (and maybe also at the end), whenever I create a `Backtrace` in my code. Anything more complicated that this is probably not worth to maintain. Whenever the number of frames to be skipped changes in my code, I can adjust for that. No need for anything more fancy, IMO.

alexcrichton:
Hm but you don't need the current `actual_start_index` field for that calculation, right? You can just use one of your own functions as an anchor for figuring out what frames to prune?

jchlapinski:
Yes, but this anchoring based on function pointer is not that trivial, is it? It does not work well for Windows 32-bit, and I am not sure whether used range of 128 bytes would not maybe need to be adjusted for some other platform to avoid false detection? This code is a bit more low-level that I envisioned for a backtrace consumer and that is why I wanted it included (and guarded by tests) in this crate.

jchlapinski:
I could of course implement more reliable filtering after backtrace is resolved, based on symbols, but I wanted to avoid the need to resolve backtrace every time it is created.

jchlapinski:
@alexcrichton, btw, I really to not want to bug You too much about this. If You feel strongly against including any more API to the crate, that is fine with me, just leave this thread as it is. 

I am a big fan of Your work! Keep it up, please ;)

alexcrichton:
Hm ok, let's keep this API conservative for the time being then (as-is). It's true though yeah that resolving frames to actual functions is quite difficult, especially in the face of optimizations such logic can often be thwarted. I've found that filtering an pruning backtraces is often an uphill battle myself :(

PR #122, opened 2018-08-15
backtrace-sys: Use target_pointer_width for BACKTRACE_ELF_SIZE

cuviper:
The former code used `target.contains("64")` to detect Elf64 targets,
but this is inaccurate in a few cases:

- `s390x-unknown-linux-gnu` is 64-bit
- `sparcv9-sun-solaris` is 64-bit
- `x86_64-unknown-linux-gnux32` is 32-bit

Instead the build script can use `CARGO_CFG_TARGET_POINTER_WIDTH` to
reliably detect 64-bit targets.

### DIFF
--- a/backtrace-sys/build.rs
+++ b/backtrace-sys/build.rs
@@ -40,7 +40,8 @@ fn main() {
         build.flag("-fvisibility=hidden");
         build.file("src/libbacktrace/elf.c");
 
-        if target.contains("64") {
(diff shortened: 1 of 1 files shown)


cuviper:
This will need fixing in `rust/src/libstd/build.rs` too.  I plan to do that along with a bump to backtrace-sys, once this is merged and published.

alexcrichton:
An excellent and much better idea, thanks!

PR #123, opened 2018-08-27
Switch to libbacktrace on windows-gnu.

### LINKED ISSUE #66: Getting a lot of <no info> blocks, Windows, rustc 1.19, stable-x86_64-pc-windows-gnu
Hello, thank you for the library (and sorry if this has already been asked...I did a search on google and the issues and didn't find anything).

I've just finished implementing error_chain in my app to get access to backtraces. However, when I'm (running with `RUST_BACKTRACE=1`) printing out the backtrace, I just get a bunch of `<no info>` blocks:

```rust
match my_app() {
    Ok(..) => {}
    Err(err) => { println!("{:?}", err.backtrace()); }
}
```

```
   0:           0xa95c23 - <no info>
   1:           0xa96688 - <no info>
   2:           0xa876b2 - <no info>
   3:           0xa87762 - <no info>
   4:           0x7be522 - <no info>
   5:           0x7bebf6 - <no info>
   6:           0x758349 - <no info>
   7:           0x7aa58e - <no info>
   8:           0x79932f - <no info>
   9:           0x7a42d7 - <no info>
  10:           0x7bc078 - <no info>
  11:           0x5187cb - <no info>
  12:           0x524f1d - <no info>
  13:           0x6a417b - <no info>
  14:           0x527244 - <no info>
  15:          0x112ef09 - <no info>
  16:           0x52578f - <no info>
  17:           0x521292 - <no info>
  18:           0x524873 - <no info>
  19:           0x606590 - <no info>
  20:          0x11299d0 - <no info>
  21:     0x7ff81ae12774 - BaseThreadInitThunk
```

I'm on Windows 10, x64, rustc 1.19.0, stable-x86_64-pc-windows-gnu. Am I doing something wrong?

crlf0710:
This commit switch to using libbacktrace to windows-gnu target. To resolve #66, however, libbacktrace currently have no way to retrieve the current path, and something more needs to be done. I have tried adding something like this to `fileline.c`, and it does work:

```
#ifdef _WIN32
#ifndef HAVE_GETEXECNAME
#define HAVE_GETEXECNAME
#include <windows.h>
char* getexecname()
{
  static char g_execname[MAX_PATH + 1] = {0};
  if(!g_execname[0])
    GetModuleFileNameA(NULL, g_execname, MAX_PATH);
  return g_execname;
}

#endif
#endif
```

However i'm not sure if this is feasible, please take a look.

### DIFF
--- a/src/symbolize/libbacktrace.rs
+++ b/src/symbolize/libbacktrace.rs
@@ -12,9 +12,11 @@
 
 extern crate backtrace_sys as bt;
 
+use std::borrow::Cow;
 use libc::uintptr_t;
-use std::ffi::{CStr, OsStr};
+use std::ffi::{CStr};
 use std::os::raw::{c_void, c_char, c_int};
+#[cfg(unix)]
 use std::os::unix::prelude::*;
 use std::path::Path;
 use std::ptr;
@@ -56,13 +58,31 @@ impl Symbol {
         if pc == 0 {None} else {Some(pc as *mut _)}
     }
 
-    pub fn filename(&self) -> Option<&Path> {
+    #[cfg(unix)]
+    pub fn filename(&self) -> Option<Cow<Path>> {
+        use std::ffi::OsStr;
+
         match *self {
             Symbol::Syminfo { .. } => None,
             Symbol::Pcinfo { filename, .. } => {
-                Some(Path::new(OsStr::from_bytes(unsafe {
+                Some(Cow::Borrowed(Path::new(OsStr::from_bytes(unsafe {
                     CStr::from_ptr(filename).to_bytes()
-                })))
+                }))))
+            }
+        }
+    }
+
+    #[cfg(windows)]
+    pub fn filename(&self) -> Option<Cow<Path>> {
+        use std::path::PathBuf;
+        use std::ffi::OsString;
+
+        match *self {
+            Symbol::Syminfo { .. } => None,
+            Symbol::Pcinfo { filename, .. } => {
+                Some(Cow::Owned(PathBuf::from(OsString::from(unsafe {
+                    CStr::from_ptr(filename).to_string_lossy().to_string()
+                }))))
             }
         }
     }
--- a/src/symbolize/libbacktrace.rs
+++ b/src/symbolize/libbacktrace.rs
@@ -12,12 +12,13 @@
 
 extern crate backtrace_sys as bt;
 
-use std::borrow::Cow;
 use libc::uintptr_t;
 use std::ffi::{CStr};
 use std::os::raw::{c_void, c_char, c_int};
 #[cfg(unix)]
 use std::os::unix::prelude::*;
+#[cfg(windows)]
+use std::path::PathBuf;
 use std::path::Path;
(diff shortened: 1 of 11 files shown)


crlf0710:
Mmm, it seems the Appveyor configuration did not checkout the submodule...

src/symbolize/mod.rs
```diff
@@ -78,7 +79,7 @@ impl Symbol {
     /// unix platforms other than OSX) and when a binary is compiled with
     /// debuginfo. If neither of these conditions is met then this will likely
     /// return `None`.
-    pub fn filename(&self) -> Option<&Path> {
+    pub fn filename(&self) -> Option<Cow<Path>> {
```
alexcrichton: This is unfortunately a breaking change for this crate, but I would it be possible to avoid this perhaps? Could some unsafe code be used to store the owned version of a file, if necessary, on the windows implementation?

src/symbolize/libbacktrace.rs
```diff
@@ -56,13 +58,31 @@ impl Symbol {
...
+    #[cfg(windows)]
+    pub fn filename(&self) -> Option<Cow<Path>> {
+        use std::path::PathBuf;
+        use std::ffi::OsString;
+
+        match *self {
+            Symbol::Syminfo { .. } => None,
+            Symbol::Pcinfo { filename, .. } => {
+                Some(Cow::Owned(PathBuf::from(OsString::from(unsafe {
+                    CStr::from_ptr(filename).to_string_lossy().to_string()
```
alexcrichton: Could this check the return value of `to_string_lossy` to preserved the `Borrowed` if possible?

src/symbolize/mod.rs
```diff
@@ -251,10 +252,14 @@ cfg_if! {
 }
 
 cfg_if! {
-    if #[cfg(all(windows, feature = "dbghelp"))] {
+    if #[cfg(all(windows, target_env = "msvc", feature = "dbghelp"))] {
         mod dbghelp;
         use self::dbghelp::resolve as resolve_imp;
         use self::dbghelp::Symbol as SymbolImp;
+    } else if #[cfg(all(windows, not(target_env = "msvc"), feature = "libbacktrace"))] {
```
alexcrichton: Could this clause be added to the libbacktrace block below?

src/symbolize/mod.rs
crlf0710: Oops, i didn't realize this is for user...  to keep the original signature, i have store the OsString/PathBuf somewhere, do you think i can add an extra field in the Symbol struct for this case? Or should i collect them together as something like a global static variable?

alexcrichton: Yeah adding a field to the `Symbol` struct should be fine as that's all private implementation details anyway

crlf0710:
I'll investigate what it takes to move the get current filename path. Will update the PR when it's ready.

alexcrichton:
It's taken quite some time but I've gotten around finally now to implementing this in https://github.com/alexcrichton/backtrace-rs/pull/183

PR #126, opened 2018-10-04
Add no_std support

Aaronepower:
This is a renewed version of #50 for [rust-lang/#39503](https://github.com/rust-lang/rust/issues/39503).

cc @Mark-Simulacrum

### DIFF
--- a/src/backtrace/mod.rs
+++ b/src/backtrace/mod.rs
@@ -1,6 +1,5 @@
-use std::fmt;
-
-use std::os::raw::c_void;
+use core::fmt;
+use types::c_void;
 
 /// Inspects the current call-stack, passing all active frames into the closure
 /// provided to calculate a stack trace.
@@ -37,9 +36,19 @@ use std::os::raw::c_void;
 /// }
 /// ```
 #[inline(never)]
+#[cfg(feature = "std")]
 pub fn trace<F: FnMut(&Frame) -> bool>(mut cb: F) {
+    let _guard = ::lock::lock();
+    unsafe { trace_imp(&mut cb) }
+}
+
+/// Without std this function now does not have synchronization guarentees.
+/// Please refer to std documentation for examples and explaination.
+#[inline(never)]
+pub unsafe fn trace_unsynchronized<F: FnMut(&Frame) -> bool>(mut cb: F) {
     trace_imp(&mut cb)
 }
+
 
 /// A trait representing one frame of a backtrace, yielded to the `trace`
 /// function of this crate.
--- a/src/backtrace/mod.rs
+++ b/src/backtrace/mod.rs
@@ -22,6 +22,11 @@ use types::c_void;
 /// example, capture a backtrace to be inspected later, then the `Backtrace`
 /// type may be more appropriate.
 ///
+/// # Required features
+///
+/// This function requires the `std` feature of the `backtrace` crate to be
+/// enabled, and the `std` feature is enabled by default.
+///
 /// # Example
 ///
 /// ```
@@ -35,15 +40,18 @@ use types::c_void;
 ///     });
 /// }
 /// ```
-#[inline(never)]
+#[inline(always)]
 #[cfg(feature = "std")]
-pub fn trace<F: FnMut(&Frame) -> bool>(mut cb: F) {
+pub fn trace<F: FnMut(&Frame) -> bool>(cb: F) {
     let _guard = ::lock::lock();
-    unsafe { trace_imp(&mut cb) }
+    unsafe { trace_unsynchronized(cb) }
 }
 
-/// Without std this function now does not have synchronization guarentees.
-/// Please refer to std documentation for examples and explaination.
+/// Same as `trace`, only unsafe as it's unsynchronized.
+///
+/// This function does not have synchronization guarentees but is available
+/// when the `std` feature of this crate isn't compiled in. See the `trace`
+/// function for more documentation and examples.
 #[inline(never)]
 pub unsafe fn trace_unsynchronized<F: FnMut(&Frame) -> bool>(mut cb: F) {
     trace_imp(&mut cb)
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -46,6 +46,8 @@
 //! extern crate backtrace;
 //!
 //! fn main() {
+//! # // Unsafe here so test passes on no_std.
+//! # #[cfg(feature = "std")] {
 //!     backtrace::trace(|frame| {
 //!         let ip = frame.ip();
 //!         let symbol_address = frame.symbol_address();
@@ -63,10 +65,15 @@
 //!         true // keep going to the next frame
 //!     });
 //! }
+//! # }
 //! ```
 
 #![doc(html_root_url = "https://docs.rs/backtrace")]
 #![deny(missing_docs)]
+#![no_std]
+
+#[cfg(feature = "std")]
+#[macro_use] extern crate std;
 
 #[cfg(unix)]
 extern crate libc;
@@ -98,18 +105,27 @@ cfg_if! {
 }
 
 #[allow(dead_code)] // not used everywhere
-#[cfg(unix)]
+#[cfg(all(unix, feature = "std"))]
 #[macro_use]
 mod dylib;
 
-pub use backtrace::{trace, Frame};
+pub use backtrace::{trace_unsynchronized, Frame};
 mod backtrace;
 
-pub use symbolize::{resolve, Symbol, SymbolName};
+pub use symbolize::{resolve_unsynchronized, Symbol, SymbolName};
 mod symbolize;
 
-pub use capture::{Backtrace, BacktraceFrame, BacktraceSymbol};
-mod capture;
+pub use types::BytesOrWideString;
+mod types;
+
+cfg_if! {
+    if #[cfg(feature = "std")] {
+        pub use backtrace::trace;
+        pub use symbolize::resolve;
+        pub use capture::{Backtrace, BacktraceFrame, BacktraceSymbol};
+        mod capture;
+    }
+}
 
 #[allow(dead_code)]
 struct Bomb {
@@ -126,8 +142,10 @@ impl Drop for Bomb {
 }
 
 #[allow(dead_code)]
+#[cfg(feature = "std")]
 mod lock {
     use std::cell::Cell;
+    use std::boxed::Box;
     use std::sync::{Once, Mutex, MutexGuard, ONCE_INIT};
 
     pub struct LockGuard(MutexGuard<'static, ()>);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -77,7 +77,8 @@
 
 #[cfg(unix)]
 extern crate libc;
-#[cfg(all(windows, feature = "winapi"))] extern crate winapi;
+#[cfg(windows)]
+extern crate winapi;
 
 #[cfg(feature = "serde_derive")]
 #[cfg_attr(feature = "serde_derive", macro_use)]
--- a/src/symbolize/dbghelp.rs
+++ b/src/symbolize/dbghelp.rs
@@ -10,11 +10,16 @@
 
 #![allow(bad_style)]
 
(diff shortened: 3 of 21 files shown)


Aaronepower:
I've fixed the problems on linux. The failures in the CI seem unrelated to the PR.

alexcrichton:
Thanks for the PR! I fear though that the strategy of sprinkling `feature = "std"` wherever necessary makes the code pretty unmaintainable and harder to read. I think there's perhaps some simple fixes though?

* The entire `capture` module should probably be removed in `no_std` mode, only the bare-bones stack-based API is probably all we need to provide.
* With locking not being possible, I think that some APIs need to be `unsafe`? I believe both libbacktrace and windows need to be synchronized to work safely
* I think that we'll maybe want to pick a different abstraction for filenames if `Path` isn't available, because almost all platforms still want a file name one way or another
* I think we still want `Debug` impls in no_std mode?
* Getting the name of a symbol is pretty important, so I think we'll want a solution for that in `no_std` mode as well

Aaronepower:
@alexcrichton Thanks for the feedback, I currently have a in development version that addresses most of this. However I wanted to ask you what kind of abstraction were you thinking for the `Path` alternatives? We could have it return a `&str`? We could make this easily maintable by having a `type BacktracePath = Path;` on std and on core we have `type BacktracePath<'a> = &'a str;`?

alexcrichton:
I think Windows may be a stickler where the information is actually `&[u16]` instead of `&[u8]`, so we can probably just add a custom enum that's something like `ByteOrWideString` and has convenience accessors when std is enabled

Aaronepower:
@alexcrichton Okay, I've updated the PR with your feeback. Please let know if there's anything else required!

src/types.rs
```diff
@@ -0,0 +1,51 @@
...
+
+        match self {
+            Bytes(slice) => String::from_utf8_lossy(slice),
+            Wide(wide) => Cow::Owned(String::from_utf16_lossy(wide)),
+        }
+    }
+
+    /// Provides a `Path` representation of `BytesOrWideString`.
+    pub fn into_path_buf(self) -> PathBuf {
+        PathBuf::from(&*self.as_str_lossy())
```
alexcrichton: This should be an os-specific implementation using the extension traits on Windows while assering that `Wide` doesn't happen on Unix (to ensure this is a lossless conversion)

src/types.rs
```diff
@@ -0,0 +1,51 @@
...
+    Bytes(&'a [u8]),
+    /// Wide strings typically from Windows.
+    Wide(&'a [u16]),
+}
+
+#[cfg(feature = "std")]
+impl<'a> BytesOrWideString<'a> {
+    /// Lossy converts to a `Cow<str>`, will allocate if `Bytes` is not valid
+    /// UTF-8 or if `BytesOrWideString` is `Wide`.
+    pub fn as_str_lossy(&self) -> Cow<'a, str> {
```
alexcrichton: I think we'll probably want to call this `to_string_lossy` because `as` implies cheap but this is an expensive operation.

src/symbolize/mod.rs
```diff
@@ -78,7 +84,7 @@ impl Symbol {
     /// unix platforms other than OSX) and when a binary is compiled with
     /// debuginfo. If neither of these conditions is met then this will likely
     /// return `None`.
-    pub fn filename(&self) -> Option<&Path> {
+    pub fn filename(&self) -> Option<BytesOrWideString> {
```
alexcrichton: I'd personally prefer to not have a breaking change to this library right now, so could this method be left as-is but a new API added along the lines of `filename_raw`?

src/symbolize/mod.rs
```diff
@@ -188,6 +196,26 @@ impl<'a> SymbolName<'a> {
...
+    fmt(&*String::from_utf8_lossy(bytes), f)
+}
+
+#[cfg(not(feature = "std"))]
+fn format_symbol_name(fmt: fn(&str, &mut fmt::Formatter) -> fmt::Result,
+                      bytes: &[u8],
+                      f: &mut fmt::Formatter)
+    -> fmt::Result
+{
+    fmt(str::from_utf8(bytes).unwrap(), f)
```
alexcrichton: I don't think we'll want to implement this here with `unwrap()` for core, we'll want the same logic for both std/core which is to attempt to interpret as utf-8 but otherwise never panic.

(for example panicking definitely means we can't integrate this later into libstd

src/symbolize/mod.rs
```diff
@@ -36,10 +33,19 @@ use rustc_demangle::{try_demangle, Demangle};
...
 /// ```
+#[cfg(feature = "std")]
+pub fn resolve<F: FnMut(&Symbol)>(addr: *mut c_void, mut cb: F) {
+    resolve_imp(addr as *mut _, &mut cb)
+}
+
+/// Without std this function now does not have synchronization guarentees.
+/// Please refer to std documentation for examples and explaination.
+#[cfg(not(feature = "std"))]
 pub fn resolve<F: FnMut(&Symbol)>(addr: *mut c_void, mut cb: F) {
```
alexcrichton: I think this forgot `unsafe`, but I'd personally prefer that we don't switch the signature of a function based on features enabled because that can mean it's a breaking change to activate a feature.

Could a new API be added, something like `resolve_unsynchronized`, which is `unsafe` and requires external synchronization? The safe `resolve` would then acquire a lock and call the raw API.

I believe this change would entail moving all locking from the inner modules to this function itself, but I think that's fine.

src/lib.rs
```diff
@@ -108,9 +115,13 @@ mod backtrace;
 pub use symbolize::{resolve, Symbol, SymbolName};
 mod symbolize;
 
+#[cfg(feature = "std")]
 pub use capture::{Backtrace, BacktraceFrame, BacktraceSymbol};
+#[cfg(feature = "std")]
 mod capture;
 
+mod types;
```
alexcrichton: I thin kwe'll want to be sure to reexport the enum in this modules at the top-level as well

src/backtrace/mod.rs
```diff
@@ -37,10 +36,20 @@ use std::os::raw::c_void;
...
+#[cfg(feature = "std")]
 pub fn trace<F: FnMut(&Frame) -> bool>(mut cb: F) {
     trace_imp(&mut cb)
 }
 
+/// Without std this function now does not have synchronization guarentees.
+/// Please refer to std documentation for examples and explaination.
+#[inline(never)]
+#[cfg(not(feature = "std"))]
+pub unsafe fn trace<F: FnMut(&Frame) -> bool>(mut cb: F) {
```
alexcrichton: Similar comments to `resolve` below in that we won't want to switch signatures based on features, but rather just not provide the safe API in `std`

src/lib.rs
```diff
@@ -63,10 +65,15 @@
...
 //! }
+//! # }
 //! ```
 
 #![doc(html_root_url = "https://docs.rs/backtrace")]
 #![deny(missing_docs)]
+#![cfg_attr(not(feature = "std"), no_std)]
+
+#[cfg(not(feature = "std"))]
+extern crate core as std;
```
alexcrichton: FWIW my preferred way at least to add `no_std` support is to not do this, but rather unconditionally use `#![no_std]` and then conditionally `extern crate std` whenever `std` is enabled. Then, when necessary, the `std::prelude::v1::*` module is imported

src/symbolize/dbghelp.rs
```diff
@@ -23,96 +28,132 @@ use winapi::um::dbghelp;
...
 
 use SymbolName;
+use types::BytesOrWideString;
 
+// Store an OsString on std so we can provide the symbol name and filename.
 pub struct Symbol {
+    #[cfg(feature = "std")]
     name: OsString,
+    #[cfg(not(feature = "std"))]
+    name: (*const u16, usize),
```
alexcrichton: I think by default regardless of the `std` feature we'll want to store this pair here. When `std` is activated the `name` only exists for the signature to work (returning `&Path` instead of `PathBuf`), so I think it's fine to, only in `std` mode, have the `name: OsString` field be something like `name_cache: OsString`

src/symbolize/dbghelp.rs
```diff
@@ -23,96 +28,132 @@ use winapi::um::dbghelp;
...
+    // due to struct alignment.
+    info.SizeOfStruct = 88;
+
+    let _c = ::dbghelp_init();
+
+    let mut displacement = 0u64;
+    let ret = dbghelp::SymFromAddrW(processthreadsapi::GetCurrentProcess(),
+    addr as DWORD64,
+    &mut displacement,
+    info);
```
alexcrichton: The indentation here look like it got interesting?

src/symbolize/mod.rs
```diff
@@ -69,26 +79,52 @@ impl Symbol {
...
     pub fn filename(&self) -> Option<&Path> {
-        self.inner.filename()
+        use std::ffi::OsStr;
+        use std::os::unix::ffi::OsStrExt;
+
+        match self.filename_raw() {
+            Some(BytesOrWideString::Bytes(slice)) => {
+                Some(Path::new(OsStr::from_bytes(slice)))
+            }
+            _ => None,
```
alexcrichton: Could this panic because it's never supposed to happen?

alexcrichton: Er or rather, panic on wide strings?

src/symbolize/mod.rs
```diff
@@ -69,26 +79,52 @@ impl Symbol {
...
-    /// Returns the line number for where this symbol is currently executing.
-    ///
-    /// This return value is typically `Some` if `filename` returns `Some`, and
-    /// is consequently subject to similar caveats.
-    pub fn lineno(&self) -> Option<u32> {
-        self.inner.lineno()
+    /// Returns the file name where this function was defined.
+    #[cfg(windows)]
+    pub fn filename(&self) -> Option<&Path> {
+        self.inner.filename().map(Path::new)
```
alexcrichton: This is definitely a bummer, but unfortunately I don't know of a better way to do it! Next time we'll for sure make this return a `PathBuf`

src/symbolize/mod.rs
```diff
@@ -188,6 +229,30 @@ impl<'a> SymbolName<'a> {
...
+    fmt(&*String::from_utf8_lossy(bytes), f)
+}
+
+#[cfg(not(feature = "std"))]
+fn format_symbol_name(fmt: fn(&str, &mut fmt::Formatter) -> fmt::Result,
+                      bytes: &[u8],
+                      f: &mut fmt::Formatter)
+    -> fmt::Result
+{
+    if let Ok(name) = str::from_utf8(bytes) {
```
alexcrichton: I think it's possible to use the `Err` here with [`error_len`](https://doc.rust-lang.org/std/str/struct.Utf8Error.html#method.error_len) to rig up something roughly equivalent to `from_utf8_lossy` to allow printing both and outside of std?

alexcrichton:
I've been working through this but unfortunately ran out of time for the near future. I don't mind picking this up from here, but some things I've noticed are:

* There's a number of warnings across various platforms to squelch
* Windows and `--no-default-features` is having issues, needs to unconditionally depend on `winapi`, and needs to remove `"std"` from `winapi`. Additionally need to be added to CI
* Windows uses `Box::new` but can use `#[repr(align)]` instead now
* APIs need to be documented now that they require the `std` feature
* `resolve` and `trace` should both be implemented in terms of `*_unsynchronized`
* Overall the `BytesOrWideString` is still somewhat uncomfortable on Windows, but I think it's workable still? We won't want to shut off things on windows entirely, we'll want some sort of fallback that still works. I think this means some utf-16 decoder may be needed
* The printing logic for symbol names still needs to work the same way in/out of std

alexcrichton:
@Aaronepower oh that's probably because the test is quite brittle and some constants in there need to change, although if that doesn't get it passing I can help dig in too

alexcrichton:
Ok I've pushed up all the changes that I think I'd like to see, now waiting on CI

PR #131, opened 2018-11-02
exclude CI files

ignatenkobrain:


alexcrichton:
Like https://github.com/alexcrichton/cc-rs/pull/358, can a description be added here as well?

Issue #135, opened 2018-12-11
Does not compile with Rust 1.18


RalfJung:
The [failure crate](https://github.com/rust-lang-nursery/failure), which depends on this, supports Rust versions as far back as Rust 1.18.  This crate seems to only test on CI against current stable.  It is hence not entirely surprising that [backtrace currently fails to build against Rust 1.18](https://travis-ci.org/rust-lang-nursery/failure/jobs/466672292).

RalfJung:
This could be resolved by bumping the minimal Rust version required by failure, or by fixing backtrace-rs -- but either way it seems the two crates should sync their CI in terms of which baseline Rust version they test.

alexcrichton:
I'm not personally invested in making crates compile on such old versions of rustc, but I'm happy to merge PRs to fix compatibility!

alexcrichton:
I'm gonna close this with #137 merged!

PR #137, opened 2018-12-12
Maintain compatibility with Rust 1.25

azriel91:
Issue #135

Not sure if you want to also include a CI config for Rust 1.18

### DIFF
--- a/src/symbolize/dbghelp.rs
+++ b/src/symbolize/dbghelp.rs
@@ -10,13 +10,18 @@
 
 #![allow(bad_style)]
 
+// This is a hack for compatibility with rustc 1.25.0. The no_std mode of this
+// crate is not supported pre-1.30.0, but in std mode the `char` module here
+// moved in rustc 1.26.0 (ish). As a result, in std mode we use `std::char` to
+// retain compatibility with rustc 1.25.0, but in `no_std` mode (which is
+// 1.30.0+ already) we use `core::char`.
 #[cfg(feature = "std")]
 use std::char;
+#[cfg(not(feature = "std"))]
+use core::char;
 
 use core::mem;
 use core::slice;
-#[cfg(not(feature = "std"))]
-use core::char;
 
 use winapi::ctypes::*;
 use winapi::shared::basetsd::*;
--- a/src/symbolize/dbghelp.rs
+++ b/src/symbolize/dbghelp.rs
@@ -10,8 +10,12 @@
 
 #![allow(bad_style)]
 
+#[cfg(feature = "std")]
+use std::char;
+
 use core::mem;
 use core::slice;
+#[cfg(not(feature = "std"))]
 use core::char;
 
 use winapi::ctypes::*;
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,10 @@ matrix:
     - rust: nightly
       script: *test_all
 
+    # Test everything on Rust 1.25.0 for backward compatibility
+    - rust: 1.25.0
+      script: *test_all
+
     # Upload docs on nightly
     - rust: nightly
       script:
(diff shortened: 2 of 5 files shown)


azriel91:
Implementation is incorrect for detecting rust features, shall find another way.

azriel91:
Ah, 1.18.0 built on windows, but not linux. I'm off for tonight, but feel free to add to the branch if it's urgent.

alexcrichton:
This is unfortunately a bit too much code duplication for me to be comfortable maintaining purely for the reason of supporting older compilers. Is there a compelling reason to support 1.18.0?

azriel91:
Hm, besides "make the `failure` PR pass", I think that's better answered by @burntsushi

BurntSushi:
I've honestly lost a lot of my own motivation for tracking older Rust releases. It is too much work, and in particular, is completely intractable without community consensus on what to do. I think we should at the very minimum put a specific Rust version in our CI config so that folks can at least know which version is supported though.

RalfJung:
`error_len` is stable since Rust 1.20.  Assuming that this removes most of the need for code duplication, would it be okay for failure to bump its minimal version to 1.20?

alexcrichton:
Heh to add to what @BurntSushi says I've also lost motivation to even list the minimum Rust version in CI, all it ends up doing for me is making it annoying to land PRs that bump it (as you always forget about it) and also confusing contributors who aren't typically aware that it can be bumped. 

It looks like the main duplication here is happening at the rust 1.25.0 boundary, so it seems fine to me to land whatever's necessary to support 1.25.0+ as it likely doesn't involve much duplication

RalfJung:
Oh dang I missed the `repr(align)` thing and that it requires Rust 1.25.

azriel91:
Okay, so with 1.25.0 as the minimum, things are much easier!
Actually many times when Rust releases a new stable version, I think "wow, this should be the minimum stable" 😅.

src/symbolize/dbghelp.rs
```diff
@@ -10,8 +10,12 @@
 
 #![allow(bad_style)]
 
+#[cfg(feature = "std")]
+use std::char;
+
 use core::mem;
 use core::slice;
+#[cfg(not(feature = "std"))]
```
alexcrichton: Hm what's going on here? Was this something we moved from std to core in the last few versions?

alexcrichton:
Looks great to me! Just one thing I'm curious about, and it also looks like CI may be failing for 1.25.0? (I don't think std has `c_void` by that point)

src/symbolize/dbghelp.rs
azriel91: Yeap, based on issue https://github.com/rust-lang/rust/issues/49319 and the corresponding PR: https://github.com/rust-lang/rust/pull/49698 which happened on 2018-04-12, which is *just* after Rust 1.26 was [released](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1260-2018-05-10).

azriel91:
Yeap, took longer than I'd've liked to figure out that `no_std` support isn't available until Rust 1.30. I've updated the CI config to test the features with `std`, though it duplicates nearly all of the main script. Yaml doesn't have a standard merge syntax (probably for good reason).

azriel91:
So, the `"gimli-symbolize"` feature transitively uses `parity-wasm` (backtrace -> addr2line 0.7 -> object 0.9 -> parity-wasm 0.31.3).

`parity-wasm` makes use of [auto dereferencing match patterns](https://github.com/rust-lang/rust/pull/49394) [introduced in Rust 1.26](https://github.com/rust-lang/rust/blob/stable/RELEASES.md#version-1260-2018-05-10).

Road from here, things that I'd consider unfeasible (maintenance cost, sanity):

* Chase down every crate and "make it backward compatible"
* Track down Rust compatibility for every dependency and introduce Rust version checks for each.

I've removed it from the 1.25 CI tests.

alexcrichton:
Thanks so much!

PR #143, opened 2018-12-13
Error on Mac while building the docker image

harishbm34:
Expected enum `std::os::raw::c_void`, found enum `libc::c_void` in libbacktrace.rs:156
Expected enum `std::os::raw::c_void`, found enum `libc::c_void` in libbacktrace.rs:172
Expected enum `std::os::raw::c_void`, found enum `libc::c_void` in libbacktrace.rs:176

### DIFF
--- a/src/symbolize/libbacktrace.rs
+++ b/src/symbolize/libbacktrace.rs
@@ -16,7 +16,8 @@ use std::ffi::CStr;
 use std::{ptr, slice};
 use std::sync::{ONCE_INIT, Once};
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! This, however I don't think should be necessary and likely indicates that there's a bug somewhere else. If you're using a supported compiler I don't think this error happens, or at least I can't reproduce it.

Do you perhaps have particular features active and/or inactive for the backtrace crate?

PR #166, opened 2019-04-24
Enable the default panic unwind hook to print backtraces again

### LINKED ISSUE #165: Capturing a backtrace prevents any future panic from printing a backtrace on Windows
As the title states, backtraces are not printed if the program panics after a backtrace was captured with the library.

```
$ rustc --version -v
rustc 1.34.0 (91856ed52 2019-04-10)
binary: rustc
commit-hash: 91856ed52c58aa5ba66a015354d1cc69e9779bdf
commit-date: 2019-04-10
host: x86_64-pc-windows-msvc
release: 1.34.0
LLVM version: 8.0
```

With `RUST_BACKTRACE=1`

```rust
let bt = Backtrace::new();
println!("{:?}", bt); // <-- this backtrace is printed
panic!("Oops!");      // <-- the backtrace here is not printed
```

I tracked it down to this commit: https://github.com/alexcrichton/backtrace-rs/commit/3cfb76a912c666d94eea5d07afd20b83283cbe35

The cleanup code was removed. 

Changing the example to the following allows for both backtraces to print:
```rust
let bt = Backtrace::new();
println!("{:?}", bt);
unsafe {
    let handle = winapi::um::processthreadsapi::GetCurrentProcess();
    winapi::um::dbghelp::SymCleanup(handle);
}
panic!("Oops!");
```

The runtime panic-unwind code is checking the result of `SymInitialize` which is probably an error due to double initialization. 

https://github.com/rust-lang/rust/blob/9ebf47851a357faa4cd97f4b1dc7835f6376e639/src/libstd/sys/windows/backtrace/mod.rs#L74

https://docs.microsoft.com/en-us/windows/desktop/api/dbghelp/nf-dbghelp-syminitializew

> A process that calls SymInitialize should not call it again unless it calls SymCleanup first. 

Perhaps the old behavior could be restored with deferred symbol loading:

https://docs.microsoft.com/en-us/windows/desktop/api/Dbghelp/nf-dbghelp-symsetoptions

> SYMOPT_DEFERRED_LOADS0x00000004 | Symbols are not loaded until a reference is made requiring the symbols be loaded. This is the fastest, most efficient way to use the symbol handler.

Alternatively, the runtime panic-unwind code could perhaps always call `SymCleanup` first (and ignore the error) prior to calling `SymInitialize`. As far as I can tell, calling `SymCleanup` without first initializing returns an error code but is otherwise harmless.

aloucks:
The default panic unwind hook attempts to print the backtrace by
first initializing DbgHelp with `SymInitializeW`. The documentation
states that this can't be called again until after `SymCleanup` is
called first. If called prior to cleanup, it returns a non-success
result. The default unwind hook detects this error and never tries
to print the backtrace.

This PR installs a new panic handler that first calls `SymCleanup`
prior to calling the previously installed (or default) panic hook.

Fixes #165

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -187,10 +187,33 @@ unsafe fn dbghelp_init() {
 
     static mut INITIALIZED: bool = false;
 
+    #[cfg(feature = "std")]
+    fn dbghelp_init_cleanup() {
+        static SET_HOOK: std::sync::Once = std::sync::ONCE_INIT;
+
+        SET_HOOK.call_once(|| {
+            let default_hook = std::panic::take_hook();
+            std::panic::set_hook(std::boxed::Box::new(move |panic_info| {
+                // The default unwind hook calls `SymInitializeW`, which will fail and prevent
+                // the panic backtrace from being printed if we do not first call `SymCleanup`.
+                unsafe {
+                    INITIALIZED = false;
+                    dbghelp::SymCleanup(processthreadsapi::GetCurrentProcess());
+                }
+                default_hook(panic_info);
+            }))
+        })
+    }
+
+    #[cfg(not(feature = "std"))]
+    fn dbghelp_init_cleanup() {}
+
     if !INITIALIZED {
+        INITIALIZED = true;
+
         dbghelp::SymInitializeW(processthreadsapi::GetCurrentProcess(),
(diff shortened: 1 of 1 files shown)


aloucks:
Superseded by #172

PR #184, opened 2019-05-14
(Pre)Add CONTEXT definition and init_frame for ARM

chandde:
This is part of adding ARM support for rust on Windows. Another PR for winapi is on going, https://github.com/retep998/winapi-rs/pull/762, ideally I'd finish that one then work on this since there are similar changes for the CONTEXT struct for ARM, however I'd like this send the PR now so the changes can be upstreamed early, the fact is I'm perhaps the only one building rust for arm Windows, and this change is well guarded by "arm", others should never run into these at all. Once the winapi PR is merged and a new version is released, we'll adapt the new version, and bump the winapi version here in backtrace cargo.toml if necessary.


chandde:
@alexcrichton can you kindly help resolving the merge conflicts? thanks! the "Resolve conflicts" button is disabled for me. I have no idea what conflict is, I don't see anything in the diff view.

alexcrichton:
Oh the file just moved to `src/windows.rs`, so the changes should be applicable to there instead

chandde:
> Oh the file just moved to `src/windows.rs`, so the changes should be applicable to there instead

Let me see what I can do, perhaps send a new PR

chandde:
OK, now there is no merge conflict once I merged the upstream!

alexcrichton:
Looks great! If possible it'd be great to circle back when things are set up in rust-lang/rust so we can add a CI builder for arm as well

chandde:
Thanks, I'm.working on getting a nightly build for rust arm32.

Sent from Outlook
________________________________

PR #193, opened 2019-05-29
Add UWP support

chouquette:
Hi,

This PR is associated with https://github.com/rust-lang/rust/pull/60260, and aims at fixing the build when targeting UWP.
I'd have preferred not to revert "Use LoadLibraryA instead of LoadLibraryW" but failed at using convert_utf16 to avoid declaring the byte array manually. I suppose it makes sense to not be able to use `std::` while compiling a dependency of `libstd`
The `GetModuleFileNameExA` patch requires https://sourceforge.net/p/mingw-w64/mailman/message/36680273/ which will hopefully be applied soon

### DIFF
--- a/src/backtrace/mod.rs
+++ b/src/backtrace/mod.rs
@@ -137,7 +137,7 @@ cfg_if! {
         mod unix_backtrace;
         use self::unix_backtrace::trace as trace_imp;
         pub(crate) use self::unix_backtrace::Frame as FrameImp;
-    } else if #[cfg(all(windows, feature = "dbghelp"))] {
+    } else if #[cfg(all(windows, feature = "dbghelp", not(target_vendor = "uwp")))] {
         mod dbghelp;
         use self::dbghelp::trace as trace_imp;
         pub(crate) use self::dbghelp::Frame as FrameImp;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -171,7 +171,7 @@ mod lock {
     }
 }
 
-#[cfg(all(windows, feature = "dbghelp"))]
+#[cfg(all(windows, feature = "dbghelp", not(target_vendor = "uwp")))]
 mod dbghelp;
 #[cfg(windows)]
 mod windows;
--- a/src/symbolize/mod.rs
+++ b/src/symbolize/mod.rs
@@ -428,7 +428,7 @@ cfg_if! {
         use self::coresymbolication::resolve as resolve_imp;
         use self::coresymbolication::Symbol as SymbolImp;
     } else if #[cfg(all(feature = "libbacktrace",
-                        any(unix, all(windows, target_env = "gnu")),
+                        any(unix, all(windows, target_vendor = "pc", target_env = "gnu")),
                         not(target_os = "fuchsia"),
                         not(target_os = "emscripten")))] {
         mod libbacktrace;


chouquette:
It seems I overlooked quite a few parts of this PR. Sorry about that, I'll fix it

src/symbolize/libbacktrace.rs
```diff
@@ -347,6 +347,28 @@ unsafe fn init_state() -> *mut bt::backtrace_state {
...
+            unsafe fn create_file(name: &[i8]) -> HANDLE {
+                CreateFile2(name.as_ptr(),
+                            GENERIC_READ,
+                            FILE_SHARE_READ | FILE_SHARE_WRITE,
+                            OPEN_EXISTING,
+                            ptr::null_mut()
+                )
+            }
+
+            #[cfg(target_vendor = "pc")]
```
alexcrichton: FWIW this I think would probably be best as `#[cfg(not(...))]` instead of `target_vendor = "pc"`

src/symbolize/libbacktrace.rs
```diff
@@ -381,7 +395,7 @@ unsafe fn init_state() -> *mut bt::backtrace_state {
             unsafe fn query_full_name(buf: &mut [i8]) -> Result<&[i8], ()> {
                 let p1 = OpenProcess(PROCESS_QUERY_INFORMATION, FALSE, GetCurrentProcessId());
                 let mut len = buf.len() as u32;
-                let rc = QueryFullProcessImageNameA(p1, 0, buf.as_mut_ptr(), &mut len);
+                let rc = GetModuleFileNameExA(p1, ptr::null_mut(), buf.as_mut_ptr(), &mut len);
```
alexcrichton: To confirm, this function has similar portability and version compatibility as the previous?

src/windows.rs
```diff
@@ -339,7 +339,7 @@ ffi! {
         pub fn GetCurrentProcess() -> HANDLE;
         pub fn GetCurrentThread() -> HANDLE;
         pub fn RtlCaptureContext(ContextRecord: PCONTEXT) -> ();
-        pub fn LoadLibraryA(a: *const i8) -> HMODULE;
```
alexcrichton: Is the A version not supported on UWP?

src/windows.rs
```diff
@@ -578,7 +586,7 @@ ffi! {
 
     pub const ARM_MAX_BREAKPOINTS: usize = 8;
     pub const ARM_MAX_WATCHPOINTS: usize = 1;
-
+
```
alexcrichton: Looks like some stray whitespace here and above?

src/symbolize/libbacktrace.rs
chouquette: `QueryFullProcessImageNameA` is available starting from Windows Vista and isn't available on UWP
`GetModuleFileNameExA` is available starting from Windows XP, and is available on UWP
So switching would definitely not impact any targeted users.

src/windows.rs
chouquette: It isn't, but I just realized that I was bitten by `libwinstorecompat` once again. LoadLibraryW isn't available either, and `LoadPackagedLibrary` needs to be used instead. In any case, that function also takes wide strings though

chouquette: But that means that dbghelper must be packaged in the app... which is probably not an option.

I'm afraid the only way to have this built on UWP is to disable it :/

src/windows.rs
chouquette: Oops, will fix

alexcrichton:
@chouquette to confirm, is it intentional that this effectively disables backtraces on uwp?

chouquette:
Sadly yes. It's forbidden to load a DLL outside of the application package on UWP, so dbghelp can't be loaded in any case. The function provided by that DLL are not available on UWP, so we can't link with it either in order to avoid the LoadLibrary calls.

alexcrichton:
Ok just wanted to make sure!

PR #199, opened 2019-06-03
Include -1 logic on Unix for libbacktrace

alexcrichton:
After tweaking where all the `-1` annotations are located this is
hopefully the final central location they need to be in... In any case
this also imports the test from libstd which is typically failing
upstream so we can get earlier signals to failures here.


alexcrichton:
Oops, accidentally pushed up as https://github.com/rust-lang/backtrace-rs/commit/2e1d85f25679f3865f83deeaba72e7fba918809e

PR #217, opened 2019-06-27
More improvements to gimli support

alexcrichton:
* Migrate to `goblin` instead of `object` to reduce dependencies
* Disable as many features as we can along the way, slimming down dependencies
* Add rudimentary support for Windows MinGW
* Don't consult symbol tables in files and instead assume `dladdr` does
  that for us
* Add a lifetime parameter to the inner `Symbol` so the gimli bindings
  can be expressed a bit more safely.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,12 +30,18 @@ rustc-serialize = { version = "0.3", optional = true }
 cpp_demangle = { default-features = false, version = "0.2.3", optional = true }
 
 # Optional dependencies enabled through the `gimli-symbolize` feature
-addr2line = { version = "0.9.0", optional = true, default-features = false, features = ['std', 'std-object'] }
+addr2line = { version = "0.9.0", optional = true, default-features = false, features = ['std'] }
 findshlibs = { version = "0.5.0", optional = true }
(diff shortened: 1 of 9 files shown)


philipc:
> Don't consult symbol tables in files and instead assume dladdr does that for us

I think that will only use `.dynsym` and not `.symtab`. On the other hand, I'm not sure how often `.symtab` is present but debuginfo isn't.

alexcrichton:
Yeah I've started seeing a few more `<unknown>` symbols than before so I think you're right. Ideally though I'd like to make sure there's a test asserting that it needs to be there, and just working on that coming about!

alexcrichton:
Ok tests are indeed failing right now when debuginfo is missing [but due to other reasons](https://github.com/rust-lang/backtrace-rs/blob/400ff0b5d92ca9a51c3bc3926ad84ebb9d9be069/ci/azure-test-all.yml#L56-L59) they're already disabled. I'll work on fixing that once 1.36.0 is out!

alexcrichton:
Ok I've gotten around now to [implementing this again](https://github.com/rust-lang/backtrace-rs/pull/224)

Issue #222, opened 2019-07-07
libbacktrace allocates ~131 mb the first time `libbacktrace::resolve()` is called and never frees it


Tomcc:
I'm not sure if this is an issue in general, but it definitely was for me :)

tl;dr: I would like a way to tell `backtrace` to NOT keep its state around and allocate everything from scratch every time. I'm up for writing a PR myself!

long version:

I had an issue with a couple Rust services killing my low-mem (1gb) linux VMs by causing it to start swapping permanently, seemingly at random.
It turned out that the unassuming 30mb RAM processes could randomly grow permanently to 170mb or more, causing the VM to run out of memory.

I managed to repro the sequence of actions on Mac OS, so that let me trace the allocations with XCode Instruments, which is a pretty neat tool, and... it turns out the culprit is this crate!

<img width="1095" alt="Screen Shot 2019-07-07 at 8 49 00 AM" src="https://user-images.githubusercontent.com/553214/60769355-37021c80-a094-11e9-8b04-d48da68daf57.png">

Around the 1:40 mark, the program logged its first `failure::Error`, causing a backtrace resolve, causing 131 mb of persistent allocations. All the allocation in that image come from backtrace. For example
<img width="1677" alt="Screen Shot 2019-07-07 at 9 04 47 AM" src="https://user-images.githubusercontent.com/553214/60769513-62860680-a096-11e9-80a6-2fae209d2771.png">

The allocations never go away after that. Each new `resolve()` may allocate more tiny objects but never spikes as much as the first time.

My suspect is https://github.com/rust-lang/backtrace-rs/blob/master/src/symbolize/libbacktrace.rs#L260
The state is `static`, so it's probably what's keeping all those buffers forever alive. 
I assume that allocating a hundred mb or two forever may be worth the performance benefit on beefier machines, but in memory-constrained systems it's a problem.

My crappy 1gb VM would survive swapping until the trace is done (errors are very rare) but it can't survive keeping all that stuff around forever.
This is matters to bigger VMs too... for example, if you tried hosting dozens of containers on the same VM, they would slowly balloon in memory size as they encounter their first error, but it's super unlikely that they would all try to resolve an error at once.

I thought of just making it a #[cfg] option and submitting a PR, but I wanted to know if you think this makes sense to address and how you'd do it :)

If you want to see the original trace and have XCode, here it is:
[backtrace.trace.zip](https://github.com/rust-lang/backtrace-rs/files/3365694/backtrace.trace.zip)

[...]

Issue #225, opened 2019-07-09
Make the gimli cache global, not thread local


alexcrichton:
All our functions are externally synchronized so there's no need for this cache to be thread-local, we can just persist it globally like all other platforms.

alexcrichton:
Done in https://github.com/rust-lang/backtrace-rs/pull/226

PR #226, opened 2019-07-11
Adding a method to clear the Gimli symbol cache

Tomcc:
As discussed in #222 , in my use case I can't really afford each process to lock hundreds of mb's per thread in the `backtrace` symbol cache.

I've added a `clear_symbol_cache()` function to the root of the crate that clears the Gimli mapping cache and uh, does nothing at all if you aren't using Gimli.
I tried to fix libbacktrace as well, but that one doesn't expose a way to free its state even in C.

I'm not that good at Rust still so this PR is more of a prototype, let me know if anything is terrible :) 
I know that a few things could be improved:
- Is it fine to take a dependency on `lazy_static`? 
- The Mutex on the static is redundant, but I couldn't placate the compiler.
- Defining `fn noop_clear_symbol_cache() {}` again for each backend is ugly.

Anyway, there it is.

### DIFF
--- a/src/symbolize/gimli.rs
+++ b/src/symbolize/gimli.rs
@@ -329,57 +329,57 @@ impl Mapping {
     }
 }
 
-thread_local! {
-    // A very small, very simple LRU cache for debug info mappings.
-    //
-    // The hit rate should be very high, since the typical stack doesn't cross
-    // between many shared libraries.
-    //
-    // The `addr2line::Context` structures are pretty expensive to create. Its
-    // cost is expected to be amortized by subsequent `locate` queries, which
-    // leverage the structures built when constructing `addr2line::Context`s to
-    // get nice speedups. If we didn't have this cache, that amortization would
-    // never happen, and symbolicating backtraces would be ssssllllooooowwww.
-    static MAPPINGS_CACHE: RefCell<Vec<(PathBuf, Mapping)>>
-        = RefCell::new(Vec::with_capacity(MAPPINGS_CACHE_SIZE));
+// A very small, very simple LRU cache for debug info mappings.
+//
+// The hit rate should be very high, since the typical stack doesn't cross
+// between many shared libraries.
+//
+// The `addr2line::Context` structures are pretty expensive to create. Its
+// cost is expected to be amortized by subsequent `locate` queries, which
+// leverage the structures built when constructing `addr2line::Context`s to
+// get nice speedups. If we didn't have this cache, that amortization would
+// never happen, and symbolicating backtraces would be ssssllllooooowwww.
+static MAPPINGS_CACHE: RefCell<Vec<(PathBuf, Mapping)>>
+    = RefCell::new(Vec::with_capacity(MAPPINGS_CACHE_SIZE));
+
+pub fn clear_symbol_cache() {
+    MAPPINGS_CACHE.borrow_mut().clear()
 }
 
 fn with_mapping_for_path<F>(path: PathBuf, f: F)
 where
     F: FnMut(&Context<'_>),
 {
-    MAPPINGS_CACHE.with(|cache| {
-        let mut cache = cache.borrow_mut();
+    let mut cache = MAPPINGS_CACHE.borrow_mut();
 
-        let idx = cache.iter().position(|&(ref p, _)| p == &path);
+    let idx = cache.iter().position(|&(ref p, _)| p == &path);
 
-        // Invariant: after this conditional completes without early returning
-        // from an error, the cache entry for this path is at index 0.
+    // Invariant: after this conditional completes without early returning
+    // from an error, the cache entry for this path is at index 0.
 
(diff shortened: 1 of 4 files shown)


alexcrichton:
Thanks for the PR! Could this perhaps avoid picking up a dependency on `lazy_static` for now? It's probably best to tackle https://github.com/rust-lang/backtrace-rs/issues/225 in a dedicated PR.

Additionally if this didn't actually fix the issue for you I wonder if there's some more investigation to do? There may be more caches that need to be cleared in this crate somewhere I'm not realizing?

(one may be libunwind which may have its own set of internal caches)

Tomcc:
Sure, I can avoid `lazy_static`. It'll be a bit ugly but it's doable.

> Additionally if this didn't actually fix the issue for you I wonder if there's some more investigation to do? There may be more caches that need to be cleared in this crate somewhere I'm not realizing?

Actually this PR fixed my problem! The extra memory leftover that I mentioned on the other issue was... the `println!` that I used to debug it.
Interestingly, printing out anything from a background thread seems to allocate a bunch of caches as well... I've already deployed this code though, and it does fix the problem. I caused a few backtraces and it's still sitting at 38mb :)

src/symbolize/gimli.rs
```diff
@@ -329,28 +329,35 @@ impl Mapping {
...
+// between many shared libraries.
+//
+// The `addr2line::Context` structures are pretty expensive to create. Its
+// cost is expected to be amortized by subsequent `locate` queries, which
+// leverage the structures built when constructing `addr2line::Context`s to
+// get nice speedups. If we didn't have this cache, that amortization would
+// never happen, and symbolicating backtraces would be ssssllllooooowwww.
+static mut MAPPINGS_CACHE: Option<Mutex<Vec<(PathBuf, Mapping)>>> = None;
+
+fn lazy_cache() -> &'static Mutex<Vec<(PathBuf, Mapping)>> {
```
alexcrichton: If you're gonna go ahead and drop the `thread_local!` then the `Mutex` here can be dropped as well. All of the function in this module are externally synchronized so it's ok to to use the `static mut` here. (FWIW this implementation isn't threadsafe since `get_or_insert_with` isn't threadsafe, but it's ok in this specific context since thread safety isn't necessary).

Could this also be switched to be the same interface as before, something like:

```rust
fn with_cache(f: impl FnOnce(&mut Vec<(PathBuf, Mapping)>)) {
    // ...
}
```

so that way the `'static` lifetime won't leak

Tomcc: You're right, I didn't realize that once the `unsafe` was there then the `Mutex` wasn't needed anymore. And the `with_cache` function can wrap the `static` entirely, so it's not visible outside, which is nice.

src/symbolize/gimli.rs
```diff
@@ -329,7 +328,9 @@ impl Mapping {
     }
 }
 
-thread_local! {
+type Cache = Vec<(PathBuf, Mapping)>;
+
+fn with_cache(f: impl FnOnce(&mut Cache)) {
```
alexcrichton: I think we'll want this to be `unsafe` with a doc comment saying it needs to be externally synchronized

src/symbolize/gimli.rs
```diff
@@ -340,17 +341,20 @@ thread_local! {
...
     // get nice speedups. If we didn't have this cache, that amortization would
     // never happen, and symbolicating backtraces would be ssssllllooooowwww.
-    static MAPPINGS_CACHE: RefCell<Vec<(PathBuf, Mapping)>>
-        = RefCell::new(Vec::with_capacity(MAPPINGS_CACHE_SIZE));
+    static mut MAPPINGS_CACHE: Option<Cache> = None;
+
+    unsafe { f(MAPPINGS_CACHE.get_or_insert_with(|| Vec::with_capacity(MAPPINGS_CACHE_SIZE))) }
+}
+
+pub fn clear_symbol_cache() {
```
alexcrichton: I think this is currently not threadsafe since this doesn't acquire any locks, but this function should be `unsafe` (surfaced from `with_cache`) and then the wrapper in `symbolize/mod.rs` should grab the lock. Additionally the function in `symbolize/mod.rs` I think should only be enabled for now with the `std` feature enabled.

Tomcc: Hmm yeah, I didn't realize that now that was all non-threadsafe. I've never used `unsafe` before and I'm too used to Rust telling me when something I'm doing is dumb :)
I guess it means it works!

Anyway, I'll make the changes you suggested

alexcrichton:
Thanks again for this!

Tomcc:
Thank you for helping me with this!
Btw, I saw that all my commits with horrible messages went in as-is... I thought I'd have a chance to clean them up before merging them :P 
I guess it's too late now!

alexcrichton:
Ah no worries I don't tend to be too strict with commit history, it's all good!

Issue #229, opened 2019-07-31
potential performance regression on windows


AxelNennker:
Hyperledger Indy moved back from backtrace 0.3.33 to 0.3.11 because it seems the Windows performance worsened. Please see
https://github.com/hyperledger/indy-sdk/pull/1763

Would be nice if you could look into this. Thanks

alexcrichton:
Thanks for the report! Is there any more information to go along with this? This is unfortunately not really actionable just knowing "windows might be slow"

AxelNennker:
I don't have a Windows system to test this.
Would the benches here reveal such a regression?

Currently Indy is before a release so the core team is  really busy.
I'll write a PR to libindy that uses the 0.3.33 version and see how the windows CI reacts.
Although the Jenkins jobs are currently often terminated because of the pre-release load.

Sorry, I have not more data now. // CC @Artemkaaas @KitHat

alexcrichton:
Ok, some small performance testing reveals two culprits for the change in performance, although there could be more:

* First is https://github.com/rust-lang/backtrace-rs/pull/231, which will get fixed if that's merged. The problem is that `SymInitializeW`, an expensive operation, is called on each call to `Backtrace::new`.
* Second is that support for generating a backtrace inlined function frames on MSVC has been added. This manifests in the usage of `StackWalkEx` instead of `StackWalk64` where supported, and this also changes the symbolication routines. 

The former is relatively easily fixed, the second isn't really fixable AFAIK. A configuration option could be added to the crate for whether you want accurate or fast backtraces, but that's about the only solution there.

AxelNennker:
The Indy project is building a C-callable library libindy from Rust source code. So, I guess, we don't care about inlined function frames too much. Thanks for looking into this and publishing 0.3.34 so quickly. We'll give that a try soon.

alexcrichton:
Ok thanks for testing it out, I think if it's an acceptable level of performance we can probably leave it as-is, otherwise I can look into providing a build-time option to forcing usage of older functions on Windows.

Artemkaaas:
I did testing of backtrace=0.3.34 on our CI (windows machine).
The usual time of running tests (with usage backtrace=0.3.11) is about ~ 23-28 minutes.
With backtrace=0.3.34 it took about 50 minutes. I tried twice.
Looks like the problem still presents for windows.

alexcrichton:
[...]

PR #231, opened 2019-07-31
Ignore return of `SymInitializeW` on Windows

alexcrichton:
This commit updates the behavior of backtraces on Windows to execute
`SymInitializeW` globally once-per-process and ignore the return value
as to whether it succeeded or not. This undoes previous work in this
crate to specifically check the return value, and this is a behavior
update for the standard library when this goes into the standard
library.

The `SymInitializeW` function is required to be called on MSVC before
any backtraces can be captured or before any addresses can be
symbolized. This function is quite slow. It can only be called
once-per-process and there's a corresponding `SymCleanup` to undo the
work of `SymInitializeW`.

The behavior of what to do with `SymInitializeW` has changed a lot over
time in this crate. The very first implementation for Windows paired
with `SymCleanup`. Then reports of slowness at rust-lang/rustup.rs#591
led to ac8c6d23db where `SymCleanup` was removed. This led to #165 where
because the standard library still checked `SymInitializeW`'s return
value and called `SymCleanup` generating a backtrace with this crate
would break `RUST_BACKTRACE=1`. Finally (and expectedly) the performance
report has come back as #229 for this crate.

I'm proposing that the final nail is put in this coffin at this point
where this crate will ignore the return value of `SymInitializeW`,
initializing symbols once per process globally. This update will go into
the standard library and get updated on the stable channel eventually,
meaning both libstd and this crate will be able to work with one another
and only initialize the process's symbols once globally. This does mean
that there will be a period of "breakage" where we will continue to make
`RUST_BACKTRACE=1` not useful if this crate is used on MSVC, but that's
sort of the extension of the status quo as of a month or so ago. This
will eventually be fixed once the stable channel of Rust is updated.

### DIFF
--- a/src/capture.rs
+++ b/src/capture.rs
@@ -124,7 +124,6 @@ impl Backtrace {
     /// enabled, and the `std` feature is enabled by default.
     #[inline(never)] // want to make sure there's a frame here to remove
     pub fn new() -> Backtrace {
-        let _guard = lock_and_platform_init();
         let mut bt = Self::create(Self::new as usize);
         bt.resolve();
         bt
@@ -155,7 +154,6 @@ impl Backtrace {
     /// enabled, and the `std` feature is enabled by default.
     #[inline(never)] // want to make sure there's a frame here to remove
     pub fn new_unresolved() -> Backtrace {
-        let _guard = lock_and_platform_init();
         Self::create(Self::new_unresolved as usize)
     }
 
@@ -205,7 +203,6 @@ impl Backtrace {
     /// This function requires the `std` feature of the `backtrace` crate to be
     /// enabled, and the `std` feature is enabled by default.
     pub fn resolve(&mut self) {
-        let _guard = lock_and_platform_init();
         for frame in self.frames.iter_mut().filter(|f| f.symbols.is_none()) {
             let mut symbols = Vec::new();
             {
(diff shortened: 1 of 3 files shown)


alexcrichton:
@sfackler mind taking a look at this and lending your opinion to this? This will be a behavior change for the standard library on MSVC, so I figured it's good to double check with at least one other libs member to ensure it's ok.

sfackler:
Under what conditions does SymInitializeW fail? Being called when initialized?

alexcrichton:
Unfortunately that's not 100% clear to me. I know for certain of one failure case, but I don't know of others. 

The for-certain failure case is that if you call `SymInitializeW` twice in a row it will fail the second time with "error invalid parameter". This is because it can only be called if symbols aren't initialized (or if they were previously torn down by `SymCleanup`).

Presumably we could also detect cases of like "actual symbol initialization failed" and ignore the "you initialized twice" errors and actually report "symbol initialization failed" failures, but I don't currently know of a way to do so.

I believe there are also not really any consequences for using symbolication/backtrace functions if `SymInitialize` fails for a legitimate reason. I think you just get weird and/or truncated results.

sfackler:
Ok, and this is going to regress the standard library since it expects SymInitializeW to succeed before taking each backtrace? It might make sense to backport a backtrace upgrade to beta in that case.

alexcrichton:
It sort of depends on how you look at it. The standard library has always behaved as it does today, which is to pair initialize/cleanup calls and skip generating a backtrace if initialization fails. This crate itself was changed a month or so ago to *also* pair initialize/cleanup calls to enusre that if you use this crate `RUST_BACKTRACE` still works. 

In that sense the standard library if you don't use this crate doesn't change, it just continues to work. The libstd from January will be as "broken" as the libstd when this PR hits crates.io for the `backtrace` crate. Pairing `backtrace` from crates.io now with libstd from June (ish) works today, but after this PR hits crates.io it'll go back to the January state of not working.

I would classify this more as a temporary regression for this crate than for the standard library myself.

sfackler:
Maybe someday we'll just be able to use Gimli :P

PR #233, opened 2019-08-01
Fix MSVC UWP builds

jdm:
Unmodified backtrace-rs breaks the rustc build when building a standard library for an experimental MSVC UWP target. These changes allow the build to proceed.

### DIFF
--- a/crates/backtrace-sys/src/lib.rs
+++ b/crates/backtrace-sys/src/lib.rs
@@ -1,6 +1,7 @@
 #![allow(bad_style)]
 #![no_std]
 
+#[allow(unused_extern_crates)]
 extern crate libc;
 
 #[cfg(not(empty))]
--- a/src/symbolize/mod.rs
+++ b/src/symbolize/mod.rs
@@ -458,7 +458,7 @@ pub fn clear_symbol_cache() {
 mod dladdr;
 
 cfg_if::cfg_if! {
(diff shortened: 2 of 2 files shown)


alexcrichton:
Hm can you gist what the build errors were? There's one uwp-specific failure but the warnings shouldn't need to be handled to get the build to work

jdm:
The build errors were unused code warnings that were apparently treated as errors as part of the rustc build :/

alexcrichton:
Those sorts of warnings are suppressed by default for crates.io crates, so it may have showed up if you were using `[patch]` perhaps?

jdm:
Ah, indeed. That is exactly what happened.

alexcrichton:
Is this still needed to gate code being compiled on UWP? Apart from warnings-as-errors was there another build error?

jdm:
Yes, using a rustc commit that adds a x86_64-uwp-windows-msvc target and builds the standard library for it, backtrace-rs does not build without https://github.com/rust-lang/backtrace-rs/pull/233/commits/5d18704547d39201bde959c986728cf94d0bdabe applied.

alexcrichton:
Could that be split into a separate PR? I'd prefer to handle the unused warnings with fewer attributes sprinkled about

jdm:
Not necessary any more.

Issue #232, opened 2019-07-31
Improve performance of gimli dwarf parsing


alexcrichton:
Today I was playing around with the various smoke benchmarks we have in this repository, and I executed `cargo +nightly bench` followed by `cargo +nightly bench --features gimli-symbolize`. The results were pretty interesting! Running them through `cargo benchcmp` I get:

```
 name                                 current ns/iter  gimli ns/iter  diff ns/iter    diff %  speedup 
 new                                  13,833           134,499             120,666   872.31%   x 0.10 
 new_unresolved                       5,432            5,490                    58     1.07%   x 0.99 
 new_unresolved_and_resolve_separate  13,896           134,745             120,849   869.67%   x 0.10 
 trace                                2,715            2,801                    86     3.17%   x 0.97 
 trace_and_resolve_callback           4,174            109,484             105,310  2523.00%   x 0.04 
 trace_and_resolve_separate           4,632            116,182             111,550  2408.25%   x 0.04 
```

Overall it looks like the gimli implementation in this crate is 10-20x slower than libbacktrace at symbolicating symbols in a backtrace.

Some of the top entries in a `perf record` are:

```
+   18.70%  benchmarks-2de6  benchmarks-2de6689c5c5a5583  [.] gimli::read::line::LineRows<R,Program,Offset>::next_row                                                                                                                        ▒
+    7.89%  benchmarks-2de6  benchmarks-2de6689c5c5a5583  [.] backtrace::symbolize::gimli::resolve                                                                                                                                           ◆
+    5.75%  benchmarks-2de6  libc-2.27.so                 [.] _dl_addr                                                                                                                                                                       ▒
+    4.39%  benchmarks-2de6  benchmarks-2de6689c5c5a5583  [.] addr2line::Context<R>::find_frames                                                                                                                                             ▒
+    3.90%  benchmarks-2de6  benchmarks-2de6689c5c5a5583  [.] gimli::leb128::read::unsigned                                                                                                                                                  ▒
+    3.66%  benchmarks-2de6  benchmarks-2de6689c5c5a5583  [.] gimli::leb128::read::signed                                                                                                                                                    ▒
+    3.08%  benchmarks-2de6  benchmarks-2de6689c5c5a5583  [.] gimli::read::unit::parse_attribute                                                                                                                                             ▒
+    2.84%  benchmarks-2de6  libc-2.27.so                 [.] _int_malloc                                                                                                                                                                    ▒
[...]

PR #235, opened 2019-08-07
gimli: Only iterate shared libraries once

alexcrichton:
This commit updates the gimli implementation to iterate over all shared
libraries only once, caching information about them to get processed
later. This information is usually static for the lifetime of the
program, and this improves the performance of resolving a backtrace
significantly in some situations (by up to 30%).

The organization of the gimli backend has been updated to reflect this
cache where `Cache` is now a dedicated type with methods hanging off it.
Some helper code had to be duplicated from the `findshlibs` crate to do
this entirely.

cc #232

### DIFF
--- a/src/symbolize/gimli.rs
+++ b/src/symbolize/gimli.rs
@@ -18,8 +18,9 @@ use findshlibs::{self, Segment, SharedLibrary};
 use libc::c_void;
 use memmap::Mmap;
 use std::env;
+use std::ffi::OsString;
 use std::fs::File;
-use std::path::{Path, PathBuf};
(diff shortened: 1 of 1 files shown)


philipc:
Do we need to handle `dlopen()` being called after we have already created the cache?

alexcrichton:
Heh I wondered the same! I don't think that case is handled in the libbacktrace library today (from what I can tell) so I think it's fine to leave that to a future implementation. That being said though I should probably open an issue on this crate. I don't think it'd be too hard to fix it if it comes up and preserve these performance semantics.

PR #237, opened 2019-08-07
gimli: Look at object dynamic symbol tables 

alexcrichton:
This commit updates the gimli parsing code to consult the dynamic symbol
table for each object in addition to the normal symbol table. This also
starts filtering symbols in the same way libbacktrace does to ensure
symbolication doesn't go down the rabbit hole with non-function
symbols.

This overall removes the need for gimli to fall back to dladdr
functionally for now. In testing this speeds up benchmarks by 15% by
avoiding calling `dladdr` in libc.

### DIFF
--- a/src/symbolize/gimli.rs
+++ b/src/symbolize/gimli.rs
@@ -208,7 +208,10 @@ cfg_if::cfg_if! {
         struct Object<'a> {
             elf: Elf<'a>,
             data: &'a [u8],
-            syms: Vec<goblin::elf::Sym>,
+            // List of pre-parsed and sorted symbols by base address. The
+            // boolean indicates whether it comes from the dynamic symbol table
+            // or the normal symbol table, affecting where it's symbolicated.
+            syms: Vec<(goblin::elf::Sym, bool)>,
         }
 
         impl<'a> Object<'a> {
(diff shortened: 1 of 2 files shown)


src/symbolize/gimli.rs
```diff
@@ -217,8 +220,24 @@ cfg_if::cfg_if! {
...
+                    .map(|s| (s, false))
+                    .chain(elf.dynsyms.iter().map(|s| (s, true)))
+                    // Only look at function/object symbols. Not entirely sure
+                    // why, but this is what libbacktrace does.
+                    .filter(|(s, _)| {
+                        s.is_function() || s.st_type() == goblin::elf::sym::STT_OBJECT
+                    })
+                    // skip anything that's in an undefined section header,
+                    // again not entirely sure why but this is what libbacktrace
+                    // does.
```
philipc: If the symbol is undefined then it doesn't have an address in this library. This happens for imported symbols.

src/symbolize/gimli.rs
```diff
@@ -217,8 +220,24 @@ cfg_if::cfg_if! {
...
                 }
-                let mut syms = elf.syms.iter().collect::<Vec<_>>();
-                syms.sort_unstable_by_key(|s| s.st_value);
+                let mut syms = elf
+                    .syms
+                    .iter()
+                    .map(|s| (s, false))
+                    .chain(elf.dynsyms.iter().map(|s| (s, true)))
+                    // Only look at function/object symbols. Not entirely sure
+                    // why, but this is what libbacktrace does.
```
philipc: I guess the assumption is that only functions contain executable code.

src/symbolize/gimli.rs
alexcrichton: Oh nice, thanks! I'll update the comment here

src/symbolize/gimli.rs
alexcrichton: The reason I ended up putting this in is that without it a few addresses were still falling back to `dladdr`. That meant that there's some symbol here which was around the same address as a function address and ended up getting mistaken as a function address.

Do you know what might be causing that? Are the relative `st_value` fields for symbols relative to their own sections, so we only care about symbols relative to the text section?

philipc: I don't know what might cause that. In executables and shared objects, `st_value` is a virtual address. I think the exception to that is TLS symbols?

Issue #230, opened 2019-07-31
Synchronize dbghelp with the standard library


alexcrichton:
The dbghelp library on MSVC is not safe to use concurrently from multiple threads. There is no synchronization between the standard library's usage of `RUST_BACKTRACE=1` and this crate generating a backtrace, meaning that it's not technically possible to 100% safely use this crate on MSVC. 

Fixing this will likely require some form of global lock exported by the standard library specifically for backtraces and specifically for MSVC (it can be a noop on all other platforms). This library would then, when compiled on crates.io, need to coordinate with the standard library in terms of locking to ensure that only one implementation is using dbghelp at a time.

arsing:
This also causes heap corruption crashes currently if both libstd and a user crate take backtraces at approximately the same time. This happens with one of our crates' tests very reliably, where one test happens to run an `assert!()` (which triggers libstd backtrace collection) at about the same time as another test creates a `failure` error (which collect its own backtrace).

I happened to spot `StackWalkEx` in the crashing thread's backtrace, but I assume it would happen even with the older `backtrace` crate that solely used `StackWalk64` ? So just downgrading the `backtrace` crate would not be useful unless we also used an older Rust that didn't use the backtrace crate?

alexcrichton:
Yes this unfortunately has always been a bug with the `backtrace` crate, no version has ever had this bug fixed. This is definitely a case where binaries can segfault because the documentation of Windows explicitly states that it needs synchronization, and we're not doing that! It "just so happens to work" most of the time because it's relatively rare to get two backtraces at the same time.

alexcrichton:
I've posted a possible mitigation for what I'd hope is 90%-ish of cases at https://github.com/rust-lang/backtrace-rs/pull/241

PR #241, opened 2019-08-15
Fix some cases of concurrent panics and backtraces

alexcrichton:
This commit adds an awful hack to this crate to attempt to mitigate the
issue of concurrently generating a backtrace in the standard library via
`RUST_BACKTRACE=1` and also generating a backtrace with this crate. On
Windows the system library that we used for doing this is required to be
single-threaded but we don't actually provide such a guarantee when
using this crate because we can't easily synchronize with the standard
library.

The hack applied in this crate is to use a *panic hook* to inject
ourselves into the panic process and ensure that panics are synchronized
with generating a backtrace. The commit itself has a lot more comments
about how this is not a complete nor a bullet-proof solution. It's just
sort of the best approximation I think we can do with stable Rust.
Long-term we're going to want to have some sort of official stable API
in libstd to do this coordination.

cc #230

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -152,3 +152,8 @@ edition = '2018'
 name = "accuracy"
 required-features = ["std", "dbghelp", "libbacktrace", "libunwind"]
 edition = '2018'
+
+[[test]]
+name = "concurrent-panics"
+required-features = ['std']
+harness = false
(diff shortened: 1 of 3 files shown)


alexcrichton:
@sfackler mind double-checking me on this logic and this patch? This doesn't actually affect the standard library at all, just the crates.io version of this crate. I'm curious to get your take on this strategy and the general problem as well

sfackler:
How about using a session-scoped named lock to synchronize between std's backtrace and the normal backtrace? https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-createmutexa

alexcrichton:
Wow that's a way better idea, nice thinking! I'll implement that soon

PR #242, opened 2019-08-16
Fix concurrent panics and backtraces

### LINKED ISSUE #230: Synchronize dbghelp with the standard library

alexcrichton:
On Windows `dbghelp.dll` is required to be used from only one thread at
a time, and while this crate provides that guarantee this crate does not
synchronize with itself in the standard library. This commit solves this
issue by using a Windows-specific trick by creating a named mutex for
synchronization.

When this crate is updated in the standard library it means that the
named mutex here will be shared with the standard library, so the
standard library and this crate should be sharing the same
synchronization primitive and should be able to coordinate calls to
`dbghelp.dll`.

More details are included in the commit itself, but this should...

Closes #230

### DIFF
--- a/src/dbghelp.rs
+++ b/src/dbghelp.rs
@@ -244,45 +244,127 @@ dbghelp! {
     }
 }
 
-pub struct Init;
+pub struct Init {
+    lock: HANDLE,
+}
 
-/// Unsafe because this requires external synchronization, must be done
-/// inside of the same lock as all other backtrace operations.
+/// Initialize all support necessary to access `dbghelp` API functions from this
+/// crate.
 ///
-/// Note that the `Dbghelp` returned must also be dropped within the same
-/// lock.
+/// Note that this function is **safe**, it internally has its own
+/// synchronization. Also note that it is safe to call this function multiple
+/// times recursively.
 #[cfg(all(windows, feature = "dbghelp"))]
-pub unsafe fn init() -> Result<Init, ()> {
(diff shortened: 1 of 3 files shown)


alexcrichton:
Note that a test was included in https://github.com/rust-lang/backtrace-rs/pull/241, but such a test won't work until we've updated this crate in the standard library itself, so I haven't included the test just yet.

r? @sfackler

src/dbghelp.rs
```diff
@@ -244,45 +244,127 @@ dbghelp! {
...
 }
 
-pub struct Init;
+pub struct Init {
+    lock: HANDLE,
+}
 
-/// Unsafe because this requires external synchronization, must be done
-/// inside of the same lock as all other backtrace operations.
+/// Initialize all supoprt necessary to access `dbghelp` API functions from this
```
sfackler:
```suggestion
/// Initialize all support necessary to access `dbghelp` API functions from this
```

PR #245, opened 2019-08-27
Custom backtrace output for Fuchsia

cramertj:
Fuchsia can't do traditional target-side symbolization. Instead, it uses a custom format containing information about the current location of libraries in memory and the addresses of the stack frames.

r? @alexcrichton @jakeehrlich


alexcrichton:
Looks great to me! Some minor things:

* Looks like a `rustfmt` is needed
* I don't think that the `Iterator` style API will be immediately usable by libstd because it's using inline iteration with backtrace frames. Perhaps some sort of `struct` with a method to print the context and then a method to call when the ip for each frame is learned?

Other than that seems all good to me!

alexcrichton:
It may be useful as well to develop the libstd patch to use this in tandem, which should be possible by using `[patch]` and compiling libstd for fuchsia

alexcrichton:
Ok I was struck with some inspiration while reading over this, and I hope you're ok that I went ahead and tacked on some more commits here. I'm hoping that these commits will also make it easier to integrate into libstd because we can basically just delete all of `sys_common/backtrace.rs` now and use the printing code here.

Mind taking a look and seeing how it works out? Do you think that the last few commits are too overzealous?

cramertj:
Looks reasonable enough for me! I'll pull this down and test locally to make sure things still work. Did you think about how you want to handle cutting out things after `__rust_begin_short_backtrace` like is currently done in libstd?

cramertj:
> I'll pull this down and test locally to make sure things still work.

Still good on my end!

alexcrichton:
Yeah that was the one thing I left out since I don't think I've ever seen a backtrace get successfully truncated with that symbol, but it's harmelss enough to put back in and it probably should happen. Some filtering logic will still need to be in libstd like the max number of frames, but I think that's fine

cramertj:
SGTM!

PR #254, opened 2019-09-23
Fix type error in SGX backtrace printing

nhynes:
Also adds the target to CI to prevent this from happening again in the future.


alexcrichton:
I tweaked CI a bit to just run the build for sgx rather than the full test suite (which was failing), otherwise thanks!

PR #263, opened 2019-10-29
dynamic get `QueryFullProcessImageNameA` for XP support

lynnux:
this is for gnu build on xp.

btw, the msvc build not use libbacktrace, it use dbghelp, but the dbhelp.dll on XP lack those functions:
```
SymInitializeW
SymFromAddrW
SymGetLineFromAddrW64
SymFromInlineContextW
SymGetLineFromInlineContextW
```
which seems a lot work to do, the easy solution for msvc build is to combined a newer version of dbghelp.dll with exe on xp.


lynnux:
Ok, `GetModuleFileName ` removed

alexcrichton:
Thanks!

PR #265, opened 2019-11-27
Add `format_trace(impl fmt::Write, PrintFmt)` 

SimonSapin:
This is similar to `write!(stream, "{:?}", Backtrace::new())`, but avoids memory allocation. This can be useful in a signal handler, where allocation may cause deadlocks.

### DIFF
--- a/examples/backtrace.rs
+++ b/examples/backtrace.rs
@@ -1,7 +1,17 @@
 extern crate backtrace;
 
-use backtrace::Backtrace;
+use backtrace::{format_trace, Backtrace, PrintFmt};
 
 fn main() {
-    println!("{:?}", Backtrace::new());
+    print!("{:?}", Backtrace::new());
+    format_trace(Utf8(std::io::stdout()), PrintFmt::Short).unwrap();
+    format_trace(Utf8(std::io::stdout()), PrintFmt::Full).unwrap();
+}
+
+struct Utf8<W>(W);
+
+impl<W: std::io::Write> std::fmt::Write for Utf8<W> {
+    fn write_str(&mut self, s: &str) -> Result<(), std::fmt::Error> {
+        self.0.write_all(s.as_bytes()).map_err(|_| std::fmt::Error)
+    }
 }
--- a/src/format.rs
+++ b/src/format.rs
@@ -1,0 +1,115 @@
+use crate::PrintFmt;
+use core::ffi::c_void;
+use core::fmt::{self, Write};
+
+/// Inspects the current call-stack, formatting it to `stream`.
+///
+/// This is similar to `write!(stream, "{:?}", Backtrace::new())`,
+/// but avoids memory allocation.
+/// This can be useful in a signal handler, where allocation may cause deadlocks.
+///
+/// # Required features
+///
+/// This function requires the `std` feature of the `backtrace` crate to be
+/// enabled, and the `std` feature is enabled by default.
+///
+/// # Panics
+///
+/// This function strives to never panic, but if the `cb` provided panics then
+/// some platforms will force a double panic to abort the process. Some
+/// platforms use a C library which internally uses callbacks which cannot be
+/// unwound through, so panicking from `cb` may trigger a process abort.
+#[cfg(feature = "std")]
+#[inline(never)]
+pub fn format_trace<W: Write>(mut stream: W, mode: PrintFmt) -> fmt::Result {
+    let _guard = crate::lock::lock();
+    write!(
+        stream,
+        "{:?}",
+        FormatTrace {
+            mode,
+            entry_point_address: format_trace::<W> as *mut c_void,
+        }
+    )
+}
+
+/// Same as `format_trace`, only unsafe as it's unsynchronized.
+///
+/// This function does not have synchronization guarentees but is available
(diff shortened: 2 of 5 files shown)


SimonSapin:
Servo has a signal handler that prints a backtrace on segfaults. Currently it uses `println!("{:?}", Backtrace::new)` but that seems to cause deadlocks, possibly in the memory allocator: https://github.com/servo/servo/issues/24881. Using lower-level API instead to avoid allocating seems to fix the deadlocks, but takes a fair amount of code that is not at all Servo-specific and could be useful to other projects.

SimonSapin:
`cargo fmt --all -- --check` fails, but it also does on `master`.

src/format.rs
```diff
@@ -0,0 +1,115 @@
+use crate::PrintFmt;
+use core::ffi::c_void;
+use core::fmt::{self, Write};
+
+/// Inspects the current call-stack, formatting it to `stream`.
+///
+/// This is similar to `write!(stream, "{:?}", Backtrace::new())`,
+/// but avoids memory allocation.
+/// This can be useful in a signal handler, where allocation may cause deadlocks.
```
alexcrichton: This unfortunately technically isn't quite true because the underlying implementation almost always allocates memory in one way or another. While this avoids all Rust-based allocations, the backtrace unwinder/etc, if called for the first time will allocate memory here.

In general nothing in this crate is suitable for usage in a signal handler right now unfortunately.

src/format.rs
```diff
@@ -0,0 +1,115 @@
...
+///
+/// This function requires the `std` feature of the `backtrace` crate to be
+/// enabled, and the `std` feature is enabled by default.
+///
+/// # Panics
+///
+/// This function strives to never panic, but if the `cb` provided panics then
+/// some platforms will force a double panic to abort the process. Some
+/// platforms use a C library which internally uses callbacks which cannot be
+/// unwound through, so panicking from `cb` may trigger a process abort.
```
alexcrichton: Could the docs here include an example of how to call this?

src/format.rs
```diff
@@ -0,0 +1,115 @@
...
+#[cfg(feature = "std")]
+#[inline(never)]
```
alexcrichton: Mind throwing a small comment on this explaining why it's `inline(never)`?

examples/backtrace.rs
```diff
@@ -1,7 +1,17 @@
...
+    print!("{:?}", Backtrace::new());
+    format_trace(Utf8(std::io::stdout()), PrintFmt::Short).unwrap();
+    format_trace(Utf8(std::io::stdout()), PrintFmt::Full).unwrap();
+}
+
+struct Utf8<W>(W);
+
+impl<W: std::io::Write> std::fmt::Write for Utf8<W> {
+    fn write_str(&mut self, s: &str) -> Result<(), std::fmt::Error> {
+        self.0.write_all(s.as_bytes()).map_err(|_| std::fmt::Error)
```
alexcrichton: Mind backing this out by the time this lands?

Useful for context though!

SimonSapin: Do you mean removing changes to this file entirely? Or should they be a separate example program?

src/format.rs
SimonSapin: I see.

For what it’s worth using something like this PR instead of `Backtrace::new()` in Servo’s signal handler does seem to fix a deadlock, presumably through reducing the number of allocations. (Or at least I couldn’t reproduce it after running the test case a hundred times, whereas with `Backtrace::new()` I would reliably reproduce within ten runs.) But it’s quite possible that there’s still another deadlock waiting to happen and maybe we should remove that handler entirely.

I’ll remove mention of signal handler in these doc-comments. What else do you think could be a good use case for this new feature? In theory this is doing less work and might be more efficient than creating a `Backtrace` only to print it and immediately drop it, but I don’t know if the difference has a practical impact.

alexcrichton: Yes it's just a deadlock waiting to happen. The underlying implementation of backtraces are not at all guaranteed to not malloc, but they try not to to be efficient. `Backtrace::new()` is guaranteed to allocate, where this API as-is just allocates some of the time, not all the time. 

Getting a backtrace from a signal handler is a really tricky thing to do and generally not covered by this crate.

TBH I'm not sure what a use case for this feature is, folks are generally ok with the mild cost that `Backtrace::new` has

examples/backtrace.rs
alexcrichton: Oh just in the sense that this example file was intended to just print one backtrace, but this might make a good `# Examples` section for the new API?

src/format.rs
SimonSapin: Right, my motivation for this PR was not the CPU time cost but trading a deadlock I could reliably reproduce for a possible one I haven’t seen happen. (Knowing this still isn’t a fully correct solution.)

If the only use case for this is incorrect and something the project doesn’t want to cover, maybe it shouldn’t be included? Even though based on what this code does it seems like within this crate would be the right location.

I’m OK with closing this PR in that case.

alexcrichton:
Ok I'm gonna close this due to the discussion at https://github.com/rust-lang/backtrace-rs/pull/265#discussion_r352654555, but perhaps clarifying in the README that this is not appropriate for usage in signal handlers would be good to indicate!

PR #266, opened 2019-12-03
Bound target_env = sgx with target_vendor = fortanix

dingelish:
As discussed in https://github.com/rust-lang/rust/pull/57243, all of codes used for Fortanix's SGX target should be `cfg`d together with target vendor check `target_vendor = "fortanix"`. If not, these codes would break all other targets using `target_env = "sgx"`


alexcrichton:
Thanks!

PR #268, opened 2019-12-12
Printing backtrace ip for x86_64-fortanix-unknown-sgx

jethrogb:
Not sure what happened, this used to work.

cc @VardhanThigle

### DIFF
--- a/src/print.rs
+++ b/src/print.rs
@@ -185,6 +185,8 @@ impl BacktraceFrameFmt<'_, '_, '_> {
         filename: Option<BytesOrWideString>,
         lineno: Option<u32>,
     ) -> fmt::Result {
+        let mut print_frame_ip = PrintFmt::Full == self.fmt.format;
+
         // No need to print "null" frames, it basically just means that the
         // system backtrace was a bit eager to trace back super far.
         if let PrintFmt::Short = self.fmt.format {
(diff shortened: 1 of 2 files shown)


src/print.rs
```diff
@@ -196,30 +198,32 @@ impl BacktraceFrameFmt<'_, '_, '_> {
         // To reduce TCB size in Sgx enclave, we do not want to implement symbol
         // resolution functionality.  Rather, we can print the offset of the
         // address here, which could be later mapped to correct function.
-        #[cfg(all(feature = "std", target_env = "sgx", target_vendor = "fortanix"))]
```
alexcrichton: I think the removal of this `feature` directive will break this building when included into libstd?

src/print.rs
```diff
@@ -196,30 +198,32 @@ impl BacktraceFrameFmt<'_, '_, '_> {
...
-            if let PrintFmt::Full = self.fmt.format {
+            if print_frame_ip {
                 write!(self.fmt.fmt, "{:1$}", "", HEX_WIDTH + 3)?;
             }
         }
 
         // Next up write out the symbol name, using the alternate formatting for
         // more information if we're a full backtrace. Here we also handle
         // symbols which don't have a name,
+        #[cfg(not(all(target_env = "sgx", target_vendor = "fortanix")))]
```
alexcrichton: Could this use an `if` instead of using an attribute?

jethrogb:
I need to take a look at https://github.com/nhynes/backtrace-rs/tree/fix-sgx

workingjubilee:
@jethrogb Hello! Do you intend to pick up this PR again? Is this even still needed?

PR #271, opened 2020-01-02
libbacktrace isn't used on HermitCore

stlankes:
Currently, HermitCore doesn't support libbacktrace. => disable it


stlankes:
@alexcrichton Do you have an idea, why 3 checks failed?

alexcrichton:
Unsure! I'll take a look though.

Issue #288, opened 2020-02-13
Get register values for frame


bjorn3:
I got a PR for https://github.com/bjorn3/pretty_backtrace which shows the values of locals for every frame.This currently requires `unwind-rs` to get the register values. Unfortunately it only supports ELF based unixes. Adding support to `backtrace-rs` would make it possible to do it on macOS too.

When using libunwind, this can use `_Unwind_GetGR`. On Windows, you can get register values from `CONTEXT`.

alexcrichton:
Seems reasonable to me to add!

bjorn3:
Looking at the libgcc source code I'm not sure this can safely be added:

https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=libgcc/unwind-dw2.c;h=d0afce7a9ea9f5b12a5a01ef1e940e1452b48cab;hb=HEAD#l247

> ```c
>   /* This will segfault if the register hasn't been saved. */
> ```

workingjubilee:
Hmm.

Alright, then we can probably close this for now.

PR #291, opened 2020-02-15
Get register values

bjorn3:
cc #288

This currently only works on x86_64 in combination with libunwind.

@alexcrichton do you have any idea how to test this? I am currently just printing the register values in the test.

### DIFF
--- a/src/backtrace/dbghelp.rs
+++ b/src/backtrace/dbghelp.rs
@@ -46,6 +46,10 @@ impl Frame {
         self.ip()
     }
 
+    pub fn registers(&self) -> Option<super::Registers> {
+        None
+    }
+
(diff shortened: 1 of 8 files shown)


alexcrichton:
This is probably best testable with inline assembly or something like that perhaps?

bjorn3:
https://github.com/gcc-mirror/gcc/blob/a7201a085cc30f89944931d8fb1d7936f02a169f/libgcc/unwind-dw2.c#L249

```
/* This will segfault if the register hasn't been saved.  */
```

It seems that `_Unwind_GetGR` is only allowed to be called with saved registers. This means that at least the approach taken in this PR is wrong.

Issue #253, opened 2019-09-16
Depends on backtrace-sys even when rustc-dep-of-std is the only feature


RalfJung:
When disabling all features except for rustc-dep-of-std, this crate should have minimal dependencies and, in particular, it would be great if it could build without a C toolchain for the target. That is needed to keep libstd (with no features) buildable without a C toolchain (modulo https://github.com/rust-lang/rust/issues/56443).

However, currently, backtrace-sys is always built when building libstd, even with no features. The reason for this is probably that rustc-dep-of-std enables `backtrace-sys/rustc-dep-of-std`, which unfortunately implicitly enables `backtrace-sys`.

@alexcrichton [suggests](https://github.com/rust-lang/rust/pull/64444#issuecomment-531819584) to work around this by making backtrace-sys with no features do basically nothing, and then enabling the "actually-build" feature only from backtrace's "libbacktrace" feature.

PR #297, opened 2020-03-01
don't let rustc-dep-of-std enable backtrace-sys

RalfJung:
Fixes https://github.com/rust-lang/backtrace-rs/issues/253

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ exclude = ['crates/without_debuginfo']
 [dependencies]
 cfg-if = "0.1.10"
 rustc-demangle = "0.1.4"
-backtrace-sys = { path = "crates/backtrace-sys", version = "0.1.32", optional = true }
+backtrace-sys = { path = "crates/backtrace-sys", version = "0.1.32", optional = true, default_features = false }
 libc = { version = "0.2.45", default-features = false }
 core = { version = "1.0.0", optional = true, package = 'rustc-std-workspace-core' }
 compiler_builtins = { version = '0.1.2', optional = true }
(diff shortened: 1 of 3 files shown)


alexcrichton:
I sort of forget the context for this at this point again, but is this somethign that we can condition the target string on? For example could a new target be added to print out the `empty` cfg? Or is this required as part of a build elsewhere or something like that?

RalfJung:
The point of this is to be able to do a cross check build of libstd without the backtrace feature for foreign architectures with only a Rust toolchain. This is needed for https://github.com/rust-lang/miri/issues/1198: since Miri is just an interpreter, *in principle* we should be able to emulate any target platform on any host platform. Since recently Miri just does check builds for everything, so https://github.com/rust-lang/rust/issues/56443 is not a problem any more.

The only thing that still fails when I run `xargo-check check --target x86_64-apple-darwin` on my Linux box is that, even though the backtrace feature in libstd is disabled, libstd still depends on backtrace-sys (transitively) and its build script fails (as I have no apple toolchain and headers installed).

libstd without the backtrace feature depends on an "empty" backtrace with no features (you requested that in https://github.com/rust-lang/rust/pull/64444 to avoid lots of `cfg` in libstd). Unfortunately, due to cargo limitations, the fact that backtrace forwards the rustc-dep-of-std feature to backtrace-sys means that the rustc-dep-of-std backtrace feature implicitly enables backtrace-sys. So, that "empty" backtrace crate is actually not quite so empty when rustc-dep-of-std is enabled.

> is this somethign that we can condition the target string on

I don't follow... how would we condition the target string on anything? We don't control the target string?

alexcrichton:
Ah ok that makes sense, thanks for the explanation!

The motivating case looks to be "I disabled backtraces in libstd but backtrace-sys is still getting built when checking for standard platforms", specifically in your cross build case. That's good enough for me at least :)

Ah and for conditioning the target string, I thought this was originally motivated by "whenever I build for platform X I want to disable backtraces" which I think is best handled in the code itself, not with more features. That's not the motivation for this, though, so it's all good!

RalfJung:
> The motivating case looks to be "I disabled backtraces in libstd but backtrace-sys is still getting built when checking for standard platforms", specifically in your cross build case. 

Yes that's a good summary. :)

PR #300, opened 2020-03-03
gimli: Implement symbol table searches on macOS

alexcrichton:
If dwarf debug info isn't available we fall back to searching the symbol
table. Typically this is done with `dladdr` on most platforms but with
gimli we typically have all the infrastructure already in place to do
the search ourselves. This functionality was already filled out on
Linux and Windows, but it wasn't implemented on macOS yet because it
wasn't necessary.

Implementing a pretty simple version, however, shows substantial
speedups for the various benchmarks. Presumably `dladdr` isn't exactly
the fastest thing in the world and our sorted list search which is
cached must be much faster here!

The current comparison of before/after this change looks like:

```
 name                                 before ns/iter  after ns/iter  diff ns/iter   diff %  speedup
 new                                  81,472          9,047               -72,425  -88.90%   x 9.01
 new_unresolved                       2,126           2,009                  -117   -5.50%   x 1.06
 new_unresolved_and_resolve_separate  82,252          9,134               -73,118  -88.90%   x 9.01
 trace                                1,273           1,185                   -88   -6.91%   x 1.07
 trace_and_resolve_callback           67,403          2,123               -65,280  -96.85%  x 31.75
 trace_and_resolve_separate           76,452          2,822               -73,630  -96.31%  x 27.09
```

### DIFF
--- a/src/symbolize/gimli.rs
+++ b/src/symbolize/gimli.rs
@@ -158,6 +158,7 @@ cfg_if::cfg_if! {
         struct Object<'a> {
             macho: MachO<'a>,
             dwarf: Option<usize>,
+            syms: Vec<(&'a str, u64)>,
         }
 
         impl<'a> Object<'a> {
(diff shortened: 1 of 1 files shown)


RalfJung:
This PR seems likely responsible for a rustc test failure after bumping the libbacktrace version there: https://github.com/rust-lang/rust/pull/69729#issuecomment-596379840

PR #303, opened 2020-03-15
Support line-tables-only when using libbacktrace

tmiasko:
Previously when `backtrace_pcinfo` succeeded, but failed to obtain a
filename or a function name, the line number would be ignored. Instead,
when successful combine all available information.

For example, when using clang `-g1` or `-gline-tables-only` before:

```
   1: baz
   2: bar
   3: foo
```

and after:

```
   1: baz
             at src/callback.c:5
   2: bar
             at src/callback.c:9
   3: foo
             at src/callback.c:13
```

Note that the difference is probably not observable on gcc which includes the
relevant debug info on level 1, nor in rustc where the differences between
debug levels 1 and 2 are currently rather limited.


alexcrichton:
Thanks for this!

PR #306, opened 2020-04-06
Add reading of general purpose registers.

yurydelendik:
This is alternative proposal for #291 or #288. The difference is that the registers content only accessible during `trace()` function. This means that all responsibility of saving registers values falls on the caller.

### DIFF
--- a/src/backtrace/libunwind.rs
+++ b/src/backtrace/libunwind.rs
@@ -51,6 +51,16 @@ impl Frame {
         unsafe { uw::_Unwind_GetIP(ctx) as *mut c_void }
     }
 
+    pub fn gr(&self, index: u32) -> Option<usize> {
+        let ctx = match *self {
+            Frame::Raw(ctx) => ctx,
+            Frame::Cloned { .. } => {
+                return None;
+            }
+        };
+        Some(unsafe { uw::_Unwind_GetGR(ctx, index) as usize })
+    }
+
     pub fn symbol_address(&self) -> *mut c_void {
         if let Frame::Cloned { symbol_address, .. } = *self {
             return symbol_address;
@@ -153,6 +163,13 @@ mod uw {
             not(all(target_os = "linux", target_arch = "arm"))
         ))]
         pub fn _Unwind_GetIP(ctx: *mut _Unwind_Context) -> libc::uintptr_t;
+
+        #[cfg(all(
+            not(all(target_os = "android", target_arch = "arm")),
+            not(all(target_os = "freebsd", target_arch = "arm")),
+            not(all(target_os = "linux", target_arch = "arm"))
+        ))]
+        pub fn _Unwind_GetGR(ctx: *mut _Unwind_Context, index: u32) -> libc::size_t;
 
         #[cfg(all(
             not(target_os = "android"),
--- a/src/backtrace/mod.rs
+++ b/src/backtrace/mod.rs
@@ -101,6 +101,14 @@ impl Frame {
     pub fn symbol_address(&self) -> *mut c_void {
         self.inner.symbol_address()
     }
+
+    /// Returns the register value at this frame.
+    ///
+    /// The general purpose register value may not be available. The register
+    /// index differs between platforms.
+    pub fn gr(&self, index: u32) -> Option<usize> {
+        self.inner.gr(index)
+    }
 }
 
 impl fmt::Debug for Frame {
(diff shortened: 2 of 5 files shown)


src/backtrace/libunwind.rs
```diff
@@ -51,6 +51,16 @@ impl Frame {
         unsafe { uw::_Unwind_GetIP(ctx) as *mut c_void }
     }
 
+    pub fn gr(&self, index: u32) -> Option<usize> {
+        let ctx = match *self {
+            Frame::Raw(ctx) => ctx,
+            Frame::Cloned { .. } => {
+                return None;
```
alexcrichton, line 58: This seems like it may be a bit of a hazard at  some point, but I think it's fine for now.

src/backtrace/mod.rs
```diff
@@ -101,6 +101,14 @@ impl Frame {
     pub fn symbol_address(&self) -> *mut c_void {
         self.inner.symbol_address()
     }
+
+    /// Returns the register value at this frame.
+    ///
+    /// The general purpose register value may not be available. The register
+    /// index differs between platforms.
+    pub fn gr(&self, index: u32) -> Option<usize> {
```
alexcrichton, line 109: Could this perhaps be renamed to `general_register`? (or something similar)

The parameter/return value here is also pretty interesting since it's highly platform-specific. What about doing something like this:

```rust
// src/lib.rs
#[cfg(target_arch = "x86_64")]
pub mod x86_64 {
    pub struct Rax;
    // ...
    impl Register for Rax { /* ... */ }
}
// ...

impl Frame {
    pub fn general_register<T: Register>(&self) -> T::Value { /* ... */ }
}

pub trait Register {
    type Value;
    #[doc(hidden)] // internal details
    fn get(frame: &Frame) -> Self::Value;
}
```

That way we can extend this to all kinds of register (simd/32-bit/64-bit/etc), highlight how it's platform-specific (module names), and avoid the need for figuring fiddling with indices which are sourced from elsewhere

Amanieu: It would be better to not make this generic and just use a `#[non_exhaustive]` enum for the registers.

alexcrichton: @Amanieu I don't understand what you mean, can you elaborate?

Amanieu: Basically something like this:
```rust
#[cfg(target_arch = "x86_64")]
pub mod x86_64 {
    #[non_exhaustive]
    pub enum GeneralReg {
        Rax,
        ...
    }
    
    impl RegisterClass for GeneralReg { ... }
}

impl Frame {
    pub fn register<T: RegisterClass>(&self, reg: T) -> T::Value { /* ... */ }
}

pub trait RegisterClass {
    type Value;
    #[doc(hidden)] // internal details
    fn get(&self, frame: &Frame) -> Self::Value;
}
```

alexcrichton: That seems largely the same? Is there a reason to prefer one over the other?

Amanieu: Using an `enum` allows for useful things like iterating over all available registers and printing them all. Doing this with only generics is not as ergonomic

```rust
use x86_64::GeneralReg::*;
for r in [Rax, Rbx, Rcx, ...] {
}
```

alexcrichton: Seems reasonable to me!

Could even add a cross-platform:

```rust
impl GeneralReg {
    pub fn all() -> impl Iterator<Item = GeneralReg> {
        // .. 
    }
}
```
or some sort of cross-platform iterator for "iterate over all the regs"

src/backtrace/libunwind.rs
```diff
@@ -51,6 +51,16 @@ impl Frame {
...
+            }
+        };
+        Some(unsafe { uw::_Unwind_GetGR(ctx, index) as usize })
```
bjorn3, line 61: This is not save: https://github.com/rust-lang/backtrace-rs/pull/291#issuecomment-621207213

> https://github.com/gcc-mirror/gcc/blob/a7201a085cc30f89944931d8fb1d7936f02a169f/libgcc/unwind-dw2.c#L249
>
> ```
> /* This will segfault if the register hasn't been saved.  */
> ```
>
> It seems that `_Unwind_GetGR` is only allowed to be called with saved registers. This means that at least the approach taken in this PR is wrong.

workingjubilee:
Per https://github.com/rust-lang/backtrace-rs/issues/288 this is no longer believed to be plausible to do without hazard so I am closing this PR. Thank you for your efforts, and feel free to open a new PR if you find a way to do this relatively soundly.

PR #315, opened 2020-05-05
WIP: Switch from `goblin` to `object`

alexcrichton:
This is a PR using @philipc's [work](https://github.com/gimli-rs/object/issues/215#issuecomment-622917687) to switch the `goblin` dependency of the `gimli-symbolize` feature to the `object` crate instead. I explained some rationale for this [here](https://github.com/gimli-rs/object/issues/215#issue-604046969) but the general idea is that we want a lightweight dependency for parsing object files and `goblin` is a bit heavyweight with its custom derive.

I mostly just wanted to open this up early-ish to verify it works across all platforms.


alexcrichton:
I should also note that this depends on https://github.com/gimli-rs/object/pull/216

alexcrichton:
Done in https://github.com/rust-lang/backtrace-rs/pull/317

PR #317, opened 2020-05-08
Remove dladdr fallback and implementation

alexcrichton:
This has been present for a very long time but I believe this has never
actually been that necessary. Non-MSVC platforms all use libbacktrace by
default, and libbacktrace will consult symbol tables of object files to
do what `dladdr` does, just inside of libbacktrace. Additionally gimli
implements the same logic. I believe that this means that `dladdr` isn't
necessary for resolving any symbols since our other strategies should
already be doing everything for us.

This commit makes the feature defunkt and otherwise removes the various
forms of fallback to dladdr.


alexcrichton:
Looks like the gimli implementation doesn't handle fat objects just yet, so that'll need to be handled first.

src/symbolize/gimli.rs
```diff
@@ -223,153 +245,303 @@ cfg_if::cfg_if! {
...
+            //
+            // The general idea is that the `bias` plus a segment's
+            // `stated_virtual_memory_address` is going to be where in the
+            // actual address space the segment resides. The other thing we rely
+            // on though is that a real address minus the `bias` is the index to
+            // look up in the symbol table and debuginfo.
+            //
+            // It turns out, though, that for system loaded libraries these
+            // calculations are correct. For native executables, however, it
+            // appears correct. Lifting some logic from LLDB's source it has
```
philipc, line 412: Should one of these `correct` be `incorrect`?

alexcrichton: Oops, yes indeed.

alexcrichton:
Ok, so I'm pretty confident that dladdr is not needed, so I'm going to merge this. When I originally added the dladdr implementation it was *ages* ago (about the first time I wrote all this). I'm pretty certain that I was calling libbacktrace correctly, misleading me to believe that it was often needed to call dladdr. Nowadays I'm pretty sure that libbacktrace and gimli do everything necessary to implement dladdr, making calling it unnecessary.

We may have a few edge cases to work through in the implementation of parsing symbol tables with gimli (or something like that), but it's something we can handle in time. And of course adding dladdr back in is pretty easy to do.

In any case this has also snowballed a bit to pick up the `object` switch from `goblin` as well as improvements targeted at macOS for parsing fat libraries and better handling of symbol tables in system libraries.

Thanks again @philipc for writing the patch to switch to `object` and helping out with the crate release and PRs merging!

PR #330, opened 2020-05-13
Avoid panicking on overflow due to bias wrapping

### LINKED ISSUE #329: gimli.rs: attempt to add with overflow
I experienced the following error after updating an application to use backtrace-rs v0.3.47:

- **Message:**  attempt to add with overflow
- **Location:** /builder/home/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.47/src/symbolize/gimli.rs:413

Which looks to be here:

https://github.com/rust-lang/backtrace-rs/blob/4578e94/src/symbolize/gimli.rs#L413

Unfortunately this error is occurring in a production build which consumes exclusive access to a hardware resource, so I wasn't able to dig into it much further beyond that.

It also hung if I tried to set `RUST_BACKTRACE` to either `1` or `full`, I assume because of the circular dependency of obtaining a backtrace in the event the backtrace code crashes...

~~I will double check I'm doing a release build as I'm a bit confused why the arithmetic is checked otherwise as it's not doing a `checked_add`.~~

Aha, this particular application enables checked arithmetic in release builds explicitly:

```toml
[profile.release]
overflow-checks = true
```

alexcrichton:
It's not entirely clear to me if this'll fix the issue, or why this
issue is coming up, but maybe some biases are negative in some systems?

Closes #329

### DIFF
--- a/src/symbolize/gimli.rs
+++ b/src/symbolize/gimli.rs
@@ -408,10 +408,18 @@ impl Cache {
                 // First up, test if this `lib` has any segment containing the
                 // `addr` (handling relocation). If this check passes then we
                 // can continue below and actually translate the address.
+                //
+                // Note that we're using `wrapping_add` here to avoid overflow
+                // checks. It's been seen in the wild that the SVMA + bias
+                // computation overflows. It seems a bit odd that would happen
+                // but there's not a huge amount we can do about it other than
+                // probably just ignore those segments since they're likely
+                // pointing off into space. This originally came up in
+                // rust-lang/backtrace-rs#329.
                 if !lib.segments.iter().any(|s| {
-                    let svma = s.stated_virtual_memory_address as usize;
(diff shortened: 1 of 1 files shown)


alexcrichton:
@tarcieri are you up for testing this out before it's merged/released? I'd like to confirm that it actually fixes the issue you're seeing. Additionally, if possible, would you be up for running a build with a bit more logging to try to dig into what's happening under the hood here?

tarcieri:
@alexcrichton sure, I can do either of those

alexcrichton:
Ok the first commit here in theory should fix the issue and the second commit should have some debug logging if you're ok testing that out

tarcieri:
It's no longer crashing 🎉 

I'll follow up with the debug output separately.

alexcrichton:
Thanks again for the assistance!

PR #338, opened 2020-05-28
sys: Split dep-of-std and symbol prefix features

roblabla:
When building a libstd port out-of-tree using [xargo's staged build system](https://github.com/japaric/xargo/#multi-stage-builds), libstd dependencies are allowed to rely on the presence of libcore, compiler-builtins, etc... in the sysroot. Such builds will thus want to remove the `rustc-dep-of-std` feature from libstd dependencies. Doing so for backtrace-sys, however, risks causing duplicate symbols between the sysroot and normal libraries. This is because backtrace-sys links in the libbacktrace library, and has some special logic associated with the `rustc-dep-of-std` feature to rename the symbols.

In order to better support such libstd crates, this commit splits the symbol renaming logic into its own feature, and makes that feature a dependency of rustc-dep-of-std.

### DIFF
--- a/crates/backtrace-sys/Cargo.toml
+++ b/crates/backtrace-sys/Cargo.toml
@@ -24,4 +24,8 @@ default = ["backtrace-sys"]
 
 # Without this feature, this crate does nothing.
 backtrace-sys = []
-rustc-dep-of-std = ['core', 'compiler_builtins']
+# Compile backtrace-sys for libstd
+rustc-dep-of-std = ['core', 'compiler_builtins', 'rustc-dep-of-std-prefix']
+# Change the prefix of the embedded libbacktrace symbols to something unique to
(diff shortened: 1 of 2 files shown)


alexcrichton:
I'd ideally prefer to not add further complications for rustc's own build system to this crate if possible. Since the crates.io default uses gimli now, is this perhaps no longer needed?

roblabla:
I don't think gimli is usable in no_std context, which would be necessary for integration into a libstd port. `gimli-symbolize` feature [pulls the std feature](https://github.com/rust-lang/backtrace-rs/blob/master/Cargo.toml#L74).

alexcrichton:
Er yes my point though is that users of `backtrace` on crates.io typically have std so there's no need for the symbol switcharoo. It's pretty unlikely now that libbacktrace is used both through libstd and through crates.io. 

Are you seeing link errors locally from multiply defined symbols?

PR #344, opened 2020-06-10
Handle compressed debug sections in ELF files

### LINKED ISSUE #342: gimli-symbolize fails with compressed debug info
I've put together a full reproduction, with commentary, here: https://github.com/benesch/rust-zdebuginfo

The gist is that Rust generates an absolutely massive amount of debug info by default. (For a deep dive into why, see https://github.com/MaterializeInc/materialize/pull/2691.) One easy workaround is to compress the debug info sections, which you can accomplish on Linux with a `.cargo/config` like the following:

```toml
rustflags = ["-C", "link-arg=-Wl,--compress-debug-sections=zlib-gabi"]
```

Unfortunately, with the new `gimli-symbolizer` feature, backtraces produced by such a binary fail to symbolicate, though the libbacktrace-based implementation handles them fine.

benesch:
@alexcrichton this will need to wait for the `unaligned` feature to land in an actual release of the object crate, but otherwise it's ready to go from my perspective. Let me know what you think! Don't think there's a good way around the `flate2` dependency (I guess we could depend on mini_oxide directly?). Do you think that'll be a showstopper for inclusion in libstd?

----

ELF files allow debug info sections to be compressed. The libbacktrace
backed supported these compressed sections, but the Gimli backend did
not. This commit adds that support to the Gimli backend.

In my tests these debug info sections do not obey the alignment
requirements that the object crate expects for the gABI compression
header (nor can I find a source documenting any alignment requirements),
so this commit additionally enables the "unaligned" feature in the
upcoming version of the object crate.

There is a bit of unsafe to ensure the lifetime of the decompressed
sections matches the lifetime of the mmap'd file. I don't think there is
a way around this unsafe code, unless we are willing to ditch Gimli's
EndianSlice for an (apparently slower) EndianReader backed by a
Cow<[u8]>.

Fix #342.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -78,6 +78,10 @@ jobs:
     - run: cargo test --features gimli-symbolize --manifest-path crates/without_debuginfo/Cargo.toml
     - run: cargo test --manifest-path crates/line-tables-only/Cargo.toml --features libbacktrace
     - run: cargo test --manifest-path crates/line-tables-only/Cargo.toml --features gimli-symbolize
+    - run: RUSTFLAGS="-C link-arg=-Wl,--compress-debug-sections=zlib-gabi" cargo test --features gimli-symbolize
+      if: contains(matrix.os, 'ubuntu')
+    - run: RUSTFLAGS="-C link-arg=-Wl,--compress-debug-sections=zlib-gnu" cargo test --features gimli-symbolize
+      if: contains(matrix.os, 'ubuntu')
 
   windows_arm64:
     name: Windows AArch64
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,11 +39,12 @@ compiler_builtins = { version = '0.1.2', optional = true }
 # Optional dependencies enabled through the `gimli-symbolize` feature, do not
 # use these features directly.
 addr2line = { version = "0.12.0", optional = true, default-features = false }
+miniz_oxide = { version = "0.3.7", optional = true }
 [dependencies.object]
-version = "0.19"
+git = "https://github.com/gimli-rs/object.git"
 optional = true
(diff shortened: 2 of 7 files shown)


philipc:
> That alignment is something we could respect with the alloc family of functions, but it's not worth dealing with it if we're using unaligned anyway.

Agree that we don't need to check `ch_addralign`, but it's because `gimli` doesn't care about the alignment, not due to the `object`'s `unaligned` feature. DWARF sections shouldn't require alignment anyway and we don't currently check `sh_addralign` for uncompressed DWARF sections either.

alexcrichton:
d'oh, right!

I suppose that because this seems like a general ELF thing that maybe non-dwarf sections may need more alignment, but for our use cases definitely doesn't matter much!

benesch:
👍 I've removed the check on `ch_addralign`. I also switched the decompression from flate2 to miniz_oxide directly. I _think_ that's everything! Please double check the unsafe code, of course.

alexcrichton:
Looks great to me, thanks so much again for your help on this!

benesch:
Absolutely! Thanks very much for the quick review and merge!

It occurs to me that the master branch of backtrace now depends on the `object` crate from Git, rather than crates.io. This is certainly fine with me, but not sure if you meant to merge before getting an official release of `object` with the `unaligned` feature, so wanted to surface!

alexcrichton:
Ah yeah I'm not too worried about that.

@philipc would you be ok making a new publish of `object` in the near future?

philipc:
Yep I can do that.

philipc:
`object` 0.20.0 is published.

alexcrichton:
Thanks!

This support is now published with 0.3.49

PR #347, opened 2020-06-15
Add the current executables directory to the search path

wickerwaka:
This change configures the dbghelp.dll search path so that it includes the directory of the running executable.

The default search path is the working directory, followed by the contents on the _NT_SYMBOL_PATH and _NT_ALTERNATE_SYMBOL_PATH environment variables. (https://docs.microsoft.com/en-us/windows/win32/api/dbghelp/nf-dbghelp-syminitialize). However it is common to deploy windows executables with the .exe and .pdb file side-by-side. This works fine as long as these files are in the working directory, but symbol resolution will fail when run from any other directory.

### DIFF
--- a/src/dbghelp.rs
+++ b/src/dbghelp.rs
@@ -73,6 +73,12 @@ mod dbghelp {
             pdwDisplacement: PDWORD,
             Line: PIMAGEHLP_LINEW64,
         ) -> BOOL;
+        pub fn SymGetSearchPathW(
+            hProcess: HANDLE,
+            SearchPath: PWSTR,
+            SearchPathLength: DWORD,
+        ) -> BOOL;
+        pub fn SymSetSearchPathW(hProcess: HANDLE, SearchPath: PCWSTR) -> BOOL;
     }
 
     pub fn assert_equal_types<T>(a: T, _b: T) -> T {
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks for the PR! Do you know if this happens for other similar implementations as well? Are there perhaps links to point to prior art?

Additionally, could a test be added for this on CI?

wickerwaka:
I don't have any prior art to point to, unfortunately. I've done this on some internal projects, but searching on github didn't turn up anything useful.

The specific case I encountered was that I could get a backtrace locally, but another user could not. It turns out it was working for me because I had built the exe locally so it was able to used to absolute path to the pdb to resolve the symbols, even though I wasn't running from the directory with the .pdb file in it. But that took a few hours to work out:).....

I can add a CI test. It will require some scripting because it will need to compile an executable, move the executable and the .pdb to a new location and then test it. Are there any existing examples of scripting for the windows CI tasks?

alexcrichton:
Hm if this is something that isn't done elsewhere, should this perhaps be deferred for external applications to configure? I'm not sure how common it is to do this myself, but I'm no Windows guru.

Is there perhaps documentation, though, on how PDB files are conventionally distributed? If it turns out that all Windows applications are deployed one way then it seems reasonable to try to support this.

For tests I think it's fine to have some scripting on CI, so long as it's on CI somewhere I don't mind too much.

Issue #328, opened 2020-05-12
Enable gimli-symbolize to be used by libstd 


alexcrichton:
Currently the `gimli-symbolize` feature also enables the `std` feature. This means that it's not suitable for inclusion into the standard library. This also means that now that we've switched this crate to using gimli by default the standard library may languish! All-in-all, let's put the final nail in the coffin of libbacktrace and switch to gimli by enabling libstd to use gimli.

Some initial words have been said about this on https://github.com/rust-lang/backtrace-rs/issues/189, but the general state of the world is that the gimli implementation *in this crate only* uses a bunch of functions from `std`. This ranges from:

* `std::env::current_exe()` - used to parse own debuginfo
* `std::fs::*` - used to find, locate, and open debuginfo. Often the current executable, macOS is much more involved though
* `std::{path,ffi}::*` - manipulating paths to get passed around in various places, along with reading C strings on Linux.

I don't think it is currently clear at this time how we're going to implement this. There's a few possible strategies that have been proposed so far:

1. Vendor an implementation of everything into this crate. This would involve duplicating code from libstd into this crate, calling all the raw syscalls ourselves. One major downside of this is that as the debuginfo-finding process gets more complicated this syscall layer is undoubtedly going to get more complicated. Additionally `std::env::current_exe()` is already significantly complicated to run on all platforms.
2. Have some sort of trait which libstd implements. The standard library would then pass down its implementation into this crate. This trait would be a bit of a pain to maintain over time, however.
3. Include this library as a submodule into libstd. This is actually much easier after the recent rejiggering of features, so this may be a a somewhat plausible route. It would involve sync'ing dependencies with libstd, however, and that may not be trivial over time. Another downside of this approach is that it's somewhat difficult to maintain this style of coding over time. It's pretty nonstandard and would likely turn away possible contributors.
4. Split the `sys` module out of libstd into a separate crate. The `backtrace` crate would then un-stably depend on this `sys` module. This is a huge change for libstd, however, and highly unlikely to happen.

I'm definitely open to more ideas if folks have them :)

alexcrichton:
One thing I should also mention though is that we likely want to let the gimli support bake on crates.io for some time before actually moving on this. There's undoubtedly various porting issues and/or bugs that will arise, and it'd be great to weed those out before we move into libstd.

fitzgen:
FWIW, option (2) seems like the best course of action to me, based on gut feeling.

alexcrichton:
[...]

PR #349, opened 2020-06-16
Prepare for this crate to go into libstd

alexcrichton:
This commit is a preparation of this crate to be included as a submodule
into the standard library. I'm not 100% sold on this yet but I'm
somewhat convinced that this is going to happen this way. This is
progress on #328 and a preview of what it might look like to implement
this strategy.

Currently I don't plan to merge this to the `master` branch unless it's
decided to move forward with this integration strategy of the
gimli feature of the backtrace crate.


alexcrichton:
Ok looks like this is all really happening, so I'm gonna merge.

PR #354, opened 2020-06-29
Bump miniz-oxide version

alexcrichton:


matklad:
@alexcrichton would appreciate a release of `backtrace` with this (otherwise we are getting two miniz_oxides in rust-analyzer. I obviously love that crate, but only once :D )

alexcrichton:
Published!

PR #359, opened 2020-07-19
Use noop symbolizer for target_env = "devkita64"

leo60228:
This is necessary for a WIP Rust target that I can hopefully upstream the rest of relatively soon.

The target (Nintendo Switch homebrew) doesn't have a standard way to distribute debug info, so I haven't implemented a full symbolizer. I had a prototype using <https://github.com/MegatonHammer/linkle>'s DWARF support, but I'd like to think about that some more before baking it into libstd.


leo60228:
I've decided to use linkle's DWARF support instead, so I'm making a new PR for that.

PR #361, opened 2020-07-20
Support Linkle's DWARF implementation for target_env = "devkita64"

leo60228:
This target is added by https://github.com/rust-lang/rust/pull/74567. Note that this requires support from the program to work properly, which is done in libstd by the aforementioned PR.

### DIFF
--- a/crates/backtrace-sys/build.rs
+++ b/crates/backtrace-sys/build.rs
@@ -14,7 +14,8 @@ fn main() {
         target.contains("hermit") ||
         target.contains("wasm32") ||
         target.contains("fuchsia") ||
-        target.contains("uclibc")
+        target.contains("uclibc") ||
+        target.contains("devkita64")
     {
         println!("cargo:rustc-cfg=empty");
         return;
--- a/crates/backtrace-sys/build.rs
+++ b/crates/backtrace-sys/build.rs
@@ -14,8 +14,7 @@ fn main() {
         target.contains("hermit") ||
         target.contains("wasm32") ||
         target.contains("fuchsia") ||
-        target.contains("uclibc") ||
-        target.contains("devkita64")
+        target.contains("uclibc")
     {
         println!("cargo:rustc-cfg=empty");
         return;
--- a/src/symbolize/gimli.rs
+++ b/src/symbolize/gimli.rs
@@ -31,8 +31,11 @@ extern crate std as mystd;
 #[cfg(windows)]
 #[path = "gimli/mmap_windows.rs"]
 mod mmap;
-#[cfg(unix)]
+#[cfg(all(unix, not(target_env = "devkita64")))]
 #[path = "gimli/mmap_unix.rs"]
+mod mmap;
+#[cfg(target_env = "devkita64")]
+#[path = "gimli/mmap_fake.rs"]
 mod mmap;
 mod stash;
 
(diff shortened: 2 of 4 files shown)


crates/backtrace-sys/build.rs
```diff
@@ -14,7 +14,8 @@ fn main() {
         target.contains("hermit") ||
         target.contains("wasm32") ||
         target.contains("fuchsia") ||
-        target.contains("uclibc")
+        target.contains("uclibc") ||
+        target.contains("devkita64")
```
alexcrichton, line 18: I think this isn't necessary any more since this crate isn't built by default?

src/symbolize/gimli.rs
```diff
@@ -31,9 +31,12 @@ extern crate std as mystd;
 #[cfg(windows)]
 #[path = "gimli/mmap_windows.rs"]
 mod mmap;
-#[cfg(unix)]
+#[cfg(all(unix, not(target_env = "devkita64")))]
 #[path = "gimli/mmap_unix.rs"]
 mod mmap;
+#[cfg(target_env = "devkita64")]
+#[path = "gimli/mmap_fake.rs"]
+mod mmap;
```
alexcrichton, line 39: While you're here, mind reorganizing this to be in a `cfg_if!` to avoid the `#[cfg(not(...))]`?

src/symbolize/gimli.rs
```diff
@@ -487,7 +518,7 @@ impl Cache {
                 if !lib.segments.iter().any(|s| {
                     let svma = s.stated_virtual_memory_address;
                     let start = svma.wrapping_add(lib.bias);
-                    let end = start.wrapping_add(s.len);
+                    let end = start.saturating_add(s.len);
```
alexcrichton, line 521: How come this changed from wrapping to saturating?

src/symbolize/gimli/mmap_fake.rs
```diff
@@ -0,0 +1,24 @@
...
+use alloc::vec::Vec;
+use core::ops::Deref;
+
+pub struct Mmap {
+    vec: Vec<u8>,
+}
+
+impl Mmap {
+    pub unsafe fn map(mut file: &File, len: usize) -> Option<Mmap> {
+        let mut mmap = Mmap { vec: Vec::new() };
```
alexcrichton, line 11: Perhaps this could do `with_capacity` to preallocate? Also I think it's fine to avoid the `truncate` since this is only ever used for reading the whole file anyway.

src/symbolize/mod.rs
```diff
@@ -467,6 +467,7 @@ cfg_if::cfg_if! {
         not(target_os = "fuchsia"),
         not(target_os = "emscripten"),
         not(target_env = "uclibc"),
+        not(target_env = "devkita64"),
```
alexcrichton, line 470: I think this isn't necessary since libbacktrace is off-by-default now?

src/symbolize/gimli.rs
leo60228: This calculation would overflow before due to the length being `usize::max_value()`.

src/symbolize/gimli/mmap_fake.rs
leo60228: I've done the latter. I was having issues with `map` getting passed a length *much* larger than the file during testing, so I didn't use a `with_capacity`. It didn't affect anything else, so I didn't debug much further, but I could try to do some more debugging on that.

src/symbolize/gimli.rs
alexcrichton: I believe this is specific to the new target added, probably from the usage of the maximum value as you mentioned. I think that should be adjusted  where the segment shouldn't have a maximum value if a bias is also set, in that case the length of the segment should be adjusted to be smaller or account for the bias.

src/symbolize/gimli/mmap_fake.rs
alexcrichton: Yes I think it would be good to debug that because this just calls `fs::metadata` effectively before calling this function, and the file shouldn't change that much in the meantime.

leo60228: I can't reproduce this anymore, so...

leo60228:
Was a bit confused until I realized rustfmt was picking up Rust's `rustfmt.toml` because I was running it from the submodule...

leo60228:
Here's a screenshot of this working, by the way:
image

alexcrichton:
Nice!

PR #364, opened 2020-07-24
Disable libbacktrace on libnx

leo60228:
This is necessary due to https://github.com/rust-lang/rust/pull/74613.


leo60228:
Oops, should've checked the diff after cherry-picking...

lzutao:
@leo60228 is this enable since gimli has been integrated to rustc again ?

PR #367, opened 2020-07-30
Include source column numbers, where available

### LINKED ISSUE #257: Support column information
Right now, `backtrace::Symbol` contains information about line and file name. This is all that libbacktrace supplies. However, gimli supports reading column information as well. `Symbol` should get a function `column(&self) -> Option<u32>` that, if gimli is being used, returns the column, otherwise None.

eggyal:
It'd be great if backtraces could include source column numbers, in addition to filename/line numbers (where available - currently only gimli).

Closes #257

### DIFF
--- a/src/print.rs
+++ b/src/print.rs
@@ -242,7 +242,12 @@ impl BacktraceFrameFmt<'_, '_, '_> {
         Ok(())
     }
 
-    fn print_fileline(&mut self, file: BytesOrWideString<'_>, line: u32, colno: Option<u32>) -> fmt::Result {
+    fn print_fileline(
+        &mut self,
+        file: BytesOrWideString<'_>,
+        line: u32,
+        colno: Option<u32>,
+    ) -> fmt::Result {
         // Filename/line are printed on lines under the symbol name, so print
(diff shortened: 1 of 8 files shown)


src/print.rs
```diff
@@ -187,6 +190,7 @@ impl BacktraceFrameFmt<'_, '_, '_> {
         symbol_name: Option<SymbolName<'_>>,
         filename: Option<BytesOrWideString<'_>>,
         lineno: Option<u32>,
+        colno: Option<u32>,
```
alexcrichton, line 193: This is unfortunately breaking a public-facing API, so could this add a new function instead (like `print_raw_with_column`) and have this function delegate to that one?

eggyal:
Thanks @alexcrichton — have added tests and resolved the breaking change to public API as suggested.  CI appears to have failed due to (presumably transient) network failure?  How can I request a retry?

alexcrichton:
Ok retried, but I think there's a legitimate error now

eggyal:
Should be fixed now, but looks like another network failure?  🤪

alexcrichton:
Thanks again!

PR #368, opened 2020-07-31
Feature: module base address added to the Frame

dmitry-zakablukov:
The base address of the module is required on the Windows to decode a backtrace when a PDB file is moved or missing:
```
<address_in_pdb> = <instruction_pointer_address> - <module_base_address>
```

The struct `Frame` is extended with the following method:
```
pub fn module_base_address(&self) -> Option<*mut c_void>;
```

The method returns an Option, because only `dbghelp::Frame::module_base_address()` is implemented in this PR. And I'm not sure if the proposed method can be implemented for `libunwind::Frame`.

### DIFF
--- a/crates/without_debuginfo/tests/smoke.rs
+++ b/crates/without_debuginfo/tests/smoke.rs
@@ -27,3 +27,18 @@ fn all_frames_have_symbols() {
         assert_eq!(missing_symbols, 0);
     }
 }
+
+#[test]
+fn all_frames_have_module_base_address() {
+    let mut missing_base_addresses = 0;
+    backtrace::trace(|frame| {
+        if frame.module_base_address().is_none() {
+            missing_base_addresses += 1;
+        }
+        true
+    });
+
+    if cfg!(windows) {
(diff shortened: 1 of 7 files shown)


alexcrichton:
Do you know if it's possible to implement this for any other platform?

dmitry-zakablukov:
@alexcrichton , it may be done with `dladdr()` function: https://man7.org/linux/man-pages/man3/dladdr.3.html
`Dl_info` structure has `dli_fbase` field.

It should be tested for executable files though, but I think they have fixed load address.

alexcrichton:
Ok, this seems reasonable to implement for now and it can always be implemented on unix in the future. Thanks again for the PR!

dmitry-zakablukov:
@alexcrichton , may I ask you when the next version of backtrace crate will be released? The latest one (0.3.50) doesn't contain my changes.

alexcrichton:
Ah I generally just do a release when requested, so I just published 0.3.51

dmitry-zakablukov:
@alexcrichton , thank you!

PR #371, opened 2020-09-19
Fix compilation error on i686-apple-darwin targets

CouleeApps:
`symbolize::gimli::macho::Object` expects `syms` to be a `Vec<(..., u64)>` but on i686 it gets initialized with a `Vec<(..., u32)>`. This just adds a cast to `u64` to the syms generator so it typechecks.


alexcrichton:
Thanks! Could this use `u64::from` or `.into()` perhaps to emphasize that it's lossless?

CouleeApps:
Both compile and work but `u64::into` seems more obvious so I'll update the PR to use that.

alexcrichton:
Thanks!

PR #373, opened 2020-10-08
use NativeEndian in symbolize::gimli::Context

cuviper:
`Object` uses `NativeEndian`, so the `Context` should too.

Cc: https://github.com/rust-lang/rust/issues/77410

### DIFF
--- a/src/symbolize/gimli.rs
+++ b/src/symbolize/gimli.rs
@@ -5,7 +5,7 @@
 //! intended to wholesale replace the `libbacktrace.rs` implementation.
 
 use self::gimli::read::EndianSlice;
-use self::gimli::LittleEndian as Endian;
+use self::gimli::NativeEndian as Endian;
 use self::mmap::Mmap;
 use self::stash::Stash;
 use super::BytesOrWideString;


src/symbolize/gimli.rs
```diff
@@ -5,7 +5,7 @@
 //! intended to wholesale replace the `libbacktrace.rs` implementation.
```
camelid, line 5: Should this comment be updated? Isn't gimli the default now? (Not necessary to do here, just wondering.)

cuviper:
I do see big-endian `powerpc64` in CI, so I'm not sure how this has been passing -- does it really **run** the tests?
- before: https://github.com/rust-lang/backtrace-rs/runs/1183751745?check_suite_focus=true#step:6:897
- now: https://github.com/rust-lang/backtrace-rs/pull/373/checks?check_run_id=1228634621#step:6:978

I used native s390x to confirm the problem and this fix.

cuviper:
Oh:
https://github.com/rust-lang/backtrace-rs/blob/2b51e00670927b58255f92f2281367187a49f370/ci/docker/powerpc64-unknown-linux-gnu/Dockerfile#L13-L15

alexcrichton:
Thanks! You've confirmed locally this fixes the issue too?

I actually thought that ELF was always little-endian for common platforms, regardless of the endianness of the platform itself, but this would make sense why things are broken on big-endian!

philipc:
> I actually thought that ELF was always little-endian for common platforms, regardless of the endianness of the platform itself

I think that's true for PE but not for ELF.

cuviper:
> Thanks! You've confirmed locally this fixes the issue too?

On s390x, yes.

> I actually thought that ELF was always little-endian for common platforms, regardless of the endianness of the platform itself, but this would make sense why things are broken on big-endian!

AFAIK it's always native in ELF, and from a quick skim of glibc it looks like the loader asserts that they match:
https://sourceware.org/git/?p=glibc.git;a=blob;f=elf/dl-load.c;h=0c8fa72c4d1d2396f956ce3eb64c02b425b6fd22;hb=HEAD#l1713

Any recollection why that powerpc64 CI has the `qemu-ppc64` line commented out?

cuviper:
> I think that's true for PE but not for ELF.

Hmm, then should we make this import part of the platform-specific `cfg_if`?

philipc:
> Hmm, then should we make this import part of the platform-specific cfg_if?

This import seems to only be used for the DWARF, and I don't know what endianness DWARF in PE uses. My intuition is that `NativeEndian` would be correct for it too.

alexcrichton:
Ok sounds good! I'm gonna go ahead and merge this since it fixes an issue for rust-lang/rust, we can sort out CI afterwards if necessary. PE I think only affects Windows (right?) and while Gimli is used for MinGW targets I don't think that there's any big-endian Windows targets right now. 

> Any recollection why that powerpc64 CI has the qemu-ppc64 line commented out?

Ah I'm sure at some point in time in the past it was broken, but it was most likely a bug in qemu. If it works now seems good to uncomment!

mati865:
> I think that's true for PE but not for ELF. 

Correct, docs: https://docs.microsoft.com/en-us/cpp/build/arm64-windows-abi-conventions?view=vs-2019#endianness

There are no official Rust MinGW ARM targets **yet**.
There are C/C++ MinGW ARM/Aarch64 targets though and some folks would like to see official Rust target already.

alexcrichton:
I'm pretty sure arm/aarch64 are little endian...

nagisa:
> I'm pretty sure arm/aarch64 are little endian...

ARM-the-architecture can go both ways, though its not awfully useful as the ecosystem around it pretty much universally uses LE. [Context](https://rust-lang.zulipchat.com/#narrow/stream/238009-t-compiler.2Fmeetings/topic/.5Bweekly.20meeting.5D.202020-10-01.20.2354818/near/211992553).

PR #378, opened 2020-10-22
Enable Mach-O on iOS in gimli.

cutsoy:
Fixes for iOS rust-lang/rust#78184.

iOS uses the same Mach-O format as macOS does. I have confirmed that this fix works on `x86_64-apple-ios` (tier 2). I expect it will also work on `aarch64-apple-ios` (tier 2), `armv7-apple-ios`, `armv7s-apple-ios`, `i386-apple-ios`, `x86_64-apple-ios-macabi` (tier 3), but I don't have the hardware to test on all of those archs.

#### Tier 2:

- [x] `x86_64-apple-ios` compiles and works
- [x] `aarch64-apple-ios` compiles

#### Tier 3:
- [ ] `armv7-apple-ios`
- [ ] `armv7s-apple-ios`
- [ ] `i386-apple-ios`
- [ ] `x86_64-apple-ios-macabi`
- [ ] `aarch64-apple-tvos`
- [ ] `x86_64-apple-tvos`

#### Nightly?:
- [ ] `armv7k-apple-watchos`
- [ ] `i386-apple-watchos`
- [ ] `x86_64-apple-watchos`

(Thanks @bjorn3!)

### DIFF
--- a/src/symbolize/gimli.rs
+++ b/src/symbolize/gimli.rs
@@ -201,7 +201,12 @@ cfg_if::cfg_if! {
                 }],
             })
         }
-    } else if #[cfg(any(target_os = "macos", target_os = "ios"))] {
+    } else if #[cfg(any(
+        target_os = "macos",
+        target_os = "ios",
+        target_os = "tvos",
+        target_os = "watchos",
+    ))] {
         // macOS uses the Mach-O file format and uses DYLD-specific APIs to
         // load a list of native libraries that are part of the appplication.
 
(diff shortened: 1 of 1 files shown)


bjorn3:
I think watchos and all other operating systems of apple should be included too.

cutsoy:
Fixed! But I am assuming the target_os will be called `watchos` since there is no existing target spec in either stable or nightly as far as I can see.

alexcrichton:
Thanks!

PR #383, opened 2020-10-28
Some minor cleanup

lzutao:
No functionality changes intended.

### DIFF
--- a/src/symbolize/gimli.rs
+++ b/src/symbolize/gimli.rs
@@ -372,28 +372,32 @@ cfg_if::cfg_if! {
         fn native_libraries() -> Vec<Library> {
             let mut ret = Vec::new();
             unsafe {
-                libc::dl_iterate_phdr(Some(callback), &mut ret as *mut _ as *mut _);
+                libc::dl_iterate_phdr(Some(callback), &mut ret as *mut Vec<_> as *mut _);
             }
             return ret;
         }
 
+        // `info` should be a valid pointers.
+        // `vec` should be a valid pointer to a `std::Vec`.
         unsafe extern "C" fn callback(
             info: *mut libc::dl_phdr_info,
             _size: libc::size_t,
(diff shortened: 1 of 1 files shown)


src/symbolize/gimli.rs
```diff
@@ -372,28 +372,31 @@ cfg_if::cfg_if! {
         fn native_libraries() -> Vec<Library> {
             let mut ret = Vec::new();
             unsafe {
-                libc::dl_iterate_phdr(Some(callback), &mut ret as *mut _ as *mut _);
+                libc::dl_iterate_phdr(Some(callback), &mut ret as *mut Vec<_> as *mut _);
             }
             return ret;
         }
 
+        // SAFETY: `info` and `vec` should be valid pointers
```
alexcrichton, line 380: I'm not personally a huge fan of these comments because the "safety" here encompasses far more than what's mentioned here. I think it would be fine to document this without the "SAFETY" header which looks like it's trying to be exhaustive.

src/symbolize/gimli.rs
```diff
@@ -372,28 +372,31 @@ cfg_if::cfg_if! {
...
         unsafe extern "C" fn callback(
             info: *mut libc::dl_phdr_info,
             _size: libc::size_t,
             vec: *mut libc::c_void,
         ) -> libc::c_int {
+            let info = &*info;
             let libs = &mut *(vec as *mut Vec<Library>);
-            let name = if (*info).dlpi_name.is_null() || *(*info).dlpi_name == 0{
+            let is_main_prog = info.dlpi_name.is_null() || info.dlpi_name.read() == 0;
```
alexcrichton, line 388: I would prefer to use `*` here instead of a method to read the pointer.

lzutao: Will do although I personally prefer postfix deref operator, but we can't change it now.

PR #385, opened 2020-10-29
Update gimli docs

camelid:
I believe gimli is now the default implementation.

Cc https://github.com/rust-lang/backtrace-rs/pull/373#discussion_r502059176

### DIFF
--- a/src/symbolize/gimli.rs
+++ b/src/symbolize/gimli.rs
@@ -1,8 +1,6 @@
 //! Support for symbolication using the `gimli` crate on crates.io
 //!
-//! This implementation is largely a work in progress and is off by default for
-//! all platforms, but it's hoped to be developed over time! Long-term this is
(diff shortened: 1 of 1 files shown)


src/symbolize/gimli.rs
```diff
@@ -1,8 +1,6 @@
 //! Support for symbolication using the `gimli` crate on crates.io
 //!
-//! This implementation is largely a work in progress and is off by default for
-//! all platforms, but it's hoped to be developed over time! Long-term this is
-//! intended to wholesale replace the `libbacktrace.rs` implementation.
+//! This is the default symbolication implementation for Rust.
```
camelid, line 3: Let me know if this wording is incorrect, I am not very knowledgeable about gimli :)

alexcrichton:
Ah yes this is indeed accurate now, thanks!

PR #393, opened 2020-11-20
Deny warnings on CI to keep codebase warning-free

alexcrichton:
There's a ton of platforms here that we compile for and so it's only
really all covered on CI. To help keep warnings out for everyone let's
deny warnings on CI.


alexcrichton:
Er accidentally merged..

PR #396, opened 2020-11-24
disable dl_iterate_phdr for uclibc

ykoehler:


alexcrichton:
Thanks for this! Is this perhaps indicative of missing bindings in the `libc` crate though? Presumably uclibc has support for these functions?

ykoehler:
Exactly, in my required environment, the uClibc didn’t have those symbols.

We are not using the latest version of uClibc thought, as such, I am unclear how to deal with this.

alexcrichton:
Ok! In that case this seems fine for now and it can always be updated later if necessary.

tsidea:
Just fyi. The `dl_iterate_phdr` is present in the latest `uclibc` release [here](https://cgit.uclibc-ng.org/cgi/cgit/uclibc-ng.git/tree/include/link.h?h=v1.0.36&id=cbfdb7abf4790bac4da52b2161a12bf8abd9f174) and added in this [commit](https://cgit.uclibc-ng.org/cgi/cgit/uclibc-ng.git/commit/include/link.h?h=v1.0.36&id=2e3a6e76fcc56433a09b700985714393de02da5e). As such, I've managed to create a pull request for `libc` to include `dl_iterate_phdr` for `uclibc` [here](https://github.com/rust-lang/libc/pull/1967). I'm not sure if that changes anything with respect to this pull request, I just thought it might be relevant.

ykoehler:
ok, that implies I will need to upgrade our uClibc version in our toolchain/env, nightmares ahead...  but I understand.  Thanks

tsidea:
If just having the `dl_iterate_phdr` binding present in `libc` will not affect your code, then maybe you don't need to change. If `backtrace-rs` can work without `dl_iterate_phdr`, then you're probably all set and don't need to make any more changes. You can check whether this is true by building your rust compiler with the latest `libc`. Try using the latest `libc` by adding this patch `libc = { git = 'https://github.com/rust-lang/libc.git' }` in your rust compiler's working directory `Cargo.toml`, in the section `[patch.crates-io]`. Build the rust compiler and test your project in your environment and see what happens. If all is well then you don't need to upgrade just yet.

PR #401, opened 2020-12-31
Use the symbol table if the DWARF only has line numbers

philipc:
This can occur for skeleton units in split DWARF. Also bump addr2line to 0.14.1 to get a related fix there so that it returns the location in this case.

This fixes backtraces with `RUSTFLAGS="-Z split-dwarf=split -C save-temps"`.


alexcrichton:
Thanks!

PR #384, opened 2020-10-28
[Blocked on libc] enable backtrace for some BSD OSs

lzutao:
Waiting for new release of libc crate after <https://github.com/rust-lang/libc/pull/1953>.

Closes #325 and closes #326.

PR #402, opened 2021-01-01
gimli: Enable backtrace for FreeBSD 

### LINKED ISSUE #325: Gimli doesn't support FreeBSD
I believe all that's necessary to support FreeBSD is to upstream a definition of `dl_iterate_phdr` into the `libc` crate, after that support should be as simple as updating [this line](https://github.com/rust-lang/backtrace-rs/blob/bb5aa64804d6ba83a768c7ed1c5d67435ddba53e/src/symbolize/gimli.rs#L270-L273).

### LINKED ISSUE #326: Gimli doesn't support OpenBSD
I believe all that's necessary to support OpenBSD is to upstream a definition of `dl_iterate_phdr` into the `libc` crate, after that support should be as simple as updating [this line](https://github.com/rust-lang/backtrace-rs/blob/bb5aa64804d6ba83a768c7ed1c5d67435ddba53e/src/symbolize/gimli.rs#L270-L273).

cemeyer:
#384 attempt number 2.

Closes #325.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ exclude = ['crates/without_debuginfo', 'crates/macos_frames_test', 'crates/line-
 cfg-if = "1.0"
 rustc-demangle = "0.1.4"
 backtrace-sys = { path = "crates/backtrace-sys", version = "0.1.35", optional = true, default_features = false }
-libc = { version = "0.2.45", default-features = false }
+libc = { version = "0.2.81", default-features = false }
 
 # Optionally enable the ability to serialize a `Backtrace`, controlled through
 # the `serialize-*` features below.
--- a/src/symbolize/gimli.rs
+++ b/src/symbolize/gimli.rs
@@ -362,6 +362,7 @@ cfg_if::cfg_if! {
         any(
             target_os = "linux",
             target_os = "fuchsia",
+            target_os = "freebsd",
         ),
         not(target_env = "uclibc"),
     ))] {


cemeyer:
It is just @lzutao's earlier PR, rebased over the uclibc merge conflict.

alexcrichton:
Thanks! Is this ready to land with libc changes published?

cemeyer:
Rust newbie — how would I find out?  On #384 lzutao mentioned:

> This doesn't block on libc anymore.

alexcrichton:
To confirm, have you tested this on the platforms enabled here? If so I think the `libc` dependency's version should be bumped in this crate's version requirement because we'll require a newer version with the bindings.

cemeyer:
I have not. I can test on FreeBSD but I don’t have any of the other systems. Would that just be `cargo test`, or any additional manual testing? I can bump the libc requirement in cargo.toml.  Thanks!

alexcrichton:
Yes a `cargo test` wouold be great. I'm hesitant to enable platforms here unless they've been tested since this switch likely pulls in a big chunk of code.

cemeyer:
> Yes a `cargo test` wouold be great.

Ok, that was fast.  I think I must have already run it.  It passes.  However, I think the tests may need platform adjustment as well to actually cover this change.  Give me a few minutes...

> I'm hesitant to enable platforms here unless they've been tested since this switch likely pulls in a big chunk of code.

They're tier 2+ platforms, right?  I'm worried we might be applying higher scrutiny to restoring functionality that was lost in rust 1.47 than we were in losing it in the first place.  At the same time, I don't feel strongly about Open/NetBSD support and am ok dropping them if you insist.

alexcrichton:
My main goal is that if you're explicitly affecting new targets in this PR there should at least be confidence that the code compiles. It doesn't sound like you're checking that for all the platforms listed here? I would ideally like confirmation that the tests work with this PR. If there are other bugs then they should be filed as new issues.

cemeyer:
Ok, I’ll drop the non-FreeBSD diff. Cargo test passes.

alexcrichton:
Thanks!

cemeyer:
Thank you! :-)

PR #411, opened 2021-03-02
Haiku: add support for finding libraries and mapping library segments

nielx:
With the new native bindings in libc 0.2.87, it is now possible to iterate over all loaded images (the executable, shared libraries and libraries using dlopen()), and map the loaded segments to the right files, so that the debuginfo of the files can be used.

This adds stacktrace support for Haiku.


alexcrichton:
Thanks!

PR #415, opened 2021-04-09
wip: Android support in gimli

kjvalencik:
Depends on https://github.com/rust-lang/libc/pull/2144

`dl_iterate_phdr` was added in API Version 21; however, it's very old and I couldn't figure out how to build for it. The old NDK toolchains don't include prebuilt tools for `android21`.

I believe this will fail to link on API version 20 and below, but I'm not certain.

Any suggestions for the best way to handle it?


kjvalencik:
@alexcrichton Do you have any suggestions for handling the Android API version check? Options that I've thought of:

* Require users to enable it with a feature flag
* Port the existing check that uses `cc` in a `build.rs`
* Don't use `libc` and instead load the symbol dynamically with `libloading`

alexcrichton:
Thanks for this! I'd probably say that the second option, using the same `build.rs` check to see what the compiler api level reports, is the best for now. That way folks who want backwards-compat can get it and otherwise we'll automatically enable backtraces for more recent android versions (which should be basically all of them)

alexcrichton:
Looks great to me! I'm happy to merge once `libc` is published with the update.

kjvalencik:
Thanks again for the review @alexcrichton. `libc` has been published with the required function and the PR has been updated to point at the latest version.

alexcrichton:
Looks great to me, thanks!

s1341:
Awesome. Can we get a release so we can integrate this?

alexcrichton:
Sure thing, done now!

s1341:
Thanks!

s1341:
Confirmed working! Thanks all!

Issue #419, opened 2021-04-21
creating a `BacktraceFrame` from `Frame` 


glommer:
Hello

I am writing some code that would acquire a backtrace from a no-std environment (a signal handler). Therefore, I capture manually with `trace_unsynchronized`.

However, that's the only part of the process that needs to be `nostd`: after the backtrace is captured I can send it for later processing, at a time in which I have `std`.

It is pretty inconvenient to print the result, though, as a `BacktraceFrame` cannot be created from a `Frame`, and all the pretty printers require a `BacktraceFrame`.

Would it be possible to have a `From` implementation that converts them ?

PR #420, opened 2021-04-21
implement From<Frame> for BacktraceFrame

glommer:
There are situations where we can only capture raw `Frame` (example:
capturing from a signal handler), but it could still be that we can
process them later and are not fully nostd in the environment.

With a conversion from Frame to BacktraceFrame, this is trivial. But
without it, we can't really use the pretty printers as they are reliant
on BacktraceFrame, not Frame.

Fixes: #419

### DIFF
--- a/src/capture.rs
+++ b/src/capture.rs
@@ -536,7 +536,8 @@ mod tests {
     fn test_frame_conversion() {
         // captures an original backtrace, and makes sure that the manual conversion
         // to frames yields the same results.
-        let bt = Backtrace::new();
+        let mut bt = Backtrace::new();
+        bt.resolve();
         let original_frames = bt.frames();
 
         let mut frames = vec![];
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! Can you add some tests for this as well?

glommer:
I can, but I noticed that there aren't any tests in-file.
What kind of tests exactly are you looking for here?

alexcrichton:
Mostly just something that exercises this code. For example creating a list from a call to `backtrace::trace` and then after the `trace` is finished the frames are inspected.

glommer:
test added

src/capture.rs
```diff
@@ -518,3 +527,32 @@ mod serde_impls {
...
+    fn test_frame_conversion() {
+        // captures an original backtrace, and makes sure that the manual conversion
+        // to frames yields the same results.
+        let bt = Backtrace::new();
+        let original_frames = bt.frames();
+        let mut original = original_frames.iter();
+
+        let mut frames = 0;
+        unsafe {
+            crate::trace_unsynchronized(|frame| {
```
alexcrichton, line 545: Could this use `trace` instead of `trace_unsynchronized`? Additionally can this use the `BacktraceFrame` after the closure to ensure that it's not accidentally holding on to any ephemeral pointeres?

glommer: yeah, we can use `trace`. I ended up using the unsynchronized version because that is what I was using in my code, no problem.

By using the backtrace frame, maybe I can return a vector and do the asserts by zipping both iterators. Would that be enough for you ?

glommer: Thanks for your comments. I also noticed that the test would fail in `--release` mode, and indeed skipping a frame is fragile as inlining can change it.

I used a better method that now succeeds with release and debug.

A new version is force-pushed.

alexcrichton: Whatever testing is fine by me,  I mainly just want to confirm that if the `BacktraceFrame` leaks out of the closure then it's still safe to use.

glommer: Ok, just saw that. I bet it is a similar issue with `--release` and `--debug`. I will try it locally and get back to you

glommer: Ok. The issue with miri is different: every backtrace generated has completely different addresses!
Fortunately they resolve to the same thing, so I am testing with that

glommer: my god, lol...

At least your code coverage is pretty good!!

The test that fails now is about features disabled. If I understand your code correctly, that's because resolving symbols need features to run. I can remove the last assert. It won't be doing that much in that case, but it is better than disabling the whole test in the absence of features. At least the conversion path is stressed.

alexcrichton:
Looks good to me, thanks!

PR #421, opened 2021-05-02
Update gimli dependencies

philipc:


philipc:
I don't understand why the `object` dependency update is failing for `cargo test --features libbacktrace`, and I can't reproduce it locally.

alexcrichton:
I've bisected this as a "regression" which was introduced in https://github.com/rust-lang/rust/pull/84635. I don't think that PR is actually breaking. The "breakage" is probably because of https://github.com/rust-lang/cargo/pull/9418 which does something to reorder CGUs or symbols or something like that.

My suspicion is that this is a bug in libbacktrace itself. Locally testing resetting the libbacktrace submodule to the `master` branch (current master, which IIRC has a lot of changes) it fixes the issue. That leads me to conclude that this is indeed a libbacktrace issue.

Actually, I can confirm that a cherry-pick of https://github.com/ianlancetaylor/libbacktrace/commit/430dc8b6fe9826192db5b15a27fa1e6107bdc1cc fixes the issue.

---

So I think there's an interesting question of what to do with this at this point. The libbacktrace implementation of this crate at this point is long-unmaintained and it's falling behind in terms of features (compression, split-debuginfo, etc). I don't think there's any reason to turn it on over the gimli implementation now that gimli has baked for quite some time.

I think, personally, it's time to remove this implementation.

Would you be ok commenting out the libbacktrace tests in CI? That'll get this green and I can do the removal myself in a future PR.

philipc:
Done, thanks for looking into that.

PR #427, opened 2021-06-09
elf: support external and supplementary debuginfo

### LINKED ISSUE #410: gimli should support supplementary object files
Fixing the `gimli-symbolize` feature to handle supplementary object files shouldn't be a lot of effort:
- [x] `gimli` already supports them (https://github.com/gimli-rs/gimli/pull/552)
- [x] `addr2line` will need updating to read the units from the supplementary object file, and handle references to entries in it when looking up names (https://github.com/gimli-rs/addr2line/pull/208)
- [ ] `backtrace-rs` will need to read the `.gnu_debugaltlink` section to find and load the supplementary object file

Also, `backtrace-rs` doesn't currently follow `.gnu_debuglink` or `NT_GNU_BUILD_ID` with the `gimli-symbolize` feature. I assume that one or both of those are needed too.

philipc:
Fixes #410 

Doesn't check the CRC for GNU debuglink. I'm not sure how important this is, and it would require adding a dependency (`crc32fast`?).

Doesn't add any tests; I'm unsure how to test this in CI. I have manually tested that the `backtrace` example can give line number information for `__libc_start_main`, which tests the build ID on Fedora, and the GNU debuglink on Ubuntu. I have manually tested the supplementary object file support using the `backtrace` and `raw` examples after running the following commands:
```
objcopy --only-keep-debug target/debug/examples/backtrace target/debug/examples/backtrace.debug
strip -g target/debug/examples/backtrace
(cd target/debug/examples; objcopy --add-gnu-debuglink=backtrace.debug backtrace)

mkdir -p target/debug/examples/.debug
objcopy --only-keep-debug target/debug/examples/raw target/debug/examples/.debug/raw.debug
strip -g target/debug/examples/raw
(cd target/debug/examples/.debug; objcopy --add-gnu-debuglink=raw.debug ../raw)

mkdir -p .dwz
dwz -m .dwz/backtrace.debug -rh target/debug/examples/backtrace.debug target/debug/examples/.debug/raw.debug
```

### DIFF
--- a/src/symbolize/gimli/elf.rs
+++ b/src/symbolize/gimli/elf.rs
@@ -251,6 +251,8 @@ impl<'a> Object<'a> {
         None
     }
 
+    // The format of build id paths is documented at:
+    // https://sourceware.org/gdb/onlinedocs/gdb/Separate-Debug-Files.html
     fn build_id_debug_path(&self) -> Option<PathBuf> {
         const BUILD_ID_PATH: &[u8] = b"/usr/lib/debug/.build-id/";
         const BUILD_ID_SUFFIX: &[u8] = b".debug";
@@ -283,6 +285,8 @@ impl<'a> Object<'a> {
         Some(PathBuf::from(OsString::from_vec(path)))
     }
 
+    // The contents of the ".gnu_debuglink" section is documented at:
+    // https://sourceware.org/gdb/onlinedocs/gdb/Separate-Debug-Files.html
     fn gnu_debuglink_path(&self, path: &Path) -> Option<(PathBuf, u32)> {
         let section = self.section_header(".gnu_debuglink")?;
         let data = section.data(self.endian, self.data).ok()?;
@@ -297,6 +301,7 @@ impl<'a> Object<'a> {
         Some((path_debug, crc))
     }
 
+    // The format if the ".gnu_debugaltlink" section is based on gdb.
     fn gnu_debugaltlink_path(&self, path: &Path) -> Option<(PathBuf, &'a [u8])> {
         let section = self.section_header(".gnu_debugaltlink")?;
         let data = section.data(self.endian, self.data).ok()?;
@@ -329,6 +334,8 @@ fn decompress_zlib(input: &[u8], output: &mut [u8]) -> Option<()> {
     }
 }
 
+// Search order matches gdb, documented at:
+// https://sourceware.org/gdb/onlinedocs/gdb/Separate-Debug-Files.html
 fn locate_debuglink(path: &Path, filename: &[u8]) -> Option<PathBuf> {
     const DEBUG_PATH: &[u8] = b"/usr/lib/debug";
     let path = fs::canonicalize(path).ok()?;
--- a/src/symbolize/gimli/elf.rs
+++ b/src/symbolize/gimli/elf.rs
@@ -1,6 +1,6 @@
-use super::mystd::ffi::OsStr;
+use super::mystd::ffi::{OsStr, OsString};
 use super::mystd::fs;
-use super::mystd::os::unix::ffi::OsStrExt;
+use super::mystd::os::unix::ffi::{OsStrExt, OsStringExt};
 use super::mystd::path::{Path, PathBuf};
 use super::{Context, Mapping, Stash, Vec};
 use core::convert::{TryFrom, TryInto};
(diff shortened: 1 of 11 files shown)


src/symbolize/gimli/elf.rs
```diff
@@ -10,11 +15,150 @@ type Elf = object::elf::FileHeader32<NativeEndian>;
...
+                            return Some(mapping);
+                        }
+                    }
+                }
+            }
+        }
+
+        let stash = Stash::new();
+        let cx = Context::new(&stash, object)?;
+        Some(Mapping {
```
philipc, line 75: This no longer uses `Mapping::mk` because it needs to parse the `Object` beforehand. Not sure if there is any way to restructure this to allow that.

src/symbolize/gimli/elf.rs
```diff
@@ -10,11 +15,150 @@ type Elf = object::elf::FileHeader32<NativeEndian>;
...
+        let map = super::mmap(path)?;
+        let object = Object::parse(&map)?;
+
+        if let Some(_crc) = crc {
+            // TODO: check crc
+        }
+
+        // Try to locate a supplementary object file.
+        let tmp_stash = Stash::new();
+        if let Some(section) = object.section(&tmp_stash, ".gnu_debugaltlink") {
```
alexcrichton, line 96: Oh deear, I can't seem to find documentation on this... Do you know of some good online docs to link to?

I'm curious about why this is handled here rather than above from the original object...

src/symbolize/gimli/elf.rs
```diff
@@ -10,11 +15,150 @@ type Elf = object::elf::FileHeader32<NativeEndian>;
...
+            // only borrow `map` and `stash` and we're preserving them below.
+            cx: unsafe { core::mem::transmute::<Context<'_>, Context<'static>>(cx) },
+            _map: map,
+            _map_sup: None,
+            _stash: stash,
+        })
+    }
+}
+
+fn locate_debuglink(path: &Path, filename: &[u8]) -> Option<PathBuf> {
```
alexcrichton, line 137: Could this link to some documentation about where this is coming from in the order of searching for paths?

src/symbolize/gimli/elf.rs
```diff
@@ -10,11 +15,150 @@ type Elf = object::elf::FileHeader32<NativeEndian>;
...
+    let filename = Path::new(OsStr::from_bytes(filename));
+
+    // Try "/parent/filename" if it differs from "path"
+    let f = parent.join(filename);
+    if f != path && f.is_file() {
+        return Some(f);
+    }
+
+    // Try "/parent/.debug/filename"
+    let f = parent.join(".debug").join(filename);
```
alexcrichton, line 149: To keep allocations down in this function (ideally) could this use `.push` and `.pop` on `PathBuf`?

src/symbolize/gimli/elf.rs
```diff
@@ -10,11 +15,150 @@ type Elf = object::elf::FileHeader32<NativeEndian>;
...
+    if f.is_file() {
+        return Some(f);
+    }
+
+    // Try "/usr/lib/debug/parent/filename"
+    let parent = parent.strip_prefix("/").unwrap();
+    let f = Path::new("/usr/lib/debug").join(parent).join(filename);
```
alexcrichton, line 156: Similar to above, if this is linux-specific should this be wrapped in a `if cfg!`?

Or perhaps something like a global `AtomicBool` which indicates the presence of `/usr/lib/debug` should prefix this so we don't keep trying on systems that don't have that directory?

alexcrichton (commented):
Awesome, thanks for this! This is definitely tricky enough though that I'd want some tests in CI for this. If this is something that varies across unix distributions it'd be great if a variety of containers could run as well and we'd run tests in those containers.

To figure that out, though, can you describe to me how this is expected to work? Does separate debuginfo always require the intervention of a separate tool? It looks like with `objcopy` you're deleting the debuginfo from one file, adding it to another, and then adding a link of one form or another from the original file to the new file? I'm not entirely sure what `dwz` is doing...

This may not be a scenario where tests can run via `cargo test`, but it might be a situation where there needs to be an auxiliary crate where `cargo run` runs all the tests so you can `cargo build`, frob the debuginfo, and then run the binary manually to run all the tests.

Finally, it looks like this is handling `/usr/lib` things, which is presumably for system-installed debuginfo? Could the tests link to some library which has debuginfo inserted there and assert that the filenames and line numbers appear ok?

src/symbolize/gimli/elf.rs
```diff
@@ -10,11 +15,150 @@ type Elf = object::elf::FileHeader32<NativeEndian>;
...
+                path.push(hex(byte >> 4));
+                path.push(hex(byte & 0xf));
+            }
+            path.extend(BUILD_ID_SUFFIX);
+            if let Some(mapping) = Mapping::new_debug(Path::new(OsStr::from_bytes(&path)), None) {
+                return Some(mapping);
+            }
+        }
+
+        // Try to locate an external debug file using the GNU debug link section.
```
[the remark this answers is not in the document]
philipc: That's the best documentation I know of for this. I'll include it.

src/symbolize/gimli/elf.rs
philipc: I don't know of any documentation for the format of this. The only documentation I found is the dwz man page, but it doesn't give this detail. I based the code on what libbacktrace does, but gdb would have been a better source. And indeed, gdb does the path resolution differently so I'll have to fix that.

It is handled here because this is a section in the external debug file, not the original object.

src/symbolize/gimli/elf.rs
philipc: gdb uses a `debug-file-directory` setting which defaults to `"${libdir}/debug`. This whole external debug scheme is distribution specific, but I think most linux distributions use it, and I think FreeBSD does too. For those that do, I don't know if any deviate from `/usr/lib/debug`. So we probably need a `if cfg!` around the whole build-id/debuglink support. I think having an `AtomicBool` for the presence of `/usr/lib/debug` would be worth it too, since I assume someone might add build-id notes to binaries for purposes other than locating debug info.

philipc:
> Awesome, thanks for this! This is definitely tricky enough though that I'd want some tests in CI for this. If this is something that varies across unix distributions it'd be great if a variety of containers could run as well and we'd run tests in those containers.

Sure, I can work on doing that.

> To figure that out, though, can you describe to me how this is expected to work? Does separate debuginfo always require the intervention of a separate tool? It looks like with `objcopy` you're deleting the debuginfo from one file, adding it to another, and then adding a link of one form or another from the original file to the new file?

That's pretty much it. https://sourceware.org/gdb/onlinedocs/gdb/Separate-Debug-Files.html describes this fairly well, but in summary:
- the linker adds a build ID note
- objcopy extracts the debuginfo to a debug file, deletes it from the original file and inserts a `.gnu_debuglink` section with the name of the debug file
- during the build, this debug file may located in either the same directory, or under a `.debug` directory (I do one of each of these for the commands I tested with)
- the packager installs the debug file in a tree under `/usr/lib/debug`, and adds a symbol link to that from `/usr/lib/debug/.build-id`

> I'm not entirely sure what `dwz` is doing...

`dwz` by itself is a DWARF compressor; it tries to rewrite the DWARF in an equivalent form that is smaller. It is run on debug files that have already been extracted from the original files. With the `-m` option it can also identify common DWARF entries in multiple debug files, and move these to a shared file called the supplementary object file. The DWARF entries in the debug files are modified to refer to the supplementary object file using special DWARF forms (`DW_FORM_ref_sup4/8` and `DW_FORM_strp_sup`). `dwz` modifies the debug files to contain  `.gnu_debugaltlink` sections with the name of the supplementary object file.

> This may not be a scenario where tests can run via `cargo test`, but it might be a situation where there needs to be an auxiliary crate where `cargo run` runs all the tests so you can `cargo build`, frob the debuginfo, and then run the binary manually to run all the tests.

Sounds doable.

> Finally, it looks like this is handling `/usr/lib` things, which is presumably for system-installed debuginfo? Could the tests link to some library which has debuginfo inserted there and assert that the filenames and line numbers appear ok?

Yeah. So firstly we'll need to ensure the container has system-installed debuginfo. And then we'll need to get a stack trace that contains that debuginfo. This is what I was asking cuviper about in #410, but I gave up and just used `__libc_start_main` instead.

philipc:
Perhaps we could just manually install the auxiliary crate debug info into the system dir. Is that something we can do for the containers in CI, and would that be a good enough test?

src/symbolize/gimli/elf.rs
bjorn3: What about MinGW? Doesn't it support build-id/debuglink too?

philipc: I don't know. Do you have a link to any information about that?

alexcrichton:
Ok that all sounds good! I'm happy to help out with CI too if desired.

philipc:
This is ready for further review. I've added a single docker test as part of the main Ubuntu tests. It doesn't really need docker except that it lets us write to `/usr/lib/debug`.

alexcrichton (approved):
This looks perfect, thanks so much!

I'm gonna fiddle locally a bit afterwards to see if I can fixup some of the transmutes to `'static`, I'd ideally like to contain that if possible. In any case though this all looks great to me, thanks again!

PR #428, opened 2021-06-10
Import some items from winapi as they are now present

JohnTitor:
https://github.com/retep998/winapi-rs/issues/768 has been fixed now, and `SymGetOptions`/`SymSetOptions` were added by retep998/winapi-rs#864. This also updates the patch version so that someone doesn't pull an older version mistakenly.

### DIFF
--- a/src/dbghelp.rs
+++ b/src/dbghelp.rs
@@ -158,7 +158,7 @@ const SYMOPT_DEFERRED_LOADS: DWORD = 0x00000004;
 dbghelp! {
     extern "system" {
         fn SymGetOptions() -> DWORD;
-        fn SymSetOptions(options: DWORD) -> ();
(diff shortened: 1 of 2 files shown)


src/dbghelp.rs
```diff
@@ -176,7 +158,7 @@ const SYMOPT_DEFERRED_LOADS: DWORD = 0x00000004;
 dbghelp! {
     extern "system" {
         fn SymGetOptions() -> DWORD;
-        fn SymSetOptions(options: DWORD) -> ();
+        fn SymSetOptions(options: DWORD) -> DWORD;
```
JohnTitor, line 161: This seems wrong: https://docs.microsoft.com/en-us/windows/win32/api/dbghelp/nf-dbghelp-symsetoptions

alexcrichton:
Thanks!

PR #446, opened 2021-10-14
Don't require a UUID in Mach-O binaries

### LINKED ISSUE #443: frames cannot be resolved on macOS
### code
```rust
fn main() {
    backtrace::trace(|frame| {
        println!("frame: {:?}", frame);

        // Resolve this instruction pointer to a symbol name
        backtrace::resolve_frame(frame, |symbol| {
            println!("\tname: {:?}", symbol.name());
            println!("\tfilename: {:?}", symbol.filename());
        });

        true // keep going to the next frame
    });
}
```

### output
```
frame: Frame { ip: 0x1046c2048, symbol_address: 0x1046c2048 }
frame: Frame { ip: 0x1046c2098, symbol_address: 0x1046c2098 }
frame: Frame { ip: 0x1046c1e18, symbol_address: 0x1046c1e18 }
frame: Frame { ip: 0x1046c1a40, symbol_address: 0x1046c1a40 }
frame: Frame { ip: 0x1046c1d98, symbol_address: 0x1046c1d98 }
frame: Frame { ip: 0x1046c1c14, symbol_address: 0x1046c1c14 }
frame: Frame { ip: 0x104780c14, symbol_address: 0x104780c14 }
frame: Frame { ip: 0x1046c1bdc, symbol_address: 0x1046c1bdc }
frame: Frame { ip: 0x1046c1e40, symbol_address: 0x1046c1e40 }
```

### rustc version info
```
rustc 1.55.0
binary: rustc
commit-hash: unknown
commit-date: unknown
host: aarch64-apple-darwin
release: 1.55.0
LLVM version: 12.0.1
```

`backtrace` version: 0.3.61

I can also reproduce this with `std::backtrace` on the nightly from `2021-10-13`, but I think this is the most fitting place to file this issue (let me know if it isn't, and I'll open an issue in the Rust repo).

winterqt:
Some Mach-O binaries do not contain the `LC_UUID` command, such as those [built with Nix](https://github.com/NixOS/nixpkgs/blob/3708f33517029d8653d3d03f9619f219679d5027/pkgs/build-support/bintools-wrapper/default.nix#L331), so this PR removes the requirement for one to exist.

This fixes #443.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -91,6 +91,11 @@ jobs:
       env:
         CARGO_PROFILE_DEV_SPLIT_DEBUGINFO: packed
         CARGO_PROFILE_TEST_SPLIT_DEBUGINFO: packed
+    # Test that, on macOS, binaries with no UUID work
+    - run: cargo clean && cargo test
+      if: matrix.os == 'macos-latest'
+      env:
+        RUSTFLAGS: "-C link-arg=-Wl,-no_uuid"
 
     # Test that, on Linux, packed/unpacked debuginfo both work
     - run: cargo clean && cargo test
--- a/src/symbolize/gimli/macho.rs
+++ b/src/symbolize/gimli/macho.rs
@@ -22,7 +22,7 @@ impl Mapping {
         let map = super::mmap(path)?;
         let (macho, data) = find_header(&map)?;
         let endian = macho.endian().ok()?;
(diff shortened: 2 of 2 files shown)


alexcrichton:
Nice find, thanks! Would you be up for adding a test for this as well? (it'd be a macos-specific test that would probably run the test suite with the `-no_uuid` linker flag)

PR #454, opened 2021-12-14
make users able to set backtrace limit

name1e5s:
Recently I found a bug on android when I was attemping to capture stack backtrace in a rust function called by Android app through JNI. The `_Unwind_Backtrace` called our callback to push the `Frame`s into a `Vec` with the same address(`art_quick_generic_jni_trampoline + 42`) over and over again, and finally makes my device out of memory and have to kill the Android App to save the whole system. I think it's better to provide a api to make developers able to set the backtrace limit by themself, even when the backtrace is captured by third-party `crate`s.

Anyone can reproduce the problem using the following demo using an arm/aarch64 device with android 7.0 ~ android 12: 
https://github.com/name1e5s/backtrace_oom_demo

### DIFF
--- a/src/backtrace/mod.rs
+++ b/src/backtrace/mod.rs
@@ -1,5 +1,13 @@
 use core::ffi::c_void;
 use core::fmt;
+use core::sync::atomic::AtomicUsize;
+use core::sync::atomic::Ordering::SeqCst;
+
+#[cfg(all(target_os = "android", target_arch = "arm"))]
+const DEFAULT_TRACE_DEPTH_LIMIT: usize = usize::MAX;
+#[cfg(not(all(target_os = "android", target_arch = "arm")))]
+const DEFAULT_TRACE_DEPTH_LIMIT: usize = 65535;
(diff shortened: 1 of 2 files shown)


src/backtrace/mod.rs
```diff
@@ -1,5 +1,13 @@
 use core::ffi::c_void;
 use core::fmt;
+use core::sync::atomic::AtomicUsize;
+use core::sync::atomic::Ordering::SeqCst;
+
+#[cfg(all(target_os = "android", target_arch = "arm"))]
+const DEFAULT_TRACE_DEPTH_LIMIT: usize = usize::MAX;
+#[cfg(not(all(target_os = "android", target_arch = "arm")))]
+const DEFAULT_TRACE_DEPTH_LIMIT: usize = 65535;
```
bjorn3, line 9: Did you mean to limit android to 64k frames?

bjorn3: Then the `not()` is on the wrong item, isn't it?

alexcrichton:
I think it might be good to fix the bug on android where an infinite backtrace is returned, but I don't think that this particular solution belongs in this crate. The callback already returns a `bool` to stop the traceback and that enables users to already limit their backtrace, so I don't think there's any need to introduce secret mutable state for this.

Issue #452, opened 2021-11-12
Implement Eq/PartialEq, Hash and Clone for Backtrace?


hrydgard:
When building things like built-in memory profilers and similar, it's quite useful to be able to deduplicate captured callstacks without having to resolve them - since resolving stack traces can be kinda expensive, you ideally shouldn't do that until it's time to display things or produce the final report.

To do this, one convenient way would be to put the captured Backtrace-s in some kind of hash map pointing at counters for example, but this isn't currently possible since Backtrace doesn't implement the traits from this issue title (Eq/PartialEq, Hash and Clone).

Would a PR be accepted that implemented these for the structures in dbghelp.rs? I guess it really needs implementation in libunwind.rs and miri.rs though?

alexcrichton:
I think this could make sense for the `Backtrace` type but probably not for the raw underlying types since those are typically OS-specific pointers which aren't otherwise persisted.

workingjubilee:
This request was reflected on and declined for reasons that are probably not going to change: users of this library can newtype things as they like and this change would be more opinionated than this crate wants to be. So, closing this issue.

PR #457, opened 2021-12-16
Implement Hash, PartialEq and Eq for Backtrace

emilk:
Closes https://github.com/rust-lang/backtrace-rs/issues/452

The `Eq` and `Hash` is based only on the instruction pointers in the frames,
ignoring everything else. In particular, resolving the symbols does not change the identity of the backtrace!

### DIFF
--- a/src/capture.rs
+++ b/src/capture.rs
@@ -20,7 +20,7 @@ use serde::{Deserialize, Serialize};
 ///
 /// This function requires the `std` feature of the `backtrace` crate to be
 /// enabled, and the `std` feature is enabled by default.
-#[derive(Clone)]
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR, but thinking about this again I'm less certain that this belongs in this library. This is a somewhat heavyweight object to implement equality/hashing for. There's arguably a lot that could go into such an equality test or hash but picking purely just the `ip` is opinionated and not necessarily correct for all consumers of this library. I think it might actually be best for eternal consumers, if so desired, to manually implement this check with a wrapper type around `Backtrace` to explicitly opt-in to hashing/equating only the `ip` field.

emilk:
Good point!

PR #469, opened 2022-05-03
Bump object to 0.28

bnjbvr:
This bumps the object dependency to 0.28, so as to reduce the number of duplicated dependencies in upstream projects. Tests all passed locally.


alexcrichton:
👍

PR #473, opened 2022-07-10
Added tracing of threads other than the calling thread on Windows

aminya:
This is a continuation of the work by @Jardynq that makes it possible to trace threads other than the current one. This can be used to profile Windows applications.

https://github.com/tikv/pprof-rs/issues/9#issuecomment-1114222675

https://github.com/tikv/pprof-rs/pull/122


aminya:
The MIR test is irrelevant to this pull request, and it is also happening on the master branch.

src/backtrace/dbghelp.rs
```diff
@@ -90,12 +90,38 @@ struct MyContext(CONTEXT);
...
+        // There is most definitely more pitfalls i haven't thought
+        // of or encountered. This is windows after all.
+
+        context.0.ContextFlags = CONTEXT_ALL; // TODO: Narrow down required flags
+        if SuspendThread(thread) as i32 == -1 {
+            ResumeThread(thread);
+            return;
+        }
+        let status = GetThreadContext(thread, &mut context.0);
+        if ResumeThread(thread) as i32 == -1 || status == 0 {
```
bjorn3, line 121: What if the other thread was already suspended? I think it shouldn't resume it in that case.

aminya: I think `SuspdendThread` in the above lines suspends the thread. We have to resume it here after we have extracted to context.

aminya: Also see 
https://github.com/tikv/pprof-rs/issues/9#issuecomment-1114222675

bjorn3: What if someone called `SuspendThread` and then tried to get a backtrace? The current code would resume the thread again, while it did be reasonable to expect that the thread stays suspended if you suspended it yourself.

aminya: `ResumeThread` does not resume the thread, it increments the suspend count.
In this code, `SuspendThread` increments the suspend count in line 116, and `ResumeThread` decrements it in line 121 or 117. Each code path has a pair of suspend/resume.   

See this for more information
https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-resumethread

alexcrichton (commented):
Personally I'm wary of adding platform-specific functionality to this crate so this may best live elsewhere. The `dbghelp.rs` file alone isn't the most complicated to copy for something like that.

Otherwise this is not a safe implementation because the thread is resumed after the register context is captured. That means the stack walking may not work because the stack could change while the stack was being walked.

src/backtrace/dbghelp.rs
bjorn3: I didn't know that.

aminya:
> I'm wary of adding platform-specific functionality to this crate so this may best live elsewhere. The `dbghelp.rs` file alone isn't the most complicated to copy for something like that.
> 

This only adds Windows for now because it is Rust cannot be profiled on Windows, and we need this to remove a **platform-specific restriction**. In the future, If other operating systems have such functionality, people can suggest implementations. 

I think it is more important to allow people to profile Rust on Windows. This is way more limiting than not having the implementation of `trace_thread` for other operating systems. 

> Otherwise this is not a safe implementation because the thread is resumed after the register context is captured. That means the stack walking may not work because the stack could change while the stack was being walked.

Based on the official Windows docs, the only possible way to get the thread's context is to suspend it. 
https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getthreadcontext

bjorn3:
> This only adds Windows for now because it is Rust cannot be profiled on Windows

backtrace-rs is not magic. You can use the same api's as for C/C++ stack walking to walk the stack for Rust. This means that you can copy `dbghelp.rs` from backtrace-rs into a separate crate and implement walking the stack of a different thread in this copy. No need to change backtrace-rs.

> and we need this to remove a platform-specific restriction.

I fail to see how it is a platform-specific restriction as backtrace-rs has no support for stack walking of a different thread on any platform at all.

> Based on the official Windows docs, the only possible way to get a valid context for a running thread is to suspend it.
https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getthreadcontext#remarks

I think @alexcrichton means that you need to keep the running thread suspended while walking the stack to prevent corruption of the resulting backtrace.

src/backtrace/dbghelp.rs
```diff
@@ -90,12 +90,36 @@ struct MyContext(CONTEXT);
...
-    RtlCaptureContext(&mut context.0);
+    if thread == GetCurrentThread() || thread.is_null() {
+        RtlCaptureContext(&mut context.0);
+    } else {
+        // The suspending and resuming of threads is very risky.
+        // It can end in a deadlock if the current thread tries to access
+        // an object thats been locked by the suspended thread.
+        // That's why we only do as little work as possible while
+        // the thread is suspended, and resume it quickly after.
+        // There might be more pitfalls we haven't thought of or encountered
```
workingjubilee, line 111: No speculative "maybe this is bad?" in fresh code, please. I'd like to hear why this is actually okay, despite concerns about corrupting the backtrace.

src/backtrace/dbghelp.rs
```diff
@@ -90,12 +90,36 @@ struct MyContext(CONTEXT);
...
+
+        context.0.ContextFlags = CONTEXT_ALL; // TODO: Narrow down required flags
```
workingjubilee, line 113: This TODO would have to be addressed.

src/windows.rs
```diff
@@ -630,6 +655,16 @@ ffi! {
         pub Low: u64,
         pub High: i64,
     }
+
+    pub const CONTEXT_AMD64: DWORD = 0x00100000;
+    pub const CONTEXT_CONTROL: DWORD = CONTEXT_AMD64 | 0x00000001;
+    pub const CONTEXT_INTEGER: DWORD = CONTEXT_AMD64 | 0x00000002;
+    pub const CONTEXT_SEGMENTS: DWORD = CONTEXT_AMD64 | 0x00000004;
+    pub const CONTEXT_FLOATING_POINT: DWORD = CONTEXT_AMD64 | 0x00000008;
+    pub const CONTEXT_DEBUG_REGISTERS: DWORD = CONTEXT_AMD64 | 0x00000010;
```
workingjubilee, line 664: Can't we avoid repeating the same 5 expressions for 3 different architectures?

src/backtrace/mod.rs
```diff
@@ -66,6 +66,33 @@ pub unsafe fn trace_unsynchronized<F: FnMut(&Frame) -> bool>(mut cb: F) {
...
+///
+/// It gets the traced thread's handle as its first argument.
+///
+/// # Safety
+///
+/// This function is intended for profiling and debugging.
+///
+/// If the given thread handle is not the current thread,
+/// this function suspends the thread for a short amount of time to be able to extract the thread context.
+/// This might end in a deadlock if the current thread tries to access an object thats been locked by the suspended thread.
```
workingjubilee, line 79: Are these statements about the briefness and the concern about deadlocking related to each other?

workingjubilee (commented):
This crate is now primarily under the maintenance of the [crate-maintainers team](https://github.com/rust-lang/team/blob/master/teams/crate-maintainers.toml). I personally view maintaining this crate primarily as part and parcel of the responsibility of improving the standard library that I have been entrusted with. This means I evaluate PRs to it with the needs of std in mind.

Frankly, we are not exactly well-furnished with plenty of resources for maintaining this, either. With recruited assistance, we just recently managed to scrape together a bunch of maintenance on this crate to keep it up to date with the needs of recent Rusts, after it had fallen way behind. It helps this crate is, in many ways, pretty simple, despite the dark arts it performs.

I think we have reason to be quite proud about that.

Unfortunately, that's still pride about barely hanging on. And this isn't the only pull request of the nature of "please make backtrace-rs more useful for profilers at the cost of additional complexity". So let me entirely set aside this discussion about "platform-specific functionality" versus "platform-specific restrictions" and be simple about it: as far as I am concerned, either a PR to this is making std's backtrace work for a platform or it doesn't get to make maintaining this crate harder.

If you can work out a way to add all the code you folks want to add to help this crate specifically support profiling better, while keeping it out of the way of the codepaths that std needs, that seems plausible to me. It wouldn't hurt to have more people get involved in opening PRs to this crate with its maintenance needs in mind. But I do mean _out of the way_... it's not necessarily about whether std lights up such codepaths, but that I need to be able to gloss over that code during review of PRs to backtrace-rs for supporting std, rather than having to also account for code that has possibly been microoptimized or given a strange API for a profiler's performance or behavior requirements... and I will still hold such PRs to a high correctness standard because even trying to keep those paths isolated, they will wind up interacting anyways.

workingjubilee:
This has fallen behind the current work on dbghelp, so I am closing it. Apologies.

PR #486, opened 2022-10-08
Apply clippy::uninlined_format_args fixes

nyurik:
This is the result of running `clippy::uninlined_format_args` lint. Currently the lint is in `pedantic`, but there are plans/hopes to move it to `style`.

How can I add this specific lint to the current build/CI tools, or is this not possible?  It would be ideal not to make the lint as the default style and then end up with hundreds of changes in an unrelated PR due to CI requirements.

Locally, I ran this to generate the changes.  Not ideal because it has produced a number of compilation errors.

```bash
rustup run nightly cargo clippy --fix --no-deps --allow-dirty -- -A clippy::all -W clippy::uninlined_format_args
```

Note that the MSRV must be at least 1.58 (which introduced inline format args)

### DIFF
--- a/examples/raw.rs
+++ b/examples/raw.rs
@@ -21,7 +21,7 @@ fn print() {
     let mut cnt = 0;
     backtrace::trace(|frame| {
         let ip = frame.ip();
-        print!("frame #{:<2} - {:#02$x}", cnt, ip as usize, HEX_WIDTH);
+        print!("frame #{cnt:<2} - {:#0HEX_WIDTH$x}", ip as usize);
         cnt += 1;
 
         let mut resolved = false;
@@ -33,13 +33,13 @@ fn print() {
             }
 
             if let Some(name) = symbol.name() {
-                print!(" - {}", name);
+                print!(" - {name}");
             } else {
                 print!(" - <unknown>");
             }
             if let Some(file) = symbol.filename() {
                 if let Some(l) = symbol.lineno() {
-                    print!("\n{:13}{:4$}@ {}:{}", "", "", file.display(), l, HEX_WIDTH);
+                    print!("\n{:13}{:HEX_WIDTH$}@ {}:{l}", "", "", file.display());
                 }
             }
             println!("");
--- a/examples/raw.rs
+++ b/examples/raw.rs
@@ -21,7 +21,7 @@ fn print() {
     let mut cnt = 0;
     backtrace::trace(|frame| {
         let ip = frame.ip();
-        print!("frame #{cnt:<2} - {:#0HEX_WIDTH$x}", ip as usize);
+        print!("frame #{:<2} - {:#02$x}", cnt, ip as usize, HEX_WIDTH);
         cnt += 1;
 
         let mut resolved = false;
@@ -39,7 +39,7 @@ fn print() {
             }
             if let Some(file) = symbol.filename() {
                 if let Some(l) = symbol.lineno() {
-                    print!("\n{:13}{:HEX_WIDTH$}@ {}:{l}", "", "", file.display());
+                    print!("\n{:13}{:4$}@ {}:{}", "", "", file.display(), l, HEX_WIDTH);
                 }
             }
             println!("");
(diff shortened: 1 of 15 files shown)


BurntSushi:
> Currently the lint is in `pedantic`, but there are plans/hopes to move it to `style`.

Where can I participate in this discussion of plans?

examples/raw.rs
```diff
@@ -21,7 +21,7 @@ fn print() {
     let mut cnt = 0;
     backtrace::trace(|frame| {
         let ip = frame.ip();
-        print!("frame #{:<2} - {:#02$x}", cnt, ip as usize, HEX_WIDTH);
+        print!("frame #{cnt:<2} - {:#0HEX_WIDTH$x}", ip as usize);
```
workingjubilee, line 24: This introduces active confusion into reading this print statement because these are now "out of order", i.e. I am no longer reading all the format arguments strictly "left to right".

examples/raw.rs
```diff
@@ -33,13 +33,13 @@ fn print() {
...
             if let Some(name) = symbol.name() {
-                print!(" - {}", name);
+                print!(" - {name}");
             } else {
                 print!(" - <unknown>");
             }
             if let Some(file) = symbol.filename() {
                 if let Some(l) = symbol.lineno() {
-                    print!("\n{:13}{:4$}@ {}:{}", "", "", file.display(), l, HEX_WIDTH);
+                    print!("\n{:13}{:HEX_WIDTH$}@ {}:{l}", "", "", file.display());
```
workingjubilee, line 42: Same issue re: added confusion, ordering, and left-to-right reading.

src/print.rs
```diff
@@ -288,7 +288,7 @@ impl BacktraceFrameFmt<'_, '_, '_> {
         // We only care about the first symbol of a frame
         if self.symbol_index == 0 {
             self.fmt.fmt.write_str("{{{bt:")?;
-            write!(self.fmt.fmt, "{}:{:?}", self.fmt.frame_index, frame_ip)?;
+            write!(self.fmt.fmt, "{}:{frame_ip:?}", self.fmt.frame_index)?;
```
workingjubilee, line 291: Same issue re: added confusion, ordering, and left-to-right reading.

src/print.rs
```diff
@@ -266,18 +266,18 @@ impl BacktraceFrameFmt<'_, '_, '_> {
         // Filename/line are printed on lines under the symbol name, so print
         // some appropriate whitespace to sort of right-align ourselves.
         if let PrintFmt::Full = self.fmt.format {
-            write!(self.fmt.fmt, "{:1$}", "", HEX_WIDTH)?;
+            write!(self.fmt.fmt, "{:HEX_WIDTH$}", "")?;
         }
         write!(self.fmt.fmt, "             at ")?;
```
workingjubilee, line 271: If this part is going to be diffed, surely there's a better to write it, in general? The part you diffed is literally formatting the empty string in, and all this is to generate what is mostly whitespace...

workingjubilee (changes requested):
Thank you for your efforts. My apologies for it being so long to get around to this PR: this crate is fairly direly under-maintained. For that reason, currently this crate does not run clippy in CI, nor do I believe there is an intention to, as adding additional CI checks on getting code in is not what we want to be doing unless those are strictly necessary for functionality.

Normally I am inclined to approve a modest beautification PR like this, but I do not believe all of these changes are equally beautifying. Some make the code harder to read, by changing things so they are no longer read in a left-to-right manner, or in general complicating the code's overall style so that some of the results are surprising. If you wish to make additional changes to remove the diffs I have noted, or to improve the formatting code further until the concerns I have are eliminated, then I believe we could accept this. Otherwise, I will probably close this when I next make a pass over the PRs for this crate.

nyurik:
thx for great feedback, agree, will fix!

nyurik:
@workingjubilee I addressed your feedback, and added a few more similar inlinings.

To keep things cleaner, I think #539 should be merged first, possibly followed by #538

workingjubilee (approved):

nyurik:
@workingjubilee i rebased this PR to make sure it is up to date. Please review & merge unless there are some new issues. Thx!

PR #488, opened 2022-10-21
Use /proc/self/maps when available instead of std::env::current_exe

pnkfelix:
rust-lang/rust#101913 illustrates a scenario on Linux where backtraces fail because of their reliance on `std::env::current_exe()` always yielding the path to the *Rust* binary that is driving execution.

This PR changes that strategy, by first trying to parse `/proc/self/maps` and extract from it the path to the Rust binary that was mmap'ed in as part of linking/loading.

It addresses the problem for my local Linux host, as illustrated by the included regression test. I have not yet tested it on other platforms, but I am hoping I got the logic right (in terms of just falling back gracefully to `std::env::current_exe()` in cases where we either cannot find a `/proc/self/maps`, or we cannot find the instruction pointer of interest anywhere in the mapping reported by `/proc/self/maps`).

### DIFF
--- a/src/symbolize/gimli/parse_running_mmaps_unix.rs
+++ b/src/symbolize/gimli/parse_running_mmaps_unix.rs
@@ -3,9 +3,9 @@
 // general purpose, but it hasn't been tested elsewhere.
 
 use super::mystd::fs::File;
-use super::mystd::io::{BufRead, BufReader};
+use super::mystd::io::Read;
 use super::mystd::str::FromStr;
-use super::{OsString, Vec};
+use super::{OsString, String, Vec};
 
 #[derive(PartialEq, Eq, Debug)]
 pub(super) struct MapsEntry {
@@ -55,11 +55,11 @@ pub(super) struct MapsEntry {
 
 pub(super) fn parse_maps() -> Result<Vec<MapsEntry>, &'static str> {
     let mut v = Vec::new();
-    let proc_self_maps =
+    let mut proc_self_maps =
         File::open("/proc/self/maps").map_err(|_| "Couldn't open /proc/self/maps")?;
-    let proc_self_maps = BufReader::new(proc_self_maps);
-    for line in proc_self_maps.lines() {
-        let line = line.map_err(|_io_error| "Couldn't read line from /proc/self/maps")?;
+    let mut buf = String::new();
+    let _bytes_read = proc_self_maps.read_to_string(&mut buf).map_err(|_| "Couldn't read /proc/self/maps")?;
+    for line in buf.lines() {
         v.push(line.parse()?);
     }
 
--- a/src/symbolize/gimli/parse_running_mmaps_unix.rs
+++ b/src/symbolize/gimli/parse_running_mmaps_unix.rs
@@ -58,7 +58,9 @@ pub(super) fn parse_maps() -> Result<Vec<MapsEntry>, &'static str> {
     let mut proc_self_maps =
         File::open("/proc/self/maps").map_err(|_| "Couldn't open /proc/self/maps")?;
     let mut buf = String::new();
-    let _bytes_read = proc_self_maps.read_to_string(&mut buf).map_err(|_| "Couldn't read /proc/self/maps")?;
+    let _bytes_read = proc_self_maps
+        .read_to_string(&mut buf)
+        .map_err(|_| "Couldn't read /proc/self/maps")?;
     for line in buf.lines() {
         v.push(line.parse()?);
     }
--- a/src/symbolize/gimli/parse_running_mmaps_unix.rs
+++ b/src/symbolize/gimli/parse_running_mmaps_unix.rs
@@ -2,7 +2,9 @@
 // in `mod libs_dl_iterate_phdr` (e.g. linux, freebsd, ...); it may be more
 // general purpose, but it hasn't been tested elsewhere.
 
-use super::mystd::io::BufRead;
+use super::mystd::fs::File;
+use super::mystd::io::{BufRead, BufReader};
+use super::mystd::str::FromStr;
 use super::{OsString, Vec};
 
 #[derive(PartialEq, Eq, Debug)]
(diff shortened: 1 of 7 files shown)


src/symbolize/gimli/parse_running_mmaps_unix.rs
```diff
@@ -0,0 +1,146 @@
...
+    /// path has been deleted.
+    ///
+    /// Note that modifications like the latter two indicated above imply that
+    /// in general the pathname may be ambiguous. (I.e. you cannot tell if the
+    /// denoted filename actually ended with the text "(deleted)", or if that
+    /// was added by the maps rendering.
+    pathname: OsString,
+}
+
+pub(super) fn parse_maps() -> Result<Vec<MapsEntry>, &'static str> {
```
bjorn3, line 50: This logic only applies to linux I think. Netbsd has `/proc/curproc/`. macOS doesn't have `/proc` at all. As for other unixes I'm not quite sure.

pnkfelix: I'm happy to revise this so that only Linux pulls in the "interesting" `parse_maps` file, and every other target just goes to the no-op variant.

pnkfelix: I looked again, and realized that the sole calling module is solely defined here:

https://github.com/rust-lang/backtrace-rs/blob/8b83ba10fc8ccfc12ff75fcc3277286c0b9fad62/src/symbolize/gimli.rs#L175-L185


That makes this a lot easier to limit to just these small subset of targets.

pnkfelix:
hmm, `cargo test --no-default-features` failed, something with how I'm using the std library. Will investigate

alexcrichton (commented):
This seems reasonable-enough to me, thanks for implementing this! One thing that might be worth checking is the binary size impact of adding this, since I'd expect that `BufReader`  is not trivial in its code size.

If the code size of this is somewhat substantial there's probably a number of ways that this can cut down on code size, for example there's a fair number of allocations which I think could be removed and/or changed to iterators. Not necessarily trivially so I don't think it's worth doing unless motivated though.

pnkfelix:
> This seems reasonable-enough to me, thanks for implementing this! One thing that might be worth checking is the binary size impact of adding this, since I'd expect that `BufReader` is not trivial in its code size.

Hmm. I honestly did not know there was a significant cost associated with `BufReader` itself. My sole motivation for using `BufReader` was just to get easy line-by-line parsing.

So here's some data:

On master, after doing `cargo build --release`:

```
% ls -sk target/release/libbacktrace.rlib
992 target/release/libbacktrace.rlib
```

On this PR's branch, after doing `cargo build --release`:

```
% ls -sk target/release/libbacktrace.rlib
1052 target/release/libbacktrace.rlib
```

I am inferring adding 60kb to a 992kb rlib is more than you want to pay for this..

> If the code size of this is somewhat substantial there's probably a number of ways that this can cut down on code size, for example there's a fair number of allocations which I think could be removed and/or changed to iterators. Not necessarily trivially so I don't think it's worth doing unless motivated though.

I cannot tell from your feedback whether you are suggesting that I leave in the `BufReader` usage while removing allocations elsewhere, or to remove the `BufReader` *in addition to* making other changes.

In any case I'll poke at it and see if I can figure out a way to cut down on the code size increase without too much pain.

(/me does some searching on crates.io.) Let me look into using https://crates.io/crates/bstr; at least, I assume that would be an acceptable dependency to pull in here.

alexcrichton:
Thanks for measuring! 60k indeed seems pretty small and probably largely inconsequential given the size of this crate already. In that case while I think it's possible to somewhat easily cut down on code size it's not a priority, so it's fine to leave as-is, sorry if that wasn't clear!

It looks like there's some minor nits on CI but feel free to bump the msrv in the CI configuration file and re-push, there's no need to strictly maintain compatibilty with that.

pnkfelix:
> Thanks for measuring! 60k indeed seems pretty small and probably largely inconsequential given the size of this crate already. In that case while I think it's possible to somewhat easily cut down on code size it's not a priority, so it's fine to leave as-is, sorry if that wasn't clear!
> 
> It looks like there's some minor nits on CI but feel free to bump the msrv in the CI configuration file and re-push, there's no need to strictly maintain compatibilty with that.

Well, I've already jumped through some other hoops, let me take a shot at seeing if I can satisfy the existing MSRV

pnkfelix:
> Thanks for measuring! 60k indeed seems pretty small and probably largely inconsequential given the size of this crate already. In that case while I think it's possible to somewhat easily cut down on code size it's not a priority, so it's fine to leave as-is, sorry if that wasn't clear!

So ... in https://github.com/rust-lang/backtrace-rs/pull/488/commits/cc0cdbb9fd7ef325f419420df86a302cb8ed6c59 I did sidestep the use of `BufRead` by instead doing a (very naive) `Read`-based implementation: namely, read the *whole file* into one `String`, and then split that into lines.

Doing this cut out 8K from the rlib size.

I'm sort of sad about doing this, but then again, even my previous implementation was allocating a `Vec` of `String` whose size was proportional to the full contents of `/proc/self/maps`, so its at most a 2x runtime-memory overhead over what I was previously doing.

Anyway, @alexcrichton , let me know if you'd prefer I put a line-by-line parsing version back. (I suspect if anyone invests time in doing that, they should also get rid of the `Vec` and just have the code directly extract the desired entry as it traverses the lines of `/proc/self/maps`.)

tests/current-exe-mismatch.rs
```diff
@@ -0,0 +1,137 @@
...
+    assert!(found_my_name);
+
+    Ok(())
+}
+
+// we use the `readelf` command to extract the path to the interpreter requested
+// by our binary.
+//
+// if we cannot `readelf` for some reason, or if we fail to parse its output,
+// then we will just give up on this test (and not treat it as a test failure).
```
jonhoo, line 108: This feels like it might fairly easily lead to the test being skipped without anyone realizing due to some unrelated change. Are there known cases where `readelf` isn't available but the test is still run? Alternatively, could we exit with an error if this is detected and we know we're on CI?

pnkfelix: Well, the way things are structured now, this test is run unconditionally regardless of target, so there's lots of cases where `readelf` is not available and the test is still run. (E.g. its run on my Mac laptop, and ends up in the `IgnoreTest("readelf invocation failed".into())` path.)

Would you feel more comfortable if the output of the test still said "test ignored" but also included the text for *why* it was ignored, like so on Mac:

```
test result: ignored: "readelf invocation failed"
```

(I guess your message explicitly said that you want a hard error from CI. I don't know if we have that contextual information available when the test is running, I'll investigate.)

pnkfelix: Also: I recommend that we land this now and file an issue to figure out how best to revise its associated test. This patch is resolving a real (if niche) issue, and it seems suboptimal to block landing it based on the concern about some other change causing the test to start being incorrectly ignored.

alexcrichton:
Seems reasonable to me, thanks again for this @pnkfelix!

tests/current-exe-mismatch.rs
jonhoo: Yep, definitely not a blocker! I just worry about this all being broken, but the breakage not being detected because some *other* issue causes readelf to not be available on CI and so the test is always ignored.

PR #493, opened 2022-11-08
consider only elf text segments

bsekura:
Question:

On Unix platforms (e.g. Linux),  `dl_iterate_phdr` is used to enumerate elf modules and their sections. These are then iterated upon in `gimli::avma_to_svma` to find the segment and calculate the address in order to look up DWARF debug data.

Is there are reason to consider any segments other than `.text`?

### DIFF
--- a/src/symbolize/gimli/libs_dl_iterate_phdr.rs
+++ b/src/symbolize/gimli/libs_dl_iterate_phdr.rs
@@ -60,6 +60,9 @@ unsafe extern "C" fn callback(
         name,
         segments: headers
             .iter()
+            .filter(|header| {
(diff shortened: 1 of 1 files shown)


bjorn3:
One some platforms like x86 there is no way to prevent execution of data, so the PF_X flag doesn't have to be set at all. While it is generally set for code on x86, I'm sure there are cases where someone forgot to set it. For example when writing assembly.

bsekura:
Thanks for the comment. I think even if it's data segment execution shenanigans, it will unlikely have symbols anyway.

But anyway, some platforms don't like this change, so closing for now.

PR #495, opened 2022-11-24
Update dependencies

a1phyr:
Includes:
- `object`
- `addr2line`
- `cpp_demangle`


JohnTitor:
Could you include the description why these should be updated? Or, that's just because they have new releases?

a1phyr:
Just because they have new releases. I think it is a good thing to keep up with the ecosystem in general.

JohnTitor (commented):
Alright, thanks!

JohnTitor (approved):

Issue #499, opened 2022-12-01
Make MAPPINGS_CACHE_SIZE configurable


hehechen:
The [MAPPINGS_CACHE_SIZE](https://github.com/rust-lang/backtrace-rs/blob/5be2e8ba9cf6e391c5fa45219fc091b4075eb6be/src/symbolize/gimli.rs#L55) is 4 by hard coding. 
Can we make it configurable? In some scenarios, the lib mappings cache miss will cause performance degrade, especially in the scenario that the `debug_section` is compressed. For example: https://github.com/pingcap/tiflash/issues/6347#issuecomment-1333038382

Issue #365, opened 2020-07-24
still reachable bytes reported by valgrind.


nmahendru:
I am trying to use backtrace to get some error information in my application and unit tests report a lot of `still reachable` bytes when run under valgrind.

I am trying to use this on a hardware device with no probing capabilities so I am little specific of the memory usage and if something is off I try to avoid it.
cargo version:
```
cargo 1.41.0 (626f0f40e 2019-12-03)
```
valgrind version:
```
valgrind --version
valgrind-3.14.0
```
Platform:
```
Linux 9eef3ca4a968 4.19.76-linuxkit #1 SMP Tue May 26 11:42:35 UTC 2020 x86_64 GNU/Linux
```
Simple example to reproduce:
```
use backtrace::Backtrace;

fn main() {
    println!("Hello, world!");
}

fn call_bt() {
    let bt = Backtrace::new();
    println!("{:?}", bt);
}
#[cfg(test)]
mod tests {
    use super::*;
    #[test]
    fn test_nothing() {
        call_bt();
    }
}
```
Cargo.toml
```

[dependencies]
backtrace = "0.3.50"
```
valgrind command:
```
valgrind --leak-check=full target/debug/deps/rust_try-116747691a975264
```
valgrind report:
```
==281== Memcheck, a memory error detector
==281== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==281== Using Valgrind-3.14.0 and LibVEX; rerun with -h for copyright info
==281== Command: target/debug/deps/rust_try-116747691a975264
==281==
==281== Syscall param statx(file_name) points to unaddressable byte(s)
==281==    at 0x493CF59: syscall (syscall.S:38)
==281==    by 0x2574BB: statx (weak.rs:90)
==281==    by 0x2574BB: std::sys::unix::fs::try_statx (fs.rs:132)
==281==    by 0x25238E: std::sys::unix::fs::stat (fs.rs:995)
==281==    by 0x159CB4: metadata<&std::path::PathBuf> (fs.rs:1591)
==281==    by 0x159CB4: term::terminfo::searcher::get_dbpath_for_term (searcher.rs:50)
==281==    by 0x158FB2: term::terminfo::TermInfo::from_name (mod.rs:87)
==281==    by 0x158DC0: term::terminfo::TermInfo::from_env (mod.rs:73)
==281==    by 0x163A09: new<std::io::stdio::Stdout> (mod.rs:232)
[...]

PR #503, opened 2022-12-09
Add API to support mapping cache capacity configurable

hehechen:
Signed-off-by: hehechen <awd123456sss@gmail.com>
closes: https://github.com/rust-lang/backtrace-rs/issues/499

In some scenarios, users may want to increase [MAPPINGS_CACHE_SIZE](https://github.com/rust-lang/backtrace-rs/blob/5be2e8ba9cf6e391c5fa45219fc091b4075eb6be/src/symbolize/gimli.rs#L55) to trade memory for performance improvement. 
This PR add some functions:
`resolve_frame_customized_cache`
`resolve_frame_unsynchronized_customized_cache`
`resolve_customized_cache`
`resolve_unsynchronized_customized_cache`
So that users can modify cache capacity to improve performance.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -113,8 +113,12 @@ extern crate alloc;
 pub use self::backtrace::{trace_unsynchronized, Frame};
 mod backtrace;
 
-pub use self::symbolize::{resolve_frame_unsynchronized, resolve_frame_unsynchronized_customized_cache};
-pub use self::symbolize::{resolve_unsynchronized, resolve_unsynchronized_customized_cache, Symbol, SymbolName};
+pub use self::symbolize::{
+    resolve_frame_unsynchronized, resolve_frame_unsynchronized_customized_cache,
+};
+pub use self::symbolize::{
+    resolve_unsynchronized, resolve_unsynchronized_customized_cache, Symbol, SymbolName,
+};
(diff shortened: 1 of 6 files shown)


hehechen:
Same as https://github.com/rust-lang/backtrace-rs/pull/500, but change the source branch.

workingjubilee:
This complicates things a... lot, in the public API, even, for this performance improvement. Can you elaborate about these "some scenarios", and what is your use case that you want specifically the ability to set a number? In the use-case of std, "backtrace" is usually "one of the last things before the world ends", so yes, we might want to cache more than 4, but I kinda think that we would want to try to lazily cache _as many mappings as can fit into memory_, because we're going to throw everything away soon anyways? But it sounds a little like you have a different need? Am I mistaken?

workingjubilee:
Hmm, judging by this issue, some users will be invoking backtrace... a lot:
https://github.com/rust-lang/backtrace-rs/issues/365

Interesting. That helps motivate this a little, but I'm still curious about what your use pattern, and thus what you want us to support, is like.

workingjubilee:
Since this seems to have already been a bit stale by the time I got to this crate, I'm going to close this for now. Feel free to reopen when you're ready to discuss it.

PR #506, opened 2022-12-16
Add symbolize support for qnx-nto

Veykril:
This enables symbol resolution on the QNX nto operating system.

These `cfg`s require the following symbols/definitions from `libc` which exist for the target:
- [`mmap`](https://www.qnx.com/developers/docs/7.1/index.html#com.qnx.doc.neutrino.lib_ref/topic/m/mmap.html)
- [`munmap`](https://www.qnx.com/developers/docs/7.1/index.html#com.qnx.doc.neutrino.lib_ref/topic/m/munmap.html)
	- `MAP_FAILED`
	- `PROT_READ`
	- `MAP_PRIVATE`
- [`dl_iterate_phdr`](
https://www.qnx.com/developers/docs/7.1/index.html#com.qnx.doc.neutrino.lib_ref/topic/d/dl_iterate_phdr.html)
	- `dl_phdr_info`
- `size_t`
- `c_void`
- `c_int`

We are currently doing a port of this for ferrocene and since backtrace-rs is a submodule in `rust-lang/rust`, having this addition upstreamed would save us from manually maintaining a fork for these two changes.

PR #507, opened 2022-12-22
Add support for QNX Neutrino

gh-tr:
Add support for QNX Neutrino.
  -- Tested on a Neutrino target for both aarch64 and x86_64 running 7.1.

Change Description:
- Splits the existing functionality (`default`) into one module and the `nto` into another
- On QNX, there is a header in the pmap file so added functionality to skip the first line of the file `/proc/self/pmap`
- Bumped libc version to the first version containing a Neutrino target.


flba-eb:
@Veykril: Could you please take a look at this PR? Would it make sense to close #506 in favor of this PR?

src/symbolize/gimli/parse_running_mmaps_unix.rs
```diff
@@ -8,7 +8,7 @@ use super::mystd::str::FromStr;
 use super::{OsString, String, Vec};
 
 #[derive(PartialEq, Eq, Debug)]
-pub(super) struct MapsEntry {
+pub struct MapsEntry {
```
workingjubilee, line 11: Do we need this to be able to be visible outside this crate? Using `pub(crate)` is preferable if `pub(super)` needs to be weakened.

src/symbolize/gimli/parse_running_mmaps_unix.rs
```diff
@@ -53,19 +53,34 @@ pub(super) struct MapsEntry {
...
-        .map_err(|_| "Couldn't read /proc/self/maps")?;
-    for line in buf.lines() {
-        v.push(line.parse()?);
-    }
+        .map_err(|_| concat_str("Couldn't read file: ", file))?;
+    Ok(buf)
+}
 
-    Ok(v)
+pub fn parse_maps() -> Result<Vec<MapsEntry>, String> {
```
workingjubilee, line 72: Same remarks about visibility discipline.

src/symbolize/gimli/parse_running_mmaps_unix.rs
flba-eb: @workingjubilee : Should be fixed now. The original `pub(super)` seems to be working after some refactoring. Same for the other finding below.

src/symbolize/gimli/parse_running_mmaps_unix.rs
```diff
@@ -53,19 +53,34 @@ pub(super) struct MapsEntry {
...
     let mut buf = String::new();
     let _bytes_read = proc_self_maps
         .read_to_string(&mut buf)
-        .map_err(|_| "Couldn't read /proc/self/maps")?;
-    for line in buf.lines() {
-        v.push(line.parse()?);
-    }
+        .map_err(|_| concat_str("Couldn't read file: ", file))?;
+    Ok(buf)
+}
```
workingjubilee, line 70: This machinery is a reimplementation of [`fs::read_to_string`](https://doc.rust-lang.org/std/fs/fn.read_to_string.html). You can import `fs` by changing
```diff
- use super::mystd::fs::File;
+ use super::mystd::fs::{self, File};
```

If you must, you can use `map_err` for the Err variant.

src/symbolize/gimli/parse_running_mmaps_unix.rs
```diff
@@ -78,6 +93,12 @@ impl MapsEntry {
     }
 }
 
+#[cfg(not(target_os = "nto"))]
+pub fn config() -> (&'static str, usize) {
+    ("/proc/self/maps", 0)
+}
```
workingjubilee, line 99: This moves to runtime what can be known at compile-time. Please describe a `static` and a `const` instead, with appropriate names. If you wish, you can put them in a `mod config {}`.

src/symbolize/gimli/parse_running_mmaps_unix.rs
```diff
@@ -53,19 +53,34 @@ pub(super) struct MapsEntry {
...
 
-    Ok(v)
+pub(super) fn parse_maps() -> Result<Vec<MapsEntry>, String> {
+    let (file, skip) = config();
+    let content = read_file(file)?;
+    parse_maps_lines(&content, skip)
+}
+
+fn parse_maps_lines(content: &str, skip: usize) -> Result<Vec<MapsEntry>, String> {
```
workingjubilee, line 78: I believe this is still functionally identical if we simply delete these lines?
```suggestion
```

src/symbolize/gimli/parse_running_mmaps_unix.rs
```diff
@@ -162,7 +243,7 @@ fn check_maps_entry_parsing_64bit() {
 
     assert_eq!(
         "7f5985f46000-7f5985f48000 rw-p 00039000 103:06 76021795                  \
-                /usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2"
+        /usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2"
```
workingjubilee, line 246: Is there a reason for diffing this line?

src/symbolize/gimli/parse_running_mmaps_unix.rs
```diff
@@ -141,13 +162,73 @@ impl FromStr for MapsEntry {
...
+
 // Make sure we can parse 64-bit sample output if we're on a 64-bit target.
-#[cfg(target_pointer_width = "64")]
+#[cfg(all(target_pointer_width = "64", not(target_os = "nto")))]
 #[test]
 fn check_maps_entry_parsing_64bit() {
     assert_eq!(
         "ffffffffff600000-ffffffffff601000 --xp 00000000 00:00 0                  \
-                [vsyscall]"
+        [vsyscall]"
```
workingjubilee, line 231: Is there a reason for diffing this line?

src/symbolize/gimli/parse_running_mmaps_unix.rs
```diff
@@ -141,13 +162,73 @@ impl FromStr for MapsEntry {
...
+        let _flags_str = parts.next().ok_or("Couldn't find flags")?;
+        let prot_str = parts.next().ok_or("Couldn't find protection")?;
+        let _maxprot_str = parts.next().ok_or("Couldn't find maximum protection")?;
+        let dev_str = parts.next().ok_or("Couldn't find device")?;
+        let ino_str = parts.next().ok_or("Couldn't find inode")?;
+        let offset_str = parts.next().ok_or("Couldn't find offset")?;
+        let _rsv_str = parts.next().ok_or("Couldn't find reserved pages")?;
+        let _guardsize_str = parts.next().ok_or("Couldn't find guard size")?;
+        let _refcnt_str = parts.next().ok_or("Couldn't find reference count")?;
+        let _mapcnt_str = parts.next().ok_or("Couldn't find mapped count")?;
```
workingjubilee, line 190: Just to be clear, on QNX Neutrino these are all required fields? Is there somewhere this format is documented for Neutrino?

src/symbolize/gimli/parse_running_mmaps_unix.rs
```diff
@@ -53,19 +53,34 @@ pub(super) struct MapsEntry {
...
-        v.push(line.parse()?);
-    }
+        .map_err(|_| concat_str("Couldn't read file: ", file))?;
+    Ok(buf)
+}
 
-    Ok(v)
+pub(super) fn parse_maps() -> Result<Vec<MapsEntry>, String> {
+    let (file, skip) = config();
+    let content = read_file(file)?;
```
workingjubilee, line 74:
```suggestion
    let content = fs::read_to_string(file).map_err(|e| e.to_string())?;
```

src/symbolize/gimli/parse_running_mmaps_unix.rs
```diff
@@ -141,13 +162,73 @@ impl FromStr for MapsEntry {
...
+        let _guardsize_str = parts.next().ok_or("Couldn't find guard size")?;
+        let _refcnt_str = parts.next().ok_or("Couldn't find reference count")?;
+        let _mapcnt_str = parts.next().ok_or("Couldn't find mapped count")?;
+        let pathname_str = parts.next().unwrap_or(""); // pathname may be omitted.
+
+        let hex =
+            |s: &str| usize::from_str_radix(&s[2..], 16).map_err(|_| "Couldn't parse hex number");
+        let address = { (hex(vaddr_str)?, hex(vaddr_str)? + hex(size_str)?) };
+
+        // TODO: Probably a rust'ier way of doing this
```
workingjubilee, line 197: You could probably cook something up using [`array::from_fn`](https://doc.rust-lang.org/stable/std/array/fn.from_fn.html) but I'm not going to block this on that.

workingjubilee (commented):
I had half-expected someone else with a bit more experience to come by and review this, so I kept my earlier comments minimal, but it looks like I probably now know the most about QNX Neutrino amongst the libs reviewers, so I might as well sit down with this. Sorry about the holdup.

workingjubilee:
@japaric The QNX `/proc/self/maps` format still isn't handled.

PR #508, opened 2023-01-12
Add AIX support in gimli symbolizer

bzEq:


ecnelises:
Do we need to upgrade `object` to `0.31.0` and enable `xcoff` feature? (there's one upgrading it #522 )

bzEq:
Running `cargo t`, testing result
```
     Running unittests src/lib.rs (target/debug/deps/backtrace-6d51f080c2298b47)

running 1 test
test capture::tests::test_frame_conversion ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.16s

     Running tests/accuracy/main.rs (target/debug/deps/accuracy-73e2b667931f3055)

running 1 test
test doit ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.21s

     Running tests/concurrent-panics.rs (target/debug/deps/concurrent_panics-4b2a6d5c8a0c925f)
test result: ok
     Running tests/current-exe-mismatch.rs (target/debug/deps/current_exe_mismatch-940c62e347df0787)
test result: ignored
     Running tests/long_fn_name.rs (target/debug/deps/long_fn_name-2f81194cd0acabae)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/skip_inner_frames.rs (target/debug/deps/skip_inner_frames-f8fb6d50f25cef3c)

running 2 tests
test backtrace_new_should_start_with_call_site_trace ... ok
test backtrace_new_unresolved_should_start_with_call_site_trace ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/smoke.rs (target/debug/deps/smoke-02393ceedff1e194)

running 3 tests
test smoke_test_frames ... FAILED
test sp_smoke_test ... ok
test many_threads ... ok

failures:

---- smoke_test_frames stdout ----
thread 'smoke_test_frames' panicked at 'assertion failed: sym - actual_fn_pointer < 1024', tests/smoke.rs:126:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

workingjubilee:
That's weird. More of the tests should be not-failing at this point. Can you rebase this on master? The object bump is merged, for note.

bzEq:
Rebased and fixed build failures.

src/symbolize/gimli/libs_aix.rs
```diff
@@ -0,0 +1,65 @@
...
+        loop {
+            if libc::loadquery(
+                libc::L_GETINFO,
+                buffer.as_mut_ptr() as *mut i8,
+                (mem::size_of::<libc::ld_info>() * buffer.len()) as u32,
+            ) >= 0
+            {
+                break;
+            } else {
+                if Error::last_os_error().raw_os_error().unwrap() != libc::ENOMEM {
```
workingjubilee, line 26: Wait, why are we checking for ENOMEM? Usually Rust programs die on encountering OOM.

src/symbolize/gimli/libs_aix.rs
```diff
@@ -0,0 +1,65 @@
...
+pub(super) fn native_libraries() -> Vec<Library> {
+    let mut ret = Vec::new();
+    unsafe {
+        let mut buffer = vec![mem::zeroed::<libc::ld_info>(); 64];
+        loop {
+            if libc::loadquery(
+                libc::L_GETINFO,
+                buffer.as_mut_ptr() as *mut i8,
+                (mem::size_of::<libc::ld_info>() * buffer.len()) as u32,
+            ) >= 0
```
workingjubilee, line 22: Aren't we checking for "not -1"?

workingjubilee (commented):
Thanks for the rebase! Is there a version of cargo that can build this while cross-compiling? With `-Zbuild-std`, perhaps?

This review is not exhaustive! Rather, having appropriate connections drawn would make it much easier to properly review and, most importantly, maintain, so later updates to this crate don't accidentally burn down all your hard work.

workingjubilee:
Also I realized my initial review includes a lot of _specific_ questions about segments in the code and what the supporting concepts are for explaining them, but you don't need to address them in such a point-by-point manner if you want to take a more high-level approach to documenting, or more likely pointing to existing documentation, of the concepts in question.

tests/smoke.rs
```diff
@@ -1,6 +1,24 @@
 use backtrace::Frame;
 use std::thread;
 
+fn get_actual_fn_pointer(fp: usize) -> usize {
+    // On AIX, the function name references a function descriptor.
+    // A function descriptor is consisted of (See https://reviews.llvm.org/D62532)
+    // * The address of the entry point of the function.
+    // * The TOC base address for the function.
+    // * The environment pointer.
```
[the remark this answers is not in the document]
bjorn3: According to https://refspecs.linuxfoundation.org/ELF/ppc64/PPC-elf64abi.html the environment pointer is a concept for Pascal and PL/1. I can't really find more information about it though unfortunately.

src/symbolize/gimli/libs_aix.rs
```diff
@@ -0,0 +1,65 @@
...
+            let filename_ptr: *const i8 = &(*current).ldinfo_filename[0];
+            let bytes = CStr::from_ptr(filename_ptr).to_bytes();
+            let member_name_ptr = filename_ptr.offset((bytes.len() + 1) as isize);
+            let mut filename = OsStr::from_bytes(bytes).to_owned();
+            if text_base == EXE_IMAGE_BASE as usize {
+                if let Ok(exe) = env::current_exe() {
+                    filename = exe.into_os_string();
+                }
+            }
+            let bytes = CStr::from_ptr(member_name_ptr).to_bytes();
```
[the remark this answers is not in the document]
bzEq: The content `member_name_ptr` points to can be only a `\0`, thus bytes is empty in such case. When `member_name` is empty, it's suggesting that the image is not an archive.

tests/smoke.rs
```diff
@@ -1,6 +1,24 @@
 use backtrace::Frame;
 use std::thread;
 
+fn get_actual_fn_pointer(fp: usize) -> usize {
+    // On AIX, the function name references a function descriptor.
+    // A function descriptor is consisted of (See https://reviews.llvm.org/D62532)
+    // * The address of the entry point of the function.
+    // * The TOC base address for the function.
```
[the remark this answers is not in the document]
bzEq: TOC is a region in `.data` section, which contains some constants and what is GOT in ELF. Will update comment to contain basic knowledge of XCOFF and AIX OS ABI.

tests/smoke.rs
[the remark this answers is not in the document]
bzEq: AIX 64-bit ABI is similar to the document above. However, the latest ABI is not formally documented. One can find information in its assembler manual https://www.ibm.com/docs/en/ssw_aix_72/assembler/assembler_pdf.pdf.

bzEq:
> you don't need to address all of those questions in such a point-by-point manner if you want to take a more high-level approach to documenting (or pointing to existing documentation of) the concepts in question.

I'll try my best to give an overview of AIX ABI. Since the latest AIX ABI is not formally documented, its behaviors largely depend on assembler manual and current implementation of IBM OpenXL/XL.

src/symbolize/gimli/libs_aix.rs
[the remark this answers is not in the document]
workingjubilee: Ohhh, okay, so the two strings together would have to be `"some_text\0\0"` in the case of this member name being empty and I don't have to worry about our strlen potentially reading into infinity.

tests/smoke.rs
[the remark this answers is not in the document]
workingjubilee: Ah so I was right.

I figured it was the GOT-equivalent, it's just nice to have these things spelled out.

workingjubilee:
Ah, I had a feeling it might be something like that. That unfortunately motivates my questions more, since we can't easily look up things and instead have to piece together answers from the scraps we can dig up. But I'm somewhat familiar with typical Power ISA assembly, at least, so this assembler manual is very helpful.

src/symbolize/gimli/libs_aix.rs
```diff
@@ -0,0 +1,65 @@
...
+const EXE_IMAGE_BASE: u64 = 0x100000000;
+
+pub(super) fn native_libraries() -> Vec<Library> {
+    let mut ret = Vec::new();
+    unsafe {
+        let mut buffer = vec![mem::zeroed::<libc::ld_info>(); 64];
+        loop {
+            if libc::loadquery(
+                libc::L_GETINFO,
+                buffer.as_mut_ptr() as *mut i8,
```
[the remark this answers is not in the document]
workingjubilee: Oh.
It should probably have been `*mut c_void` in the libc bindings, then.
You know that the libc crate's stability policy means that you virtually **cannot** change the bindings in the libc crate, right? It does breaking changes only very, very rarely, and considers virtually everything to be breaking.

bzEq:
Gentle ping.

src/symbolize/gimli/libs_aix.rs
```diff
@@ -0,0 +1,68 @@
...
+                (mem::size_of::<libc::ld_info>() * buffer.len()) as u32,
+            ) != -1
+            {
+                break;
+            } else {
+                if Error::last_os_error().raw_os_error().unwrap() != libc::ENOMEM {
+                    return Vec::new();
+                }
+                buffer.resize(buffer.len() * 2, mem::zeroed::<libc::ld_info>());
+            }
```
workingjubilee, line 33: I would prefer this consistently uses if-else-if nesting or consistently use "rely on earlier break/return" instead of this crossing-over between the two styles. My understanding of this control flow is each of these expressions can only be reached if all conditions before them fail to be met, just like in an if-else-if-else chain, so there is no need for the early return to have its own if-statement, unless the entire set is "flat".

I do realize they are conceptually separate but that also means the `break;` doesn't need an `else` after it.

src/symbolize/gimli/libs_aix.rs
```diff
@@ -0,0 +1,68 @@
...
+                buffer.as_mut_ptr() as *mut libc::c_char,
+                (mem::size_of::<libc::ld_info>() * buffer.len()) as u32,
+            ) != -1
+            {
+                break;
+            } else {
+                if Error::last_os_error().raw_os_error().unwrap() != libc::ENOMEM {
+                    return Vec::new();
+                }
+                buffer.resize(buffer.len() * 2, mem::zeroed::<libc::ld_info>());
```
workingjubilee, line 32: Okay, my question re: why the control flow is like this also becomes: is there a reason this unwraps instead of using something like
```suggestion
                 match Error::last_os_error().raw_os_error() {
                     Some(libc::ENOMEM) => buffer.resize(buffer.len() * 2, mem::zeroed::<libc::ld_info>()),
                     Some(libc::EINVAL | libc::EFAULT) => /* something? */,
                     _ => unreachable!(),
                 }
```

When reasoning by negation and positive reasoning are equally available, I prefer to avoid negation.

src/symbolize/gimli/libs_aix.rs
```diff
@@ -0,0 +1,68 @@
...
+            }
+        }
+        let mut current = buffer.as_ptr() as *mut libc::ld_info;
```
workingjubilee, line 35: Is there a reason this goes from `&Vec<libc::ld_info>` to `*const libc::ld_info` to `*mut libc::ld_info` instead of calling `as_mut_ptr`? Does it need to mutate? I'm not seeing any mutation happen on each `ld_info`'s contents? Is there a reason just taking a reference is not okay, does the C library leave it in a partially initialized state?

src/symbolize/gimli/xcoff.rs
```diff
@@ -0,0 +1,184 @@
...
+        let map = super::mmap(path)?;
+        Mapping::mk(map, |data, stash| {
+            if member_name.is_empty() {
+                Context::new(stash, Object::parse(data)?, None, None)
+            } else {
+                let archive = ArchiveFile::parse(data).ok()?;
+                for member in archive.members() {
+                    if let Ok(member) = member {
+                        let name = OsStr::from_bytes(member.name());
+                        if name == member_name {
```
workingjubilee, line 30: While it's not _pretty_ either way, this should look something more like
```suggestion
                for member in archive.members()
                                     .filter_map(|m| m.ok())
                                     .filter(|m| OsStr::from_bytes(m.name()) == member_name) {
```
to eliminate the severe rightward drift. Probably rustfmt disagrees with exactly how I wrote it, even after rebalancing the braces.

src/symbolize/gimli/xcoff.rs
```diff
@@ -0,0 +1,184 @@
...
+    pub size: usize,
+}
+
+pub fn parse_xcoff(data: &[u8]) -> Option<Image> {
+    let mut offset = 0;
+    let header = Xcoff::parse(data, &mut offset).ok()?;
+    let _ = header.aux_header(data, &mut offset).ok()?;
+    let sections = header.sections(data, &mut offset).ok()?;
+    if let Some(section) = sections.iter().find(|s| {
+        let name = str::from_utf8(&s.s_name()[0..5]).unwrap();
```
workingjubilee, line 67: Is this really guaranteed to be UTF-8?

src/symbolize/gimli/xcoff.rs
```diff
@@ -0,0 +1,184 @@
...
+        return name == ".text";
```
workingjubilee, line 68:
```suggestion
        name == ".text"
```

src/symbolize/gimli/xcoff.rs
```diff
@@ -0,0 +1,184 @@
...
+    let map = super::mmap(path)?;
+    let data = map.deref();
+    if member_name.is_empty() {
+        return parse_xcoff(data);
+    } else {
+        let archive = ArchiveFile::parse(data).ok()?;
+        for member in archive.members() {
+            if let Ok(member) = member {
+                let name = OsStr::from_bytes(member.name());
+                if name == member_name {
```
workingjubilee, line 89: Same remarks regarding eliminating rightward drift as earlier.

src/symbolize/gimli/xcoff.rs
```diff
@@ -0,0 +1,184 @@
...
+    fn get_concrete_size(file: &XcoffFile<'a, Xcoff>, sym: &XcoffSymbol<'a, '_, Xcoff>) -> u64 {
+        match sym.flags() {
+            SymbolFlags::Xcoff {
+                n_sclass: _,
+                x_smtyp: _,
+                x_smclas: _,
+                containing_csect: Some(index),
+            } => {
+                if let Ok(tgt_sym) = file.symbol_by_index(index) {
+                    return Self::get_concrete_size(file, &tgt_sym);
```
workingjubilee, line 111:
```suggestion
                    Self::get_concrete_size(file, &tgt_sym)
```

src/symbolize/gimli/xcoff.rs
```diff
@@ -0,0 +1,184 @@
...
+                    Some(ParsedSym {
+                        address,
+                        size,
+                        name,
+                    })
+                }
+            })
+            .collect::<Vec<_>>();
+        syms.sort_by_key(|s| s.address);
+        return Some(Object { syms, file });
```
workingjubilee, line 144:
```suggestion
        Some(Object { syms, file })
```

src/symbolize/gimli/xcoff.rs
```diff
@@ -0,0 +1,184 @@
...
+        let name = str::from_utf8(&s.s_name()[0..5]).unwrap();
+        return name == ".text";
+    }) {
+        return Some(Image {
+            offset: section.s_scnptr() as usize,
+            base: section.s_paddr() as u64,
+            size: section.s_size() as usize,
+        });
+    }
+    return None;
```
workingjubilee, line 76:
```suggestion
        Some(Image {
            offset: section.s_scnptr() as usize,
            base: section.s_paddr() as u64,
            size: section.s_size() as usize,
        });
    } else {
        None
    }
```

workingjubilee (commented):
Took a new look. Thank you for addressing the prior round of review, having scanned the documentation a bit I feel relatively confident we can land this once a few last questions and cleanups are done.

workingjubilee:
Also feel free to squash/rebase this with your next pass of changes to however the commit history should look.

bzEq:
Addressed comment and rebased.

bzEq:
Gentle ping.

workingjubilee:
Excellent, thank you!

PR #509, opened 2023-01-18
(Temporarily?) Disable CI builds for `x86_64-fuchsia`

flba-eb:
CI fails because there is no std lib available:
```
error: toolchain 'nightly-x86_64-unknown-linux-gnu' does not contain component 'rust-std' for target 'x86_64-fuchsia'
```
  Availability monitor: https://rust-lang.github.io/rustup-components-history/

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -204,9 +204,9 @@ jobs:
         target:
         - wasm32-unknown-unknown
         - wasm32-wasi
-        - x86_64-fuchsia
         - x86_64-fortanix-unknown-sgx
         - x86_64-unknown-illumos
+        # - x86_64-fuchsia # error: toolchain 'nightly-x86_64-unknown-linux-gnu' does not contain component 'rust-std' for target 'x86_64-fuchsia'
     steps:
     - uses: actions/checkout@v3
       with:


JohnTitor:
Could we pin the nightly version for that target instead?

workingjubilee:
I agree: using `cargo {command} -Zbuild-std` is much more robust as a solution. It should still be removed eventually since we _should_ be shipping a tier 2 target??? I don't know what's happening there.

Also this should be rebased on master just to make sure the CI run accurately reflects the current CI.

workingjubilee:
Though also, hmmm.

I wonder, is rustbot functional on this repo? Because this should really be

@rustbot ping fuchsia

workingjubilee:
RIP.

JohnTitor:
Ah, I see!
https://github.com/rust-lang/rust/pull/106429 renamed the target and the components are no longer shipped via that name.
Using `x86_64-unknown-fuchsia` should resolve the problem here.
(More info: https://github.com/rust-lang/rust/pull/106636 re-added them to the supported targets, but it doesn't mean we can do `rustup target add x86_64-fuchsia`, it seems.)

flba-eb:
CI for Fuchsia seems to work now 😃 , but not all tests have been run (because of https://github.com/rust-lang/backtrace-rs/issues/511).

JohnTitor (approved):
Let's fix other errors in another PR, thanks!

PR #513, opened 2023-01-30
Handle addr2line API changes.

khuey:
This doesn't actually implement loading for split DWARF but it will make backtrace continue to compile with the existing functionality once addr2line cuts a new release.

Cc @philipc


philipc:
addr2line 0.20.0 has been released.

khuey:
Ok the latest csets I've pushed support *packed* split DWARF on Linux. They don't yet support unpacked split DWARF (that's harder because we need to load the DWOs and presumably stash them somewhere). And bumping addr2line causes the MSRV test to start failing.

philipc:
Is it possible to add a CI test for this?

> They don't yet support unpacked split DWARF (that's harder because we need to load the DWOs and presumably stash them somewhere).

We could hash by DWO id, but an alternative would be to change `SplitDwarfLoad` to include the `ResUnit` index, and add an API to query the number of `ResUnit`s so that we can preallocate a `Vec` for lazy loading (similar to [unpacked Mach-O](https://github.com/rust-lang/backtrace-rs/blob/b3e5bb857773fcc6fc8247374264bd4c8acd5387/src/symbolize/gimli/macho.rs#L168)).

khuey:
There is a CI test for this.

philipc:
> There is a CI test for this.

Oh, so there is, but I must be missing something still. How was it passing before this PR?

khuey:
What you're missing is that it's not passing before this PR. These tests are failing on the tip of this repo.

philipc:
The difference from Mach-O that I overlooked is that `addr2line` is internally doing the indexing itself. I don't think we actually need to look up anything. `addr2line` isn't going to ask for the same DWO twice. All we need is somewhere to stash it before returning it to `addr2line`, so we could just append to a `Vec`.

khuey:
Hmm, yes, addr2line's `ResUnit` will cache the `gimli::Dwarf` for the DWO, and backtrace's `Context` contains addr2line's `Context, so that does sound right.

Perhaps instead of having `mmap_aux` and `mmap_dwp` in `Stash` we should just have an append-only `Vec<Mmap>` like we do for `Stash::buffers`.

khuey:
And that adds support for unpacked split DWARF, passing all the relevant tests.

khuey:
Does @philipc review changes for this repo too?

philipc:
I don't approve PRs in this repo. LGTM though.

khuey:
It would be nice to get this merged and a new release cut. Split DWARF inlining was turned off by default in 1.68.

khuey:
Realistically I don't know that you're going to find a reviewer with more subject matter expertise than me and @philipc here.

Amanieu (approved):
LGTM to my limited understanding of DWARF. I trust @philipc to have reviewed this properly.

src/symbolize/gimli/elf.rs
```diff
@@ -421,3 +449,53 @@ fn locate_debugaltlink(path: &Path, filename: &[u8], build_id: &[u8]) -> Option<
...
+        if let Ok(p) = convert_path(p) {
+            path.push(p);
+        }
+    }
+
+    if let Some(p) = load.path.as_ref() {
+        if let Ok(p) = convert_path(p) {
+            path.push(p);
+        }
+    }
```
philipc, line 480: If either `comp_dir` or `path` are `None`, or if `convert_path` fails, then we should immediately return `None`.

khuey: perhaps for `path=None` and `convert_path` failing, but I think `comp_dir=None` is legal.

philipc: I think you're right.

khuey: Alright, addressed.

khuey:
Can we get a new release with this?

workingjubilee:
It would be nice if we passed the test suite on the commit we cut our release from. I will make sure a release is cut once https://github.com/rust-lang/backtrace-rs/issues/531 is resolved.

workingjubilee:
backtrace 0.3.68 is out, let us know if there are any bugs to be squashed!

PR #522, opened 2023-04-17
Bump object to 0.31

lnicola:


src/symbolize/gimli/coff.rs
```diff
@@ -1,6 +1,7 @@
 use super::{Context, Mapping, Path, Stash, Vec};
 use core::convert::TryFrom;
 use object::pe::{ImageDosHeader, ImageSymbol};
+use object::read::coff::ImageSymbol as _;
```
ecnelises, line 4: It seems we don't need this line to bump object.

lnicola: It's strange, it breaks in a different way if I remove that. I think I'll close this PR, since I can't really test it on Windows.

philipc: You do need this line, but you also need to update https://github.com/rust-lang/backtrace-rs/blob/b3e5bb857773fcc6fc8247374264bd4c8acd5387/crates/as-if-std/Cargo.toml#L22

philipc:
I think the nightly build failure needs #513

workingjubilee (approved):
I don't believe we're gating PRs on that test.

lnicola:
Rebased. The failures seem unrelated, but it's a bit of a mess there :grimacing:.

workingjubilee:
There's a bunch of CI failures, yeah. We're slowly knocking them down. Down to about 3 on latest.

src/symbolize/gimli/coff.rs
chenyukang: @lnicola 
This commit will make `rustc` failing in `mingw-check`:

https://github.com/rust-lang/rust/actions/runs/5407279353/jobs/9825113481

image

philipc: @chenyukang If you're bumping the `backtrace` version in `libstd`, then you also need to bump the `object` version.

chenyukang: oh, sorry, I missed the previous comments.
Got it, thanks!

PR #524, opened 2023-05-12
Correctly set `doc(html_root_url)` attribute

GuillaumeGomez:
I'm working on turning https://github.com/rust-lang/rust/issues/82730 into an error. When I did, the build for this crate failed because when included from `libstd` because of this attribute. This fixes it.


jyn514:
This doesn't seem like the right fix? it's possible to use backtrace as a dependency outside of std without enabling the `std` feature.

I'd suggest one of the following:
- just removing the attribute altogether: https://github.com/rust-lang/api-guidelines/issues/229
- adding a cfg or feature to backtrace-rs that's only set when building as a dependency in rust-lang/rust
- switching backtrace to being used as a crate instead of using `include` or `path`.

GuillaumeGomez:
I think using `backtrace` as dependency rather than including it seems like a better approach. I'll check first why it was done this way.

GuillaumeGomez:
This turns out to be actually quite tricky: `backtrace` cannot be built as a dependency of `std` apparently because it requires `std`, hence why it's included using `path`. I tried to go around for some time but the changes were quite massive and in the end, I'm not sure it's worth it. I decided to just remove the attribute to make things simpler for the time being. But it might be worth discussing simplifying the integration of `backtrace` into `std`.

GuillaumeGomez:
This is ready then. Ping @JohnTitor

Amanieu (approved):

GuillaumeGomez:
Thanks!

PR #526, opened 2023-05-18
make it easy to resolve symbols by frame

fraillt:
Hello,

Resolving backtrace (at least for the first time) is very slow, but this is not something that can be solved (I guess). 
However, not all frames are equally useful, and at least in the code base that I'm working on, we simply ignore bunch of them.
We have code something like this:
```rust
let mut bt = Backtrace::new_unresolved();
// ...
bt.resolve();
let bt: Vec<String> = bt.frames()
    .iter()
    .flat_map(BacktraceFrame::symbols)
    .take_while(is_not_async_stuff)
    .map(format_symbols)
    .collect();
```
For us, it takes takes around ~700ms (sometimes even ~1.5s) to resolve whole backtrace.

With this PR we're able to write like this:
```rust
let bt: Vec<String> = bt.frames_mut()
    .iter_mut()
    .flat_map(BacktraceFrame::symbols_resolved)
    .take_while(is_not_async_stuff)
    .map(format_symbols)
    .collect();
```
And it takes around 25ms to resolve the frames that we actually need.

I'm aware of low level utils `backtrace::trace` and `backtrace::resolve_frame`, but this would basically mean that I would basically need to reimplement Backtrace, including Serialize/Deserialize capability, which is too much.

### DIFF
--- a/src/capture.rs
+++ b/src/capture.rs
@@ -26,9 +26,6 @@ use serde::{Deserialize, Serialize};
 pub struct Backtrace {
     // Frames here are listed from top-to-bottom of the stack
     frames: Vec<BacktraceFrame>,
-    // The index we believe is the actual start of the backtrace, omitting
-    // frames like `Backtrace::new` and `backtrace::trace`.
-    actual_start_index: usize,
 }
 
 fn _assert_send_sync() {
@@ -85,6 +82,27 @@ impl Frame {
                 ..
             } => module_base_address.map(|addr| addr as *mut c_void),
         }
+    }
+
+    /// Resolve all addresses in the frame to their symbolic names.
+    fn resolve_symbols(&self) -> Vec<BacktraceSymbol> {
+        let mut symbols = Vec::new();
+        let sym = |symbol: &Symbol| {
+            symbols.push(BacktraceSymbol {
+                name: symbol.name().map(|m| m.as_bytes().to_vec()),
+                addr: symbol.addr().map(|a| a as usize),
+                filename: symbol.filename().map(|m| m.to_owned()),
+                lineno: symbol.lineno(),
+                colno: symbol.colno(),
+            });
+        };
+        match *self {
+            Frame::Raw(ref f) => resolve_frame(f, sym),
+            Frame::Deserialized { ip, .. } => {
+                resolve(ip as *mut c_void, sym);
+            }
+        }
+        symbols
     }
 }
 
@@ -172,23 +190,22 @@ impl Backtrace {
 
     fn create(ip: usize) -> Backtrace {
         let mut frames = Vec::new();
-        let mut actual_start_index = None;
         trace(|frame| {
             frames.push(BacktraceFrame {
                 frame: Frame::Raw(frame.clone()),
                 symbols: None,
             });
 
-            if frame.symbol_address() as usize == ip && actual_start_index.is_none() {
-                actual_start_index = Some(frames.len());
+            // clear inner frames, and start with call site.
+            if frame.symbol_address() as usize == ip {
+                frames.clear();
             }
+
             true
         });
+        frames.shrink_to_fit();
 
-        Backtrace {
-            frames,
-            actual_start_index: actual_start_index.unwrap_or(0),
-        }
+        Backtrace { frames }
     }
 
     /// Returns the frames from when this backtrace was captured.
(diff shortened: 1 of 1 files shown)


workingjubilee:
I think the motivation you have is good. But I feel slightly nervous about exposing `&'a mut [BacktraceFrame]` and thus `&'a mut BacktraceFrame`, allowing mutations that can make the backtrace inconsistent. This is partly because I've noticed the problems you observe, and I think your direction _overall_ is a good idea: making this crate's backtrace lazier is clearly the "way to go". But exposing the entire slice mutably, and without resolving first, is a footgun both for users, who might reasonably expect `frames` and `frames_mut` to be "identical, but one is mutable", _and_ problematic for later optimizations and better interfaces that might rely on internally avoiding doing work where unnecessary, but now would have to account for "you might have called `frames_mut` and done something with them first". So I think addressing the needs you've cited might be better done by adding a function to Backtrace that returns an iterator that lazily resolves a BacktraceFrame on iteration.

fraillt:
@workingjubilee Thanks for your insights, it makes sense.
I tried to implement lazy iterator, but it's not straightforward as it requires interior mutability. 
I noticed that we already have `From and Into<Vec<BacktraceFrame>> for Backtrace` so I guess we can simply provide a way to resolve each frame manually. 
(I updated revision with minimum changes)

fraillt:
Summary of all changes so far:
* provided new public function `BacktraceFrame::resolve(&mut self)` (this is the essence of the PR)
* replaced `impl Into<Vec<BacktraceFrame>> for Backtrace` with `impl From<Backtrace> for Vec<BacktraceFrame>`
> One should avoid implementing [Into](https://doc.rust-lang.org/std/convert/trait.Into.html) and implement [From](https://doc.rust-lang.org/std/convert/trait.From.html) instead. ([source](https://doc.rust-lang.org/std/convert/trait.Into.html))
* don't expose/store backtrace implementation frames. Now `Backtrace::frames(&self)` returns same frames as in `impl From<Backtrace> for Vec<BacktraceFrame>`.

workingjubilee:
> replaced `impl Into<Vec<BacktraceFrame>> for Backtrace` with `impl From<Backtrace> for Vec<BacktraceFrame>`

I know that's the "correct" style, but [we've had several breakages due to "fixing" this](https://github.com/rust-lang/rust/issues/113238), because this crate gets included directly into std, and sometimes rustc gets confused as to whether impls in this crate are publicly accessible, especially when it comes to resolving type inference.

So, 
unfortunately,
I must tell you:
Do it wrong.

fraillt:
Thanks for fast replay! but I really thing that `impl From<Backtrace> for Vec<BacktraceFrame>` is not an issue. There's my reasoning. 
There's conversion only from Backtrace to BacktraceFrame:
* both are local types, and not widely used.
* both are not generic "container like types", so there's no transitive converions (e.g. impl <T> From<Vec<T>> for Backtrace<T>) or something)
Type inference here should be pretty easy here, and there are a lot of code like this in the the wild.
I really cannot see how this could interfere with other type coercions... :(

If this is really a problem, then:
a) I think the problem is elsewhere, and we should fix that problem instead :)
b) we should be able to "proof" that there's a problem, revert it back and add big comment in code, explaining exactly that is the issue.

workingjubilee:
> Type inference here should be pretty easy here, and there are a lot of code like this in the the wild.
> I really cannot see how this could interfere with other type coercions... :(

Part of why it is not disruptive "in the wild" is because there either already is two implementations to infer types from, or all the types to infer based on local implementations are local and it resolves one of those. There is an unfortunate quirk where if there is only one implementation it is always inferred by default, even though it would be ambiguous if there were two in scope. Thus adding additional directions for `Vec<T>` can weaken type inference even though, by all rights, if the type inference algorithm did not have this quirk, it would not infer in the first place (and possibly instead use some other information to guide inference).

std is unique in appearing in what rounds off to every single crate.
Specifically, `Vec<T>` is in scope in every single crate.

workingjubilee:
So I agree the problem is elsewhere, but I did not want to block accepting this PR on landing a new type inference algorithm in rustc, or making sure type resolution + inference _never_ can "leak" in a more "by construction" manner than how it is currently done.

workingjubilee (approved):

PR #529, opened 2023-06-09
Add an exception for QNX Neutrino 7.0 as a 'no-op' platform

samkearney:
The QNX Neutrino 7.0 target ships a toolchain without a native unwinder implementation. libgcc with unwinding symbols is present in version 7.1 and newer, which is why 7.0 is the only version called out here.


workingjubilee:
Thanks!

workingjubilee (approved):

PR #530, opened 2023-06-19
Fix warning on latest nightly

Amanieu:
This was causing CI to fail.

### DIFF
--- a/crates/as-if-std/src/lib.rs
+++ b/crates/as-if-std/src/lib.rs
@@ -14,7 +14,8 @@ mod __internal {
     pub use std::*;
 }
 
-pub use __internal::*;
+#[allow(unused_imports)]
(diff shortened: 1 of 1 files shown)


workingjubilee (approved):
I assume the continued failure on some platforms is intentional.

Amanieu:
Yes, android CI is broken because of the libunwind changes (and has been broken for a while). I can have a go at fixing it though...

workingjubilee:
That seems to be... less fixed, as it had 3 failures and now it has 5. Can you open an issue for the Android failures and then we can tag in the Android team?

Amanieu:
Issue for Android CI: https://github.com/rust-lang/backtrace-rs/issues/531

workingjubilee:
Thank you!

PR #532, opened 2023-06-21
Add fmt method for BacktraceFmt

chenyukang:
From comment https://github.com/rust-lang/rust/pull/112843#discussion_r1236911788

### DIFF
--- a/src/print.rs
+++ b/src/print.rs
@@ -95,6 +95,14 @@ impl<'a, 'b> BacktraceFmt<'a, 'b> {
     pub fn message(&mut self, msg: &str) -> fmt::Result {
         self.fmt.write_str(msg)
     }
+
+    /// Return the inner formatter.
+    ///
+    /// This is used for writing custom information between frames with `write!` and `writeln!`,
+    /// and won't increment the `frame_index` unlike the `frame` method.
+    pub fn formatter(&mut self) -> &mut fmt::Formatter<'b> {
+        self.fmt
+    }
 }
 
 /// A formatter for just one frame of a backtrace.


JohnTitor (approved):

src/print.rs
```diff
@@ -95,6 +95,14 @@ impl<'a, 'b> BacktraceFmt<'a, 'b> {
     pub fn message(&mut self, msg: &str) -> fmt::Result {
         self.fmt.write_str(msg)
     }
+
+    /// Return the inner formatter.
+    ///
+    /// This is used for writing custom information between frames with `write!` and `writeln!`,
+    /// and won't increment the `frame_index` unlike the `frame` method.
+    pub fn fmt(&mut self) -> &mut fmt::Formatter<'b> {
```
workingjubilee, line 103: With this we're opening up expressions like `displayer.fmt(bt.fmt())` and `bt.fmt().write_fmt(format_args!("{} foo {:?}", 1, 2))`

Here, everything is local and it's clear what the field is, but to head off semantic satiation and/or confusion when it is non-local, could we name this function anything else? `fmtter`?

chenyukang: yes, `fmtter` is a better name.
or `get_fmt`?

workingjubilee: Or `formatter` or whatever seems good.

chenyukang: Fixed, thanks!

workingjubilee:
Looks good, and looks like no one else has a paint can, so I am going to merge this, but first I am going to play a little with CI to see if I can increase my understanding of how GitHub Actions works.

workingjubilee:
Cool! Useful information: "re-run failed jobs" uses the exact commit tested at the moment of failure, but closing and reopening to trigger a new workflow run forces a partial or complete re-reckoning of the repo state as it would be after that PR was merged, i.e. it uses a potentially-different commit. Pushing another commit on an open PR's branch may do either of these, possibly the latter?

workingjubilee:
Thank you!

PR #538, opened 2023-07-02
Upgrade to 2021 ed and inline panics

nyurik:
Panics and asserts do not work the same way with inlined format args, so this updates all crates to 2021 edition.

This PR complements #486, but is independent of it

Note that just like #486, the MSRV must be at least 1.58 (which introduced inline format args)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ A library to acquire a stack trace (backtrace) at runtime in a Rust program.
 """
 autoexamples = true
 autotests = true
-edition = "2018"
+edition = "2021"
 exclude = ["/ci/"]
 
 [workspace]
@@ -118,12 +118,12 @@ required-features = ["std"]
 [[test]]
 name = "smoke"
 required-features = ["std"]
-edition = '2018'
+edition = '2021'
 
 [[test]]
 name = "accuracy"
 required-features = ["std"]
-edition = '2018'
+edition = '2021'
 
 [[test]]
 name = "concurrent-panics"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ autoexamples = true
 autotests = true
 edition = "2021"
 exclude = ["/ci/"]
+rust-version = "1.65.0"
 
 [workspace]
 members = ['crates/cpp_smoke_test', 'crates/as-if-std']
--- a/crates/as-if-std/Cargo.toml
+++ b/crates/as-if-std/Cargo.toml
@@ -2,7 +2,7 @@
 name = "as-if-std"
 version = "0.1.0"
 authors = ["Alex Crichton <alex@alexcrichton.com>"]
-edition = "2018"
+edition = "2021"
 publish = false
 
 [lib]
--- a/crates/debuglink/Cargo.toml
+++ b/crates/debuglink/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "debuglink"
 version = "0.1.0"
-edition = "2018"
+edition = "2021"
 
 [dependencies]
 backtrace = { path = "../.." }
--- a/crates/dylib-dep/Cargo.toml
+++ b/crates/dylib-dep/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "dylib-dep"
 version = "0.1.0"
-edition = "2018"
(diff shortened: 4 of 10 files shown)


tests/accuracy/main.rs
```diff
@@ -103,7 +103,7 @@ fn verify(filelines: &[Pos]) {
         loop {
             let sym = match symbols.next() {
                 Some(sym) => sym,
-                None => panic!("failed to find {}:{}", file, line),
+                None => panic!("{}", "failed to find {file}:{line}"),
```
bjorn3, line 106: This will panic with the literal message `failed to find {file}:{line}` now without substituting `{file}` and `{line}`.

nyurik: @bjorn3 good catch, thanks!

crates/as-if-std/Cargo.toml
```diff
@@ -2,7 +2,8 @@
 name = "as-if-std"
 version = "0.1.0"
 authors = ["Alex Crichton <alex@alexcrichton.com>"]
-edition = "2018"
+edition = "2021"
+rust-version = "1.58.0"
```
workingjubilee, line 6: I would strongly prefer we do not document rust-version in a bunch of example crates if we're just going to pointlessly churn it.
```suggestion
```

crates/debuglink/Cargo.toml
```diff
@@ -1,7 +1,8 @@
 [package]
 name = "debuglink"
 version = "0.1.0"
-edition = "2018"
+edition = "2021"
+rust-version = "1.58.0"
```
workingjubilee, line 5:
```suggestion
```

crates/dylib-dep/Cargo.toml
```diff
@@ -1,7 +1,8 @@
 [package]
 name = "dylib-dep"
 version = "0.1.0"
-edition = "2018"
+edition = "2021"
+rust-version = "1.58.0"
```
workingjubilee, line 5:
```suggestion
```

crates/line-tables-only/Cargo.toml
```diff
@@ -1,7 +1,8 @@
 [package]
 name = "line-tables-only"
 version = "0.1.0"
-edition = "2018"
+edition = "2021"
+rust-version = "1.58.0"
```
workingjubilee, line 5:
```suggestion
```

crates/macos_frames_test/Cargo.toml
```diff
@@ -2,7 +2,8 @@
 name = "macos_frames_test"
 version = "0.1.0"
 authors = ["Aaron Hill <aa1ronham@gmail.com>"]
-edition = "2018"
+edition = "2021"
+rust-version = "1.58.0"
```
workingjubilee, line 6:
```suggestion
```

crates/without_debuginfo/Cargo.toml
```diff
@@ -2,7 +2,8 @@
 name = "without_debuginfo"
 version = "0.1.0"
 authors = ["Alex Crichton <alex@alexcrichton.com>"]
-edition = "2018"
+edition = "2021"
+rust-version = "1.58.0"
```
workingjubilee, line 6:
```suggestion
```

workingjubilee (changes requested):

Cargo.toml
```diff
@@ -13,7 +13,8 @@ A library to acquire a stack trace (backtrace) at runtime in a Rust program.
 """
 autoexamples = true
 autotests = true
-edition = "2018"
+edition = "2021"
+rust-version = "1.58.0"
```
workingjubilee, line 17: MSRV should be 1.69, actually, but this is what's currently in CI, so.
```suggestion
rust-version = "1.65.0"
```

.github/workflows/main.yml
```diff
@@ -229,7 +229,7 @@ jobs:
       with:
         submodules: true
     - name: Install Rust
-      run: rustup update 1.55.0 && rustup default 1.55.0
+      run: rustup update 1.58.0 && rustup default 1.58.0
```
workingjubilee, line 232:
```suggestion
      run: rustup update 1.65.0 && rustup default 1.65.0
```

workingjubilee (approved):

PR #540, opened 2023-07-02
Track permissions as bytes

workingjubilee:
Because we expect certain inputs, we don't need chars. The code also doesn't need to validate exact lengths. This reduces the generated code size slightly.

I have some more exciting experiments which I am hoping may take off more in final Rust binaries, but which increase the intermediate rlib by more, so this is the only unequivocal win.


workingjubilee:
We need to parse up to the pathname anyways, in order to be confident the parse is correct. However, once we have done so, we can simply discard this information, yes. I experimented with this but the generated rlib got bigger, and I figured it might reduce the final built and optimized size, but that required shoving it inside the compiler.

ChrisDenton (approved):
Fair enough!

workingjubilee:
I'm going to hold onto this PR since it almost certainly should result in a (small) final-binary diff until either @Kobzol or myself fix the CI check so graciously added, apologies for the noise.

workingjubilee:
That sounds unlikely!

workingjubilee:
However, more importantly: it sounds! Thanks, @Kobzol!

Kobzol:
Did you expect a larger increase? Currently we just compile a hello world program with `panic!()` with `-O`. We could try other things, like `-Copt-level=s` or `strip`.

workingjubilee:
I expected a decrease, actually! Let me tinker with the build settings, then.

workingjubilee:
First pass: Let's try my hypothesis that discarding all the fields will decrease the built size.

workingjubilee:
Okay, opt-level doesn't make a difference, but 24 bytes less means I am on to something! Maybe!

workingjubilee:
well that backfired lmao.

workingjubilee:
WHAT.

workingjubilee:
"zero-cost abstractions" MY ASS.

Issue #541, opened 2023-07-04
Track binary size impact on Rust executables


workingjubilee:
Rust's backtrace implementation is a significant factor in how large every Rust binary is. We should be testing to see how much commits here regress binary size for stdlib-using binaries, before we land them. We may be able to ~~bait~~ invite some people working on [rustc-perf](https://github.com/rust-lang/rustc-perf) to help us investigate this issue. It is very likely a size increase due to recent commits created a [notable regression](https://github.com/rust-lang/rust/pull/113176), although one that was probably worth the price if so.

workingjubilee:
> [9:30 PM] detly: one really, truly fun factor in all of this is how much of an impact rust versions actually make. I might be missing something, but doing nothing other than switching to nightly changes the post-strip binary size from 358,324 B to 407,484 B
> [9:31 PM] detly: (this is without even using the build-std feature, just the same build --release --target etc. command)
> [9:53 PM] detly: so... building with nightly-2023-07-03 and build-std, I get: 342,000 B. Same thing with nightly-2023-07-01, I get: 313,328 B. That's kind of wild. If I'd done that experiment today, I'd have figured it didn't help much. Fortunately I did it three days ago as well, and remembered that it had a noticeable impact.

So approximately 30KiB.

Repo of interest with further information: https://gitlab.com/detly/prod-self-test

detly:
Thanks for reporting this! If it helps, here's a breakdown from `cargo-bloat`. Note `RUSTC_TARGET_ARCH="mipsel-unknown-linux-musl"`, which tends to have slightly bigger code size overall from `x86_64`.

```none
----
Stripped size: 313328
----

buildbot@3156e9e8fc34:/crate$ cargo +nightly-2023-07-01 bloat -Z build-std=std,panic_abort --release --target $RUSTC_TARGET_ARCH
    Finished release [optimized] target(s) in 0.03s
    Analyzing target/mipsel-unknown-linux-musl/release/prod-self-test

 File  .text     Size          Crate Name
 5.0%   7.2%  19.0KiB            std addr2line::ResDwarf<R>::parse
 4.4%   6.3%  16.6KiB            std std::backtrace_rs::symbolize::gimli::resolve
 2.3%   3.3%   8.8KiB            std std::backtrace_rs::symbolize::gimli::elf::decompress_zlib
 2.2%   3.2%   8.4KiB            std gimli::read::unit::parse_attribute
[...]

PR #542, opened 2023-07-05
CI: add workflow for checking binary size

Kobzol:
This PR adds a CI workflow that tracks the binary size effect of `backtrace` on Rust programs. After each commit to a PR, it downloads rustc, patches in `backtrace` using the base SHA of the PR, builds stage 0 libstd, and compiles a simple program that panics. Then it patches in the updated (head) SHA of the PR, builds stage 0 libstd again, and compiles the program again. If the binary sizes differ, it posts a comment to the PR.

The comment (when I artificially increased the binary size of `backtrace`) can be seen [here](https://github.com/Kobzol/backtrace-rs/pull/1#issuecomment-1621246777).

The workflow takes about 3 minutes on the default `ubuntu-latest` CI runner.

Once/if https://github.com/rust-lang/rust/pull/113341 gets merged, the workflow can be simplified slightly (four lines can be removed).

Fixes: https://github.com/rust-lang/backtrace-rs/issues/541

### DIFF
--- a/.github/workflows/check-binary-size.yml
+++ b/.github/workflows/check-binary-size.yml
@@ -1,0 +1,86 @@
+# This workflow checks if a PR commit has changed the size of a hello world Rust program.
+# It downloads Rustc and compiles two versions of a stage0 compiler - one using the base commit
+# of the PR, and one using the latest commit in the PR.
+# If the size of the hello world program has changed, it posts a comment to the PR.
+name: Check binary size
+
+on:
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  test:
+    name: Check binary size
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    steps:
+      - name: Print info
+        run: |
+          echo "Current SHA: ${{ github.event.pull_request.head.sha }}"
+          echo "Base SHA: ${{ github.event.pull_request.base.sha }}"
+      - name: Clone Rustc
+        uses: actions/checkout@v3
+        with:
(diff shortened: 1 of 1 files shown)


workingjubilee:
Oh excellent! This means I get to try it out on https://twitter.com/capcorvidart/status/1676443948162613250

.github/workflows/check-binary-size.yml
```diff
@@ -0,0 +1,88 @@
...
+        uses: actions/github-script@v6
+        with:
+          script: |
+            const reference = process.env.SIZE_REFERENCE;
+            const updated = process.env.SIZE_UPDATED;
+            const diff = updated - reference;
+            let diff_str = `${diff}B`;
+            if (diff > 0) {
+              diff_str = `+${diff_str}`;
+            }
```
workingjubilee, line 71: One less reassignment would make for zero, which is a very pleasing amount of mutable state in a Java Script:
```suggestion
            const plus = diff > 0 ? "+" : "";
            const diff_str = `${plus}${diff}B";
```

Kobzol: Good point, applied.

workingjubilee (approved):
Thank you so much!

klensy:
Sadly that don't work with windows targets, for example #543 cuts about 30kb from stdlib (all gains from miniz_oxide, but anyway).

Kobzol:
Once the workflow actually works, I don't think that it should be that difficult to also run it on Windows on CI:)

Issue #451, opened 2021-11-09
Make `gimli` / `addr2line` optional on windows


Jasper-Bekkers:
Right now `gimli` is always compiled in it looks like. However, since it only targets DWARF, I think it could be an optional feature on windows.

On our windows builds we seem to be spending a good ~15 seconds building `gimli` (since addr2line needs it) without actually every needing or using it.

workingjubilee:
If a PR is submitted adding the appropriate cfg and diligently modularizing the relevant code, we will likely accept it.

klensy:
Should be fixed by #543

PR #543, opened 2023-07-05
reduce dependencies for windows-msvc targets

klensy:
Some dependencies don't used on windows-msvc targets, so exclude them:
* `miniz_oxide` (+ `adler`) 
* `addr2line` (+ `gimli`)
* `object` (+ `memchr`)
* `libc` too

Will sync with rust repo

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,6 @@ exclude = [
 [dependencies]
 cfg-if = "1.0"
 rustc-demangle = "0.1.4"
-libc = { version = "0.2.146", default-features = false }
 
 # Optionally enable the ability to serialize a `Backtrace`, controlled through
 # the `serialize-*` features below.
@@ -38,10 +37,12 @@ rustc-serialize = { version = "0.3", optional = true }
 # Optionally demangle C++ frames' symbols in backtraces.
 cpp_demangle = { default-features = false, version = "0.4.0", optional = true, features = ["alloc"] }
 
-addr2line = { version = "0.21.0", default-features = false }
+[target.'cfg(not(all(windows, target_env = "msvc", not(target_vendor = "uwp"))))'.dependencies]
 miniz_oxide = { version = "0.7.0", default-features = false }
+addr2line = { version = "0.21.0", default-features = false }
+libc = { version = "0.2.146", default-features = false }
 
-[dependencies.object]
+[target.'cfg(not(all(windows, target_env = "msvc", not(target_vendor = "uwp"))))'.dependencies.object]
 version = "0.32.0"
 default-features = false
 features = ['read_core', 'elf', 'macho', 'pe', 'unaligned', 'archive']
--- a/crates/as-if-std/Cargo.toml
+++ b/crates/as-if-std/Cargo.toml
@@ -15,10 +15,12 @@ bench = false
 cfg-if = "1.0"
 rustc-demangle = "0.1.21"
 libc = { version = "0.2.146", default-features = false }
-addr2line = { version = "0.21.0", default-features = false, optional = true }
-miniz_oxide = { version = "0.7", default-features = false }
 
-[dependencies.object]
+[target.'cfg(not(all(windows, target_env = "msvc", not(target_vendor = "uwp"))))'.dependencies]
+miniz_oxide = { version = "0.7.0", optional = true, default-features = false }
+addr2line = { version = "0.21.0", optional = true, default-features = false }
+
+[target.'cfg(not(all(windows, target_env = "msvc", not(target_vendor = "uwp"))))'.dependencies.object]
 version = "0.32.0"
 default-features = false
 optional = true
@@ -30,4 +32,4 @@ cc = "1.0.67"
 
 [features]
 default = ['backtrace']
-backtrace = ['addr2line', 'object']
+backtrace = ['addr2line', 'miniz_oxide', 'object']
--- a/src/symbolize/mod.rs
+++ b/src/symbolize/mod.rs
@@ -471,7 +471,7 @@ cfg_if::cfg_if! {
         mod dbghelp;
         use dbghelp as imp;
     } else if #[cfg(all(
-        any(unix, windows),
+        any(unix, all(windows, target_env = "gnu")),
         not(target_vendor = "uwp"),
         not(target_os = "emscripten"),
         any(not(backtrace_in_libstd), feature = "backtrace"),


klensy:
Should anything be fixed in https://github.com/rust-lang/backtrace-rs/blob/master/crates/as-if-std/Cargo.toml?

Looking at
https://github.com/rust-lang/backtrace-rs/blob/e1c49fbd6124a1b626cdf19871aff68c362bdf07/src/symbolize/mod.rs#L466-L485
gimli module never used on windows-msvc targets?

addr2line used only in gimli module which is unused on windows-msvc and backtrace_in_libstd feature?

workingjubilee:
I believe this would/should address https://github.com/rust-lang/backtrace-rs/issues/451

klensy:
> Should anything be fixed in https://github.com/rust-lang/backtrace-rs/blob/master/crates/as-if-std/Cargo.toml?

Last question left.

workingjubilee:
So I *believe* they should have matching cfgs for their deps, especially if that would allow us to build less deps while building std? But perhaps I am wrong somehow?

philipc:
Yes, the deps in `as-if-std/Cargo.toml` must match the main `Cargo.toml`. They are only duplicated there because there's no other way to do it, not because they might be different.

klensy:
Okay, looks like it ready.

klensy:
nano ping.

klensy:
Can this be reviewed, so able to be merged in rust repo? Thanks.

workingjubilee:
This is still on my radar, my most recent agenda has been pushing a few PRs that were opened in 2022(!) or even 2021(!) over the edge now that some of them have been unblocked.

workingjubilee:
Blech, merge conflicts...

crates/as-if-std/Cargo.toml
```diff
@@ -15,15 +15,12 @@ bench = false
...
 
-[dependencies.object]
-version = "0.31.1"
-default-features = false
-optional = true
-features = ['read_core', 'elf', 'macho', 'pe', 'unaligned', 'archive']
+[target.'cfg(not(all(windows, target_env = "msvc", not(target_vendor = "uwp"))))'.dependencies]
+miniz_oxide = { version = "0.7.0", optional = true, default-features = false }
+addr2line = { version = "0.20.0", optional = true, default-features = false }
+object = { version = "0.31.1", default-features = false, optional = true, features = ['read_core', 'elf', 'macho', 'pe', 'unaligned', 'archive'] }
```
workingjubilee, line 22: I would prefer we keep this as an "exploded-out" table here, as recent events have demonstrated that: it is extremely likely this will be diffed and that it may only diff a single field, so I would prefer to minimize the number of merge conflicts we have. I realize this means it looks like

```toml
[target.'cfg(not(all(windows, target_env = "msvc", not(target_vendor = "uwp"))))'.dependencies.object]
version = "0.32.0"
default-features = false
optional = true
features = ['read_core', 'elf', 'macho', 'pe', 'unaligned', 'archive']
```

Cargo.toml
```diff
@@ -37,13 +36,11 @@ rustc-serialize = { version = "0.3", optional = true }
...
-addr2line = { version = "0.20.0", default-features = false }
+[target.'cfg(not(all(windows, target_env = "msvc", not(target_vendor = "uwp"))))'.dependencies]
 miniz_oxide = { version = "0.7.0", default-features = false }
-
-[dependencies.object]
-version = "0.31.1"
-default-features = false
-features = ['read_core', 'elf', 'macho', 'pe', 'unaligned', 'archive']
+addr2line = { version = "0.20.0", default-features = false }
+object = { version = "0.31.1", default-features = false, features = ['read_core', 'elf', 'macho', 'pe', 'unaligned', 'archive'] }
```
workingjubilee, line 42: Same remark, this should be

```toml
[target.'cfg(not(all(windows, target_env = "msvc", not(target_vendor = "uwp"))))'.dependencies.object]
version = "0.32.0"
default-features = false
optional = true
features = ['read_core', 'elf', 'macho', 'pe', 'unaligned', 'archive']
```

workingjubilee (commented):
Sorry for not getting back to this earlier!

crates/as-if-std/Cargo.toml
klensy: For some reason, `optional = true` exist there, but not in top Cargo.toml

Cargo.toml
klensy: No `optional = true` here.

workingjubilee: Oh true.

workingjubilee (approved):
Nice!

klensy:
For some reason CI have troubles downloading syn on for windows.

ChrisDenton:
syn has been causing a lot of intermittent issues in CI lately, across a number of different repositories.

workingjubilee:
It is possible that the author's recent choice to ship precompiled binaries for certain relevant proc-macros is taxing various points of network IO more severely: https://github.com/serde-rs/serde/issues/2538

pacak:
> is taxing various points of network IO more severely

The binary in question is 680kb compresses down to 300kb. Shipping a binary is not the best idea, but I don't think it should affect network that much.

workingjubilee:
Yah, [dtolnay thinks it's a cargo bug](https://github.com/serde-rs/serde/issues/2577) and I think he's right.

klensy:
Funny thing, now i see this here: https://github.com/rust-lang/rust/actions/runs/5887880758/job/15968022846?pr=114923
```
Thu, 17 Aug 2023 07:15:39 GMT       Updating crates.io index
Thu, 17 Aug 2023 07:16:10 GMT   warning: spurious network error (3 tries remaining): [28] Timeout was reached (Operation too slow. Less than 10 bytes/sec transferred the last 30 seconds)
Thu, 17 Aug 2023 07:16:40 GMT   warning: spurious network error (2 tries remaining): [28] Timeout was reached (Operation too slow. Less than 10 bytes/sec transferred the last 30 seconds)
Thu, 17 Aug 2023 07:17:14 GMT   warning: spurious network error (1 tries remaining): [28] Timeout was reached (Operation too slow. Less than 10 bytes/sec transferred the last 30 seconds)
Thu, 17 Aug 2023 07:17:50 GMT   error: failed to get `zerofrom-derive` as a dependency of package `zerofrom v0.1.2`
Thu, 17 Aug 2023 07:17:50 GMT       ... which satisfies dependency `zerofrom = "^0.1.1"` (locked to 0.1.2) of package `icu_provider v1.2.0`
Thu, 17 Aug 2023 07:17:50 GMT       ... which satisfies dependency `icu_provider = "^1.2.0"` (locked to 1.2.0) of package `icu_list v1.2.0`
Thu, 17 Aug 2023 07:17:50 GMT       ... which satisfies dependency `icu_list = "^1.2"` (locked to 1.2.0) of package `rustc_baked_icu_data v0.0.0 (/checkout/compiler/rustc_baked_icu_data)`
Thu, 17 Aug 2023 07:17:50 GMT       ... which satisfies path dependency `rustc_baked_icu_data` (locked to 0.0.0) of package `rustc_error_messages v0.0.0 (/checkout/compiler/rustc_error_messages)`
Thu, 17 Aug 2023 07:17:50 GMT       ... which satisfies path dependency `rustc_error_messages` (locked to 0.0.0) of package `rustc_driver_impl v0.0.0 (/checkout/compiler/rustc_driver_impl)`
Thu, 17 Aug 2023 07:17:50 GMT       ... which satisfies path dependency `rustc_driver_impl` (locked to 0.0.0) of package `rustc-main v0.0.0 (/checkout/compiler/rustc)`
```

PR #546, opened 2023-07-06
CI: switch to `pull_request_target`

Kobzol:
I don't know how else to test this other than just merging the first commit.


workingjubilee (approved):
Such is life!

PR #548, opened 2023-07-09
Use a safer way for posting comments on CI

Kobzol:

PR #549, opened 2023-07-10
Modularise CI workflow and validate outputs for binary size checks.

detly:
I was amazed at how quickly you spun up your code size check CI workflow. It prompted me to take that approach and play around with it in a private repo to dig into it more, which resulted in these changes and notes you might be interested in.

Notes:

## Modularisation

Most of the workflow structure changes were just to help me understand the process. I found it useful to separate the step that does the code size check itself from all the prep work. This same separation lent itself well to eg. converting the workflow into a Docker multi-stage image so I could automate running the measurement over a lot of commits in the history.

Modularising CI workflow this way is kind of a style preference though, so feel free to decline/bikeshed that aspect of it.

## Compiler flags

I changed the flags to optimise for size and strip all symbols, because that is generally the lowest hanging fruit for anyone concerned about binary size and might reflect the more "realistic" impact on such a project. OTOH, I don't think the difference in flags would actually turn a change in size into a non-change, so you might consider it spurious.

## Transitive dependency breaks

One large-ish problem I found with _historical_ analysis was that, specifically, I wanted to understand the jump in size between `nightly-2023-07-01` and `nightly-2023-07-03`. This was a change in backtrace:

- from 0.3.67 / `8ad84ca5ad88ade697637387e7cb4d7c3cf4bde8`
- to 0.3.68 / `e1c49fbd6124a1b626cdf19871aff68c362bdf07`

The problem is, running the "build with patched std" procedure on the old commit resulted in errors:

<details>
<summary>
(Expand for the full output.)

```none
   --> library/std/src/../../backtrace/src/symbolize/gimli.rs:377:24
    |
377 |                 if let Ok(mut frames) = object_cx.dwarf.find_frames(object_addr) {
    |                        ^^^^^^^^^^^^^^   ---------------------------------------- this expression has type `LookupResult<impl LookupContinuation<Output = core::result::Result<FrameIter<'_, EndianSlice<'_, addr2line::gimli::LittleEndian>>, addr2line::gimli::Error>, Buf = EndianSlice<'_, addr2line::gimli::LittleEndian>>>`
    |                        |
    |                        expected `LookupResult<impl LookupContinuation<Output = ..., Buf = ...>>`, found `Result<_, _>`
    |
   ::: /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/addr2line-0.20.0/src/lib.rs:450:23
    |
450 | ...lt<impl LookupContinuation<Output = Result<FrameIter<'_, R>, Error>, Buf = R>>
    |       -------------------------------------------------------------------------- the expected opaque type
    |
    = note: expected enum `LookupResult<impl LookupContinuation<Output = ..., Buf = ...>>`
            the full type name has been written to '/rust/build/x86_64-unknown-linux-gnu/stage0-std/x86_64-unknown-linux-gnu/release/deps/std-561bd21e49def8cf.long-type-9162012492364665641.txt'
               found enum `core::result::Result<_, _>`
```
</summary>

```none
root@3770f30d41d9:/rust# python3 x.py build library --stage 0
Building bootstrap
    Finished dev [unoptimized] target(s) in 0.04s
warning: x.py has made several changes recently you may want to look at
help: consider looking at the changes in `src/bootstrap/CHANGELOG.md`
note: to silence this warning, add `changelog-seen = 2` at the top of `config.toml`
Updating submodule src/tools/cargo
Submodule 'src/tools/cargo' (https://github.com/rust-lang/cargo.git) registered for path 'src/tools/cargo'
Cloning into '/rust/src/tools/cargo'...
remote: Enumerating objects: 1896, done.        
remote: Counting objects: 100% (1896/1896), done.        
remote: Compressing objects: 100% (1358/1358), done.        
remote: Total 1896 (delta 468), reused 1094 (delta 343), pack-reused 0        
Receiving objects: 100% (1896/1896), 2.73 MiB | 1.73 MiB/s, done.
Resolving deltas: 100% (468/468), done.
remote: Total 0 (delta 0), reused 0 (delta 0), pack-reused 0
remote: Enumerating objects: 32, done.
remote: Counting objects: 100% (32/32), done.
remote: Compressing objects: 100% (14/14), done.
remote: Total 17 (delta 15), reused 5 (delta 3), pack-reused 0
Unpacking objects: 100% (17/17), 6.95 KiB | 790.00 KiB/s, done.
From https://github.com/rust-lang/cargo
 * branch            5b377cece0e0dd0af686cf53ce4637d5d85c2a10 -> FETCH_HEAD
Submodule path 'src/tools/cargo': checked out '5b377cece0e0dd0af686cf53ce4637d5d85c2a10'
Updating submodule library/backtrace
Updating submodule library/stdarch
Submodule 'library/stdarch' (https://github.com/rust-lang/stdarch.git) registered for path 'library/stdarch'
Cloning into '/rust/library/stdarch'...
remote: Enumerating objects: 302, done.        
remote: Counting objects: 100% (302/302), done.        
remote: Compressing objects: 100% (247/247), done.        
remote: Total 302 (delta 45), reused 155 (delta 22), pack-reused 0        
Receiving objects: 100% (302/302), 1.16 MiB | 2.60 MiB/s, done.
Resolving deltas: 100% (45/45), done.
remote: Total 0 (delta 0), reused 0 (delta 0), pack-reused 0
remote: Enumerating objects: 87, done.
remote: Counting objects: 100% (87/87), done.
remote: Compressing objects: 100% (46/46), done.
remote: Total 50 (delta 33), reused 13 (delta 3), pack-reused 0
Unpacking objects: 100% (50/50), 104.11 KiB | 1.12 MiB/s, done.
From https://github.com/rust-lang/stdarch
 * branch            d77878b7299dd7e286799a6e8447048b65d2a861 -> FETCH_HEAD
Submodule path 'library/stdarch': checked out 'd77878b7299dd7e286799a6e8447048b65d2a861'
Updating submodule library/backtrace
downloading https://static.rust-lang.org/dist/2023-05-30/rustfmt-nightly-x86_64-unknown-linux-gnu.tar.xz
######################################################################################### 100.0%
extracting /rust/build/cache/2023-05-30/rustfmt-nightly-x86_64-unknown-linux-gnu.tar.xz to /rust/build/x86_64-unknown-linux-gnu/rustfmt
downloading https://static.rust-lang.org/dist/2023-05-30/rustc-nightly-x86_64-unknown-linux-gnu.tar.xz
######################################################################################### 100.0%
extracting /rust/build/cache/2023-05-30/rustc-nightly-x86_64-unknown-linux-gnu.tar.xz to /rust/build/x86_64-unknown-linux-gnu/rustfmt
downloading https://ci-artifacts.rust-lang.org/rustc-builds/4b6749b21e680a6280cf05ace533ae20c93d9bff/rust-dev-nightly-x86_64-unknown-linux-gnu.tar.xz
######################################################################################### 100.0%
extracting /rust/build/cache/llvm-4b6749b21e680a6280cf05ace533ae20c93d9bff-false/rust-dev-nightly-x86_64-unknown-linux-gnu.tar.xz to /rust/build/x86_64-unknown-linux-gnu/ci-llvm
Building stage0 library artifacts (x86_64-unknown-linux-gnu)
    Updating crates.io index
  Downloaded addr2line v0.20.0
  Downloaded miniz_oxide v0.7.1
  Downloaded unicode-width v0.1.10
  Downloaded adler v1.0.2
  Downloaded getopts v0.2.21
  Downloaded rustc-demangle v0.1.23
  Downloaded allocator-api2 v0.2.15
  Downloaded cc v1.0.79
  Downloaded hashbrown v0.14.0
  Downloaded compiler_builtins v0.1.93
  Downloaded libc v0.2.147
  Downloaded gimli v0.27.3
  Downloaded 12 crates (1.6 MB) in 1.34s
   Compiling compiler_builtins v0.1.93
   Compiling core v0.0.0 (/rust/library/core)
   Compiling libc v0.2.147
   Compiling cc v1.0.79
   Compiling memchr v2.5.0
   Compiling std v0.0.0 (/rust/library/std)
   Compiling unwind v0.0.0 (/rust/library/unwind)
   Compiling rustc-std-workspace-core v1.99.0 (/rust/library/rustc-std-workspace-core)
   Compiling alloc v0.0.0 (/rust/library/alloc)
   Compiling cfg-if v1.0.0
   Compiling adler v1.0.2
   Compiling rustc-demangle v0.1.23
   Compiling rustc-std-workspace-alloc v1.99.0 (/rust/library/rustc-std-workspace-alloc)
   Compiling panic_abort v0.0.0 (/rust/library/panic_abort)
   Compiling panic_unwind v0.0.0 (/rust/library/panic_unwind)
   Compiling gimli v0.27.3
   Compiling hashbrown v0.14.0
   Compiling miniz_oxide v0.7.1
   Compiling object v0.31.1
   Compiling std_detect v0.1.5 (/rust/library/stdarch/crates/std_detect)
   Compiling addr2line v0.20.0
   --> library/std/src/../../backtrace/src/symbolize/gimli.rs:361:16
    |
361 |         if let Ok(mut frames) = cx.dwarf.find_frames(addr as u64) {
    |                ^^^^^^^^^^^^^^   --------------------------------- this expression has type `LookupResult<impl LookupContinuation<Output = core::result::Result<FrameIter<'_, EndianSlice<'_, addr2line::gimli::LittleEndian>>, addr2line::gimli::Error>, Buf = EndianSlice<'_, addr2line::gimli::LittleEndian>>>`
    |                |
    |                expected `LookupResult<impl LookupContinuation<Output = ..., Buf = ...>>`, found `Result<_, _>`
    |
   ::: /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/addr2line-0.20.0/src/lib.rs:450:23
    |
450 | ...lt<impl LookupContinuation<Output = Result<FrameIter<'_, R>, Error>, Buf = R>>
    |       -------------------------------------------------------------------------- the expected opaque type
    |
    = note: expected enum `LookupResult<impl LookupContinuation<Output = ..., Buf = ...>>`
            the full type name has been written to '/rust/build/x86_64-unknown-linux-gnu/stage0-std/x86_64-unknown-linux-gnu/release/deps/std-561bd21e49def8cf.long-type-12642544034931188750.txt'
               found enum `core::result::Result<_, _>`

   --> library/std/src/../../backtrace/src/symbolize/gimli.rs:377:24
    |
377 |                 if let Ok(mut frames) = object_cx.dwarf.find_frames(object_addr) {
    |                        ^^^^^^^^^^^^^^   ---------------------------------------- this expression has type `LookupResult<impl LookupContinuation<Output = core::result::Result<FrameIter<'_, EndianSlice<'_, addr2line::gimli::LittleEndian>>, addr2line::gimli::Error>, Buf = EndianSlice<'_, addr2line::gimli::LittleEndian>>>`
    |                        |
    |                        expected `LookupResult<impl LookupContinuation<Output = ..., Buf = ...>>`, found `Result<_, _>`
    |
   ::: /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/addr2line-0.20.0/src/lib.rs:450:23
    |
450 | ...lt<impl LookupContinuation<Output = Result<FrameIter<'_, R>, Error>, Buf = R>>
    |       -------------------------------------------------------------------------- the expected opaque type
    |
    = note: expected enum `LookupResult<impl LookupContinuation<Output = ..., Buf = ...>>`
            the full type name has been written to '/rust/build/x86_64-unknown-linux-gnu/stage0-std/x86_64-unknown-linux-gnu/release/deps/std-561bd21e49def8cf.long-type-9162012492364665641.txt'
               found enum `core::result::Result<_, _>`

For more information about this error, try `rustc --explain E0308`.
error: could not compile `std` (lib) due to 2 previous errors
Build completed unsuccessfully in 0:03:03
```
</details>

I'm not familiar enough with backtrace and std to interpret this. If it's because there's a change in interface between the two versions of backtrace, leading to incompatibilities where you can't just splice the new or old version into the same version of std — I don't know of a way around it with this approach except to do it manually with known, specific versions of std.

OTOH, if it's because of some transitive dependencies that are now mismatched between those calculated in some lockfile and those resolved during this later step, that might be something that could be dealt with in this process.

The reason I mention it is that this class of code size increase (due to changes in dependencies or ones use of them) can be amongst the biggest and most well-hidden (IME, anyway, because of the "but I didn't change any code!" factor). And so it might be the most worthy of being caught by CI, but I can't quite tell (a) if this approach misses it and, if so, (b) how to change it so it doesn't miss it.

## Sudden failure of copying files

**Update:** fixed now.

I was going to point you to [an example PR](https://github.com/detly/backtrace-rs/pull/3) I had in my own fork so you could verify that this worked. BUT in, literally, the 20 minutes it took me to write this PR, the workflow broke. Including ones that worked an hour ago. The [failure is](https://github.com/detly/backtrace-rs/actions/runs/5502836396/jobs/10027446552?pr=3#step:6:1051):

```none
cp: 'build/x86_64-unknown-linux-gnu/stage0/lib/libLLVM-16-rust-1.71.0-beta.so' and 'build/x86_64-unknown-linux-gnu/stage0-sysroot/lib/libLLVM-16-rust-1.71.0-beta.so' are the same file
cp: 'build/x86_64-unknown-linux-gnu/stage0/lib/librustc_driver-e73eb5c47f849c7b.so' and 'build/x86_64-unknown-linux-gnu/stage0-sysroot/lib/librustc_driver-e73eb5c47f849c7b.so' are the same file
cp: 'build/x86_64-unknown-linux-gnu/stage0/lib/libstd-9ef2a438251e3a63.so' and 'build/x86_64-unknown-linux-gnu/stage0-sysroot/lib/libstd-9ef2a438251e3a63.so' are the same file
cp: 'build/x86_64-unknown-linux-gnu/stage0/lib/libtest-c03aaf58feba71c6.so' and 'build/x86_64-unknown-linux-gnu/stage0-sysroot/lib/libtest-c03aaf58feba71c6.so' are the same file
```

I haven't debugged this yet, but it might be that something in the bootstrap process is (now) hardlinking outputs, and `cp` doesn't like that.

### DIFF
--- a/.github/actions/build-with-patched-std/action.yml
+++ b/.github/actions/build-with-patched-std/action.yml
@@ -1,0 +1,48 @@
+# Github composite action to build a single-source-file test binary with an
+# already-checked-out version of Rust's stdlib, that will be patched with a
+# given revision of the backtrace crate.
+
+name: Build with patched std
+description: >
+  Build a binary with a version of std that's had a specific revision of
+  backtrace patched in.
+inputs:
+  backtrace-commit:
+    description: The git commit of backtrace to patch in to std
+    required: true
+  main-rs:
+    description: The (single) source code file to compile
+    required: true
+  rustc-dir:
+    description: The root directory of the rustc repo
+    required: true
+outputs:
+  test-binary-size:
+    description: The size in bytes of the built test binary
+    value: ${{ steps.measure.outputs.test-binary-size }}
+runs:
+  using: composite
+  steps:
+    - shell: bash
+      id: measure
+      env:
+        RUSTC_FLAGS: -Copt-level=3 -Cstrip=symbols
+        # This symlink is made by Build::new() in the bootstrap crate, using a
+        # symlink on Linux and a junction on Windows, so it will exist on both
+        # platforms.
+        RUSTC_BUILD_DIR: build/host
+      working-directory: ${{ inputs.rustc-dir }}
+      run: |
+        rm -rf "$RUSTC_BUILD_DIR/stage0-std"
+
+        (cd library/backtrace && git checkout ${{ inputs.backtrace-commit }})
+        git add library/backtrace
+
+        python3 x.py build library --stage 0
+
+        TEMP_BUILD_OUTPUT=$(mktemp test-binary-XXXXXXXX)
+        "$RUSTC_BUILD_DIR/stage0-sysroot/bin/rustc" $RUSTC_FLAGS "${{ inputs.main-rs }}" -o "$TEMP_BUILD_OUTPUT"
+        BINARY_SIZE=$(stat -c '%s' "$TEMP_BUILD_OUTPUT")
+        rm "$TEMP_BUILD_OUTPUT"
+
+        echo "test-binary-size=$BINARY_SIZE" >> "$GITHUB_OUTPUT"
--- a/.github/actions/report-code-size-changes/action.yml
+++ b/.github/actions/report-code-size-changes/action.yml
@@ -1,14 +1,21 @@
-# Github composite action to report on code size changes
+# Github composite action to report on code size changes across different
+# platforms.
+
 name: Report binary size changes on PR
 description: |
-  Report on code size changes resulting from a PR as a comment on the PR
-  (accessed via context).
+  Report on code size changes across different platforms resulting from a PR.
+  The only input argument is the path to a directory containing a set of
+  "*.json" files (extension required), each file containing the keys:
+
+   - platform: the platform that the code size change was measured on
+   - reference: the size in bytes of the reference binary (base of PR)
+   - updated: the size in bytes of the updated binary (head of PR)
+
+  The size is reported as a comment on the PR (accessed via context).
 inputs:
-   reference:
-    description: The size in bytes of the reference binary (base of PR).
-    required: true
-   updated:
-    description: The size in bytes of the updated binary (head of PR).
+  data-directory:
+    description: >
+      Path to directory containing size data as a set of "*.json" files.
     required: true
 runs:
   using: composite
@@ -16,56 +23,89 @@ runs:
     - name: Post a PR comment if the size has changed
       uses: actions/github-script@v6
       env:
-        SIZE_REFERENCE: ${{ inputs.reference }}
-        SIZE_UPDATED: ${{ inputs.updated }}
+        DATA_DIRECTORY: ${{ inputs.data-directory }}
       with:
         script: |
-          const reference = process.env.SIZE_REFERENCE;
-          const updated = process.env.SIZE_UPDATED;
+          const fs = require("fs");
 
-          if (!(reference > 0)) {
-            core.setFailed(`Reference size invalid: ${reference}`);
-            return;
-          }
+          const size_dir = process.env.DATA_DIRECTORY;
 
-          if (!(updated > 0)) {
-            core.setFailed(`Updated size invalid: ${updated}`);
-            return;
-          }
+          // Map the set of all the *.json files into an array of objects.
+          const globber = await glob.create(`${size_dir}/*.json`);
+          const files = await globber.glob();
+          const sizes = files.map(path => {
+            const contents = fs.readFileSync(path);
+            return JSON.parse(contents);
+          });
 
-          const formatter = Intl.NumberFormat("en", {useGrouping: "always"});
+          // Map each object into some text, but only if it shows any difference
+          // to report.
+          const size_reports = sizes.flatMap(size_data => {
+            const platform = size_data["platform"];
+            const reference = size_data["reference"];
+            const updated = size_data["updated"];
 
-          const updated_str = formatter.format(updated);
-          const reference_str = formatter.format(reference);
+            if (!(reference > 0)) {
+              core.setFailed(`Reference size invalid: ${reference}`);
+              return;
+            }
 
-          const diff = updated - reference;
-          const diff_pct = (updated / reference) - 1;
+            if (!(updated > 0)) {
+              core.setFailed(`Updated size invalid: ${updated}`);
+              return;
+            }
 
-          const diff_str = Intl.NumberFormat("en", {
-            useGrouping: "always",
-            sign: "exceptZero"
-          }).format(diff);
+            const formatter = Intl.NumberFormat("en", {
+              useGrouping: "always"
+            });
 
-          const diff_pct_str = Intl.NumberFormat("en", {
-            style: "percent",
-            useGrouping: "always",
-            sign: "exceptZero",
-            maximumFractionDigits: 2
-          }).format(diff_pct);
+            const updated_str = formatter.format(updated);
+            const reference_str = formatter.format(reference);
 
-          if (diff !== 0) {
-            // The body is created here and wrapped so "weirdly" to avoid whitespace at the start of the lines,
-            // which is interpreted as a code block by Markdown.
-            const body = `Below is the size of a hello-world Rust program linked with libstd with backtrace.
+            const diff = updated - reference;
+            const diff_pct = (updated / reference) - 1;
 
-          Original binary size: **${reference_str} B**
-          Updated binary size: **${updated_str} B**
-          Difference: **${diff_str} B** (${diff_pct_str})`;
+            const diff_str = Intl.NumberFormat("en", {
+              useGrouping: "always",
+              sign: "exceptZero"
+            }).format(diff);
+
+            const diff_pct_str = Intl.NumberFormat("en", {
+              style: "percent",
+              useGrouping: "always",
+              sign: "exceptZero",
+              maximumFractionDigits: 2
+            }).format(diff_pct);
+
+            if (diff !== 0) {
+              // The body is created here and wrapped so "weirdly" to avoid whitespace at the start of the lines,
+              // which is interpreted as a code block by Markdown.
+              const report = `On platform \`${platform}\`:
+
+          - Original binary size: **${reference_str} B**
+          - Updated binary size: **${updated_str} B**
+          - Difference: **${diff_str} B** (${diff_pct_str})
+
+          `;
+
+              return [report];
+            } else {
+              return [];
+            }
+          });
+
+          // If there are any size changes to report, format a comment and post
+          // it.
+          if (size_reports.length > 0) {
+            const comment_sizes = size_reports.join("");
+            const body = `Code size changes for a hello-world Rust program linked with libstd with backtrace:
+
+          ${comment_sizes}`;
 
             github.rest.issues.createComment({
               issue_number: context.issue.number,
               owner: context.repo.owner,
               repo: context.repo.repo,
               body
-            })
+            });
           }
--- a/.github/actions/report-code-size-changes/action.yml
+++ b/.github/actions/report-code-size-changes/action.yml
@@ -33,19 +33,34 @@ runs:
             return;
           }
 
+          const formatter = Intl.NumberFormat("en", {useGrouping: "always"});
+
+          const updated_str = formatter.format(updated);
+          const reference_str = formatter.format(reference);
+
           const diff = updated - reference;
-          const plus = diff > 0 ? "+" : "";
-          const diff_str = `${plus}${diff}B`;
+          const diff_pct = (updated / reference) - 1;
+
+          const diff_str = Intl.NumberFormat("en", {
+            useGrouping: "always",
+            sign: "exceptZero"
+          }).format(diff);
+
+          const diff_pct_str = Intl.NumberFormat("en", {
+            style: "percent",
+            useGrouping: "always",
+            sign: "exceptZero",
+            maximumFractionDigits: 2
+          }).format(diff_pct);
 
           if (diff !== 0) {
-            const percent = (((updated / reference) - 1) * 100).toFixed(2);
             // The body is created here and wrapped so "weirdly" to avoid whitespace at the start of the lines,
             // which is interpreted as a code block by Markdown.
             const body = `Below is the size of a hello-world Rust program linked with libstd with backtrace.
 
-          Original binary size: **${reference}B**
-          Updated binary size: **${updated}B**
-          Difference: **${diff_str}** (${percent}%)`;
+          Original binary size: **${reference_str} B**
+          Updated binary size: **${updated_str} B**
+          Difference: **${diff_str} B** (${diff_pct_str})`;
 
             github.rest.issues.createComment({
               issue_number: context.issue.number,
--- a/.github/actions/report-code-size-changes/action.yml
+++ b/.github/actions/report-code-size-changes/action.yml
@@ -1,0 +1,56 @@
+# Github composite action to report on code size changes
+name: Report binary size changes on PR
+description: |
+  Report on code size changes resulting from a PR as a comment on the PR
+  (accessed via context).
+inputs:
+   reference:
+    description: The size in bytes of the reference binary (base of PR).
+    required: true
+   updated:
+    description: The size in bytes of the updated binary (head of PR).
+    required: true
+runs:
+  using: composite
+  steps:
+    - name: Post a PR comment if the size has changed
+      uses: actions/github-script@v6
+      env:
+        SIZE_REFERENCE: ${{ inputs.reference }}
+        SIZE_UPDATED: ${{ inputs.updated }}
+      with:
+        script: |
+          const reference = process.env.SIZE_REFERENCE;
+          const updated = process.env.SIZE_UPDATED;
+
+          if (!(reference > 0)) {
+            core.setFailed(`Reference size invalid: ${reference}`);
+            return;
+          }
+
+          if (!(updated > 0)) {
+            core.setFailed(`Updated size invalid: ${updated}`);
+            return;
+          }
+
+          const diff = updated - reference;
+          const plus = diff > 0 ? "+" : "";
+          const diff_str = `${plus}${diff}B`;
+
+          if (diff !== 0) {
+            const percent = (((updated / reference) - 1) * 100).toFixed(2);
+            // The body is created here and wrapped so "weirdly" to avoid whitespace at the start of the lines,
+            // which is interpreted as a code block by Markdown.
+            const body = `Below is the size of a hello-world Rust program linked with libstd with backtrace.
+
+          Original binary size: **${reference}B**
+          Updated binary size: **${updated}B**
+          Difference: **${diff_str}** (${percent}%)`;
+
+            github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body
+            })
+          }
--- a/.github/workflows/check-binary-size.yml
+++ b/.github/workflows/check-binary-size.yml
@@ -94,44 +94,10 @@ jobs:
     permissions:
       pull-requests: write
     steps:
-      - name: Post a PR comment if the size has changed
-        uses: actions/github-script@v6
-        env:
-          SIZE_REFERENCE: ${{ needs.measure.outputs.binary-size-reference }}
-          SIZE_UPDATED: ${{ needs.measure.outputs.binary-size-updated }}
+      # Clone backtrace to access Github composite actions to report size.
+      - uses: actions/checkout@v3
+      # Run the size reporting action.
+      - uses: ./.github/actions/report-code-size-changes
         with:
-          script: |
-            const reference = process.env.SIZE_REFERENCE;
-            const updated = process.env.SIZE_UPDATED;
-
-            if (!(reference > 0)) {
-              core.setFailed(`Reference size invalid: ${reference}`);
-              return;
-            }
-
-            if (!(updated > 0)) {
-              core.setFailed(`Updated size invalid: ${updated}`);
-              return;
-            }
-
-            const diff = updated - reference;
-            const plus = diff > 0 ? "+" : "";
-            const diff_str = `${plus}${diff}B`;
-
-            if (diff !== 0) {
-              const percent = (((updated / reference) - 1) * 100).toFixed(2);
-              // The body is created here and wrapped so "weirdly" to avoid whitespace at the start of the lines,
-              // which is interpreted as a code block by Markdown.
-              const body = `Below is the size of a hello-world Rust program linked with libstd with backtrace.
-
-            Original binary size: **${reference}B**
-            Updated binary size: **${updated}B**
-            Difference: **${diff_str}** (${percent}%)`;
-
-              github.rest.issues.createComment({
-                issue_number: context.issue.number,
-                owner: context.repo.owner,
-                repo: context.repo.repo,
-                body
-              })
-            }
+          reference: ${{ needs.measure.outputs.binary-size-reference }}
+          updated: ${{ needs.measure.outputs.binary-size-updated }}
--- a/.github/workflows/check-binary-size.yml
+++ b/.github/workflows/check-binary-size.yml
@@ -9,13 +9,20 @@ on:
     branches:
       - master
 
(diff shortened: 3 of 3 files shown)


philipc:
> OTOH, if it's because of some transitive dependencies that are now mismatched

Yes it's because of this. You need to patch these too:
https://github.com/rust-lang/rust/blob/70d7283d24ed2d664fc4cefad6f240aeb62e98f2/library/std/Cargo.toml#L27-L35

and this should be done if you want to get accurate measurements of size changes, because these dependencies can cause size changes too.

Kobzol:
The `cp` commands should now be removed, they are performed by `rustc` bootstrap automatically.

Kobzol:
This is pretty great! One problem with this workflow is that of GH permissions regarding sending comments. We probably shouldn't use `pull_request_target` at all, since in theory it enables GHA cache poisoning. But without it, we can't easily send PR comments. The `pull_request_target` approach and building code from "foreign" PRs is not deemed to be safe though, so we should probably do something about it.

One way to solve this would be to have one workflow that calculates the size, and then another workflow that just reads the calculated outputs and posts it to the PR. I started drafting something in https://github.com/rust-lang/backtrace-rs/pull/548, but it's not finished. I realized that it's not even that trivial to even recognize if the original workflow came from a PR (and from which one). Maybe this should better be implemented in JavaScript, in a first-class GitHub action. Or it should just be a part of bors, I don't know.

detly:
> The `cp` commands should now be removed, they are performed by `rustc` bootstrap automatically.

I'll rebase this after #550, I just can't do it in parallel because I can only test workflows as they exist on my fork's `master` branch and it's already a bit of a juggle.

Re. the safety aspect - I am more familiar with Gitlab's CI than Github's, but they have the concept of "metrics" that have a similar kind of "first class" recognition to things like test and coverage reports. I say similar because they don't _actually_ have a good way to display them in MRs, they're more for post-deployment monitoring (_I think_), but it's in the works. My point is that if Github have some other first class way to report basic numeric metrics that don't involve posting a comment to the PR via the API, that could be an easy win. I'm trying to find that out.

For my own understanding:

> The `pull_request_target` approach and building code from "foreign" PRs is not deemed to be safe though

What's the potential issue? I thought that any PR triggered the workflow as written in the target.

.github/actions/build-with-patched-std/action.yml
```diff
@@ -0,0 +1,48 @@
...
+  test-binary-size:
+    description: The size in bytes of the built test binary
+    value: ${{ steps.measure.outputs.test-binary-size }}
+runs:
+  using: "composite"
+  steps:
+    - shell: bash
+      id: measure
+      env:
+        RUSTC_FLAGS: -Copt-level=s -Cstrip=symbols
```
workingjubilee, line 29: I do not want to use `-Copt-level=s` as the main check.
1. It's iffy! It sometimes results in larger sizes than 3, or even 2!
2. It is not representative of "default" `--release` builds, which are my primary concern.

workingjubilee: Example of how rustc sometimes is missing basic opts that would guarantee a smaller size for `-Copt-level=s`: https://github.com/rust-lang/rust/pull/100035

Kobzol:
> What's the potential issue? I thought that any PR triggered the workflow as written in the target.

There are (at least) two issues currently:
1) The part of the workflow that checks out the untrusted part currently has `pull-requests: write` permission, meaning that it could do whatever it wants with PRs, if the built code was somehow executed. This should be trivial to fix by splitting the current workflow into two jobs (one would measure the binary size, the other would post the comment) and giving no permissions to the first job, and PR permissions to the second job (you cannot change permissions per steps, only per jobs).
2) The `pull_request_target` workflow is prone to GHA cache poisoning attacks. I haven't found any way to prevent this, and GH considers this not to be an issue apparently, so it's probably marked as `wontfix`. This is mostly a theoretical attack vector, because (barring some really obscure bug in rustc), `backtrace` shouldn't have any way of executing code during its build, and this repository does not use GHA cache in any way currently (AFAIK). But I suppose that all attack vectors first start as theoretical...

Kobzol:
FWIW, fixing the first issue to make sure that no permissions are available to the job that could actually in theory run untrusted code, would be a good enough fix for now.

.github/actions/build-with-patched-std/action.yml
detly: I mean, one possibility is to do _both_ `-Os` and `-O` (default release), but then we're into matrix-style CI (I think?) and I'm not sure I'm up for doing that myself.

It's not a huge deal and I'm happy to change it back though. The code size varies between `-Os` and `-O3` _less_ than it does between architectures (IME). It's far more important to have good and well-defined sensitivity to things that affect code size, which as you say, `-Os` is probably not the best for.

detly:
> FWIW, fixing the first issue to make sure that no permissions are available to the job that could actually in theory run untrusted code, would be a good enough fix for now.

I could do that, the sizes are just numbers so they can be conveyed via env vars without even needing artifacts passed around.

.github/actions/build-with-patched-std/action.yml
workingjubilee: `-O` is **not** default `--release`, it is `-Copt-level=2`.

Kobzol: `rustc --help` claims this currently: `    -O                  Equivalent to -C opt-level=2`.

Kobzol: Oh, the default for `--release` is `-Copt-level=3`, I didn't realize. Maybe we should just create a new cargo project and build it with `--release` to have the most "realistic" situation?

workingjubilee: lmao am I the only one who knew?

yeah probably, if it's not too much additional effort.

detly: The case against that is that it's potentially a source of variability. Controlling the Rust flags directly rules out a sudden change in baseline because Cargo changed a default, or how it interprets a simplified option. (Granted, it would be a change in code size _between_ PRs, not in a single one, but it might still make accounting for changes difficult.)

Kobzol: I think that this CI workflow is now specifically only designed to ever care about changes *within* a single PR. The size between PRs will change a lot anyways because of unrelated compiler/libstd/... changes.

detly: Strictly correct, but you might find it a little useful to be able to compare between PRs _somewhat_, even if it's not super scientific. Involving Cargo might trade that off a little.

If you _do_ want to go the route of using a full-fledged Cargo project, would it then be better to store it as actual files in the repo in some designated directory, instead of conjuring it out of a bash script embedded in a workflow YAML file like a devops matryoshka doll?

workingjubilee: I don't have a strong opinion beyond "it should roughly match the flags used by typical `--release` builds".

detly:
> It could be done via [job outputs](https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs) though, that's quite similar.

Yep, this was easy enough. [Here's an example](https://github.com/detly/backtrace-rs/actions/runs/5582660130/jobs/10202179855?pr=4) of it posting a comment.

.github/actions/build-with-patched-std/action.yml
detly: @workingjubilee I am reverting the `opt-level` flag in my fork, but there is also the question of `strip`. Cargo's docs say [it defaults to `none`](https://doc.rust-lang.org/cargo/reference/profiles.html#strip), but I don't know if they mean "for the default profile of `debug`" or if that's true even for `release`. It's not listed in the [`release` profile settings](https://doc.rust-lang.org/cargo/reference/profiles.html#release).

detly:
I _also_ have a solution to the problem of (some) dependency bumps resulting in mismatched symbols. Would you like me to include it in this PR, or would you prefer to limit the scope here and have it in a different PR?

.github/actions/build-with-patched-std/action.yml
Kobzol: I would just do `cargo new --bin foo`, then overwrite `src/main.rs` and run `cargo +local build --release`. Before that we'd just need to `rustup toolchain link` the `build/<target>/stage0-sysroot` directory. At least it will also test that it can be linked with rustup :)

Kobzol:
I would keep it for a separate PR. Good job with the job split, now it should be safer.

detly:
One other very minor thing, but I am finding numbers like **391568B** kind of hard to read. Would anyone like me to format the number with, say, `Intl.NumberFormat({useGrouping: "always"}).format()` to add a thousands separator (comma), and add a non-breaking space between the number and unit?

.github/actions/build-with-patched-std/action.yml
```diff
@@ -0,0 +1,45 @@
...
+        RUSTC_BUILD_DIR: build/x86_64-unknown-linux-gnu
```
klensy, line 30: Only difference to allow this build on windows is that string?

.github/actions/build-with-patched-std/action.yml
detly: It's not clear to me what the consensus is here (except to put `opt-level` back to 3). Would it be better to stick with the current single-file approach and implement the cargo-project-approach in a new PR?

.github/actions/build-with-patched-std/action.yml
klensy: Yes, https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell

replace `runs-on: ubuntu-latest` with `windows-2019`.

.github/actions/build-with-patched-std/action.yml
Kobzol: Let's switch to `-Copt-level=3` here and do the cargo changes in a new PR, it will require more changes (as explained in my previous comment).

.github/actions/build-with-patched-std/action.yml
detly: Oh I saw that, but in order to test it locally, is there a published Docker image that relates to? AFAIK Github make their own modifications to underlying images to support their CI eg. `ubuntu-latest` is the official `ubuntu:latest` Docker image plus various GH/CI related changes. But I have not been able to find out whether these images are available to pull, or how to refer to them if so.

klensy: I don't think windows can run in docker, it's some azure stuff that runs windows images here. You can try to test it on your fork of backtrace, i think.

detly: Ah that makes so much more sense. Yes, I have a weird fork setup where I test one fake test branch against another. I'll give it a shot.

.github/actions/build-with-patched-std/action.yml
detly: Just rebased with the `opt-level` changed back to 3 _but_ still with the additional strip option. Let me know if I should drop that one too.

.github/actions/build-with-patched-std/action.yml
detly: Actually I think it can just be replaced with `build/host`.

detly: I got this working on Windows without a lot of changes at all, [see this CI run](https://github.com/detly/backtrace-rs/actions/runs/5595206423/jobs/10230858124). I'll port the changes over but leave the jobs as running on `ubuntu-latest` unless there's a preference for `windows-2019`.

workingjubilee:
> One other very minor thing, but I am finding numbers like **391568B** kind of hard to read. Would anyone like me to format the number with, say, `Intl.NumberFormat({useGrouping: "always"}).format()` to add a thousands separator (comma), and add a non-breaking space between the number and unit?

I'd like something that splits out KiB and B, personally.

detly:
> I'd like something that splits out KiB and B, personally.

I have been playing around with [`Intl.NumberFormat()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat) which _mostly_ does a decent job, but unfortunately abbreviates the unit of "byte" to... "byte". This is dictated by [the source spec](https://unicode.org/reports/tr35/tr35-general.html#Unit_Elements) they defer to. [Here's an example.](https://runjs.co/s/KAfm47O00)

For now I've used `Intl.NumberFormat()` for the numbers and just put ` B` in the strings.

.github/actions/build-with-patched-std/action.yml
klensy: I don't understand. If you leave only linux runner, how it will catch changes for windows? You can specify [matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) so it will run on windows-2019 and ubuntu-22.04. Mac can be skipped, i think.

detly: Fair point, especially since a lot of contribution to code size comes from platform specific code (eg. gimli and coff). But the downside to this would be that for changes that affect both, you'll get two comments per PR _per update_, which might be a bit spammy. (Plus, if you ever want to add a dimension to the matrix eg. compiler flags, features, you'll potentially get N×M comments per PR update.)

I had thought there would be a way to change the comment job to summarise it all, but AFAICT Github [don't have a way](https://github.com/orgs/community/discussions/17245) to operate on the combined/collected output from matrices. Even using outputs or artifacts seems like it's [subject to race conditions](https://github.com/orgs/community/discussions/17245#discussioncomment-6261191). It's designed for matrix elements to map to independent workflows that don't need the full set of results later.

klensy: I guess it's ok to have 2 comments instead of 1 for now, it can be tweaked later. Looks like you can edit comments, so potentially it can stack all of them for one commit into one, but that can be done later.

detly: I was wrong about the two comments thing. If other jobs have a `needs`-dependency on matrix jobs, they run only once, and get only the last output (with order being arbitrary). So accessing outputs is possible by uploading artifacts where collisions are avoided by having distinct (matrix-based) filenames. The comment step would then loop over these outputs (knowing that all are complete, because it gets fired on the last job).

I will see if this works and let you know.

detly: Not gonna lie, this is more javascript than I've written for a while. But. I wanted to make it not-horrible to add dimensions or values to the matrix in the future. Basically, the size-measuring-jobs write a file each to an artifact store, and the reporting job reads them all. If there are any changes, it posts a comment that looks like:

---

Code size changes for a hello-world Rust program linked with libstd with backtrace:

On platform `ubuntu-latest`:

- Original binary size: **399,760 B**
- Updated binary size: **379,280 B**
- Difference: **-20,480 B** (-5.12%)

On platform `windows-latest`:

- Original binary size: **160,256 B**
- Updated binary size: **151,552 B**
- Difference: **-8,704 B** (-5.43%)

detly:
(I'm happy to rebase and squash up commits when you're ready.)

detly:
This has languished a bit, and I left it alone because I was busy, but if it's still of interest let me know. I should have time now to address anything that's still outstanding.

workingjubilee:
Sorry!

https://www.youtube.com/watch?v=fksu6FENojY

This looks good, I can approve it as soon as you're happy with it re: commit squashing or w/e.

detly:
> https://www.youtube.com/watch?v=fksu6FENojY

Oh I hear that. No pressure. I won't be doing anything faster than can be measured in days either. I will tidy up the history and ping you here when that's done. If others have feedback on the details too I'm happy to hear it in the meantime.

detly:
Okay I'm much happier with this. If you want to further squash the commits, I don't mind, but the history as it is now should be clear and valid taken commit by commit too.

detly:
Docker was already mad, right?

workingjubilee:
That failure is new, I think.

workingjubilee (approved):
Weird and annoying and I'm going to ignore it and merge this.

Issue #535, opened 2023-06-28
Windows: Consider alternatives to dbghelp.dll


ChrisDenton:
The advantage of dbghelp.dll is that it can nearly always be found on any Windows system. However, the major disadvantage is it's [ridiculously single threaded](https://learn.microsoft.com/en-us/windows/win32/debug/about-dbghelp):

> Note that all [DbgHelp functions](https://learn.microsoft.com/en-us/windows/win32/debug/dbghelp-functions) are single threaded. Therefore, calls from more than one thread to this function will likely result in unexpected behavior or memory corruption. To avoid this, you must synchronize all concurrent calls from more than one thread to this function.

We have to implement synchronization across whatever random dlls may be thrown together and hope to hell nothing else is using it in the same process. It also may be old, The PDB debug format can and does change over time.

A stack backtrace can be retrieved without dbghelp using either [`CaptureStackBackTrace`](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/nf-ntifs-rtlcapturestackbacktrace) or [`RtlVirtualUnwind`](https://learn.microsoft.com/en-us/windows/win32/api/winnt/nf-winnt-rtlvirtualunwind).

Reading PDB files is more complex but can be done using the [Debug Interface Access (DIA) SDK](https://learn.microsoft.com/en-us/visualstudio/debugger/debug-interface-access/debug-interface-access-sdk?view=vs-2022). The downside is this requires an msdia*.dll but at least it's redistributable. The upside is that it has a proper API that doesn't break in the presence of threads.

---

Note: I don't personally have a ton of time atm to follow up on this but I'm posting it here in case someone else wants to. Maybe that someone will be future!me, or maybe not. Also it's fine to conclude that dbghelp is our best option but I'm kinda hoping it isn't.

ChrisDenton:
As I noted in the experiment linked above, decoupling tracing from symbol loading can be done even without deciding on dia vs. dbghelp. Not needing to load an external dll is great if you just need to store a backtrace for later. Also doing this would mean the `windows-gnu` targets don't need dbghelp or dia at all since they use DWARF.

If anyone wants to try this then I think the easiest thing would be to try using `RtlVirtualUnwind` on `x86_64` only initially.

PR #552, opened 2023-07-17
[DO NOT MERGE] Experiment with `RtlCaptureStackBackTrace`

ChrisDenton:
Experiment with `RtlCaptureStackBackTrace` as per #535. This is not intended to be merged. I'm leaving this here temporarily in case anyone wants to use it to experiment further.

The great thing about `RtlCaptureStackBackTrace` is that it doesn't require symbol loading and simply works across architectures without much fuss. This decouples symbol loading from tracing and allows a more lightweight way of storing backtrace information if you don't mind looking up the symbols later. The big downside is that it only gets the `ip`. Also I'm seeing off-by-one line numbers in `tests\accuracy`. Probably because of the more limited info.

`RtlVirtualUnwind` has the same advantages as `RtlCaptureStackBackTrace` and also provide all the information you need (symbols aside). However, it is much more involved and it isn't supported by x86. Still, it might be a better direction to go in.

### DIFF
--- a/src/backtrace/dbghelp.rs
+++ b/src/backtrace/dbghelp.rs
@@ -11,14 +11,12 @@
 
 #![allow(bad_style)]
 
-use super::super::{dbghelp, windows::*};
+use super::super::windows::*;
 use core::ffi::c_void;
-use core::mem;
 
 #[derive(Clone, Copy)]
-pub enum StackFrame {
-    New(STACKFRAME_EX),
-    Old(STACKFRAME64),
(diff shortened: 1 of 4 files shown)


ChrisDenton:
(I'm not really working on this, I just had a fit of insomnia last night and this is the result)

ChrisDenton:
Closing this now as I'm not planning on pursuing it further at this time. But I'll leave the branch in place in case anyone else does.

PR #553, opened 2023-07-24
Bug fix: correctly handle spaces in /proc/pid/maps pathnames

MasonRemaley:
Prior to this change, pathnames were parsed from `/proc/pid/maps` using `s.split(' ')`. This is incorrect, because pathnames may contain spaces.

In practice, on the relevant operating systems, this resulted in truncation of pathnames containing spaces. This lead to a failure resolve any symbols or offsets, which in turn causes the stack trace to be printed containing only `<unknown>`s to the end user.

This PR corrects the parsing code, resolving the issue.

### DIFF
--- a/src/symbolize/gimli/parse_running_mmaps_unix.rs
+++ b/src/symbolize/gimli/parse_running_mmaps_unix.rs
@@ -115,7 +115,7 @@ impl FromStr for MapsEntry {
         }
 
         // Pathname may be omitted in which case it will be empty
-        let pathname_str = s.trim();
+        let pathname_str = s.trim_start();
 
         let hex = |s| usize::from_str_radix(s, 16).map_err(|_| "Couldn't parse hex number");
         let address = if let Some((start, limit)) = range_str.split_once('-') {
@@ -275,7 +275,7 @@ fn check_maps_entry_parsing_32bit() {
             offset: 0x00000000,
             dev: (0x08, 0x01),
             inode: 0x60662705,
-            pathname: "/executable/path/with  multiple-continues    spaces".into(),
+            pathname: "/executable/path/with  multiple-continues    spaces  ".into(),
         }
     );
     assert_eq!(
--- a/src/symbolize/gimli/parse_running_mmaps_unix.rs
+++ b/src/symbolize/gimli/parse_running_mmaps_unix.rs
@@ -95,22 +95,22 @@ impl FromStr for MapsEntry {
         }
 
         let (perms_str, s) = s.trim().split_once(' ').unwrap_or((s, ""));
-        if range_str.is_empty() {
+        if perms_str.is_empty() {
             return Err("Couldn't find permissions");
         }
 
         let (offset_str, s) = s.trim().split_once(' ').unwrap_or((s, ""));
-        if range_str.is_empty() {
+        if offset_str.is_empty() {
             return Err("Couldn't find offset");
         }
 
         let (dev_str, s) = s.trim().split_once(' ').unwrap_or((s, ""));
-        if range_str.is_empty() {
+        if dev_str.is_empty() {
             return Err("Couldn't find dev");
         }
 
         let (inode_str, s) = s.trim().split_once(' ').unwrap_or((s, ""));
-        if range_str.is_empty() {
+        if inode_str.is_empty() {
             return Err("Couldn't find inode");
         }
 
--- a/src/symbolize/gimli/parse_running_mmaps_unix.rs
+++ b/src/symbolize/gimli/parse_running_mmaps_unix.rs
@@ -264,4 +264,18 @@ fn check_maps_entry_parsing_32bit() {
             pathname: "/executable/path/with some spaces".into(),
         }
     );
+    assert_eq!(
+        "b7c79000-b7e02000 r--p 00000000 08:01 60662705   \
+                /executable/path/with  multiple-continues    spaces  "
+            .parse::<MapsEntry>()
+            .unwrap(),
+        MapsEntry {
+            address: (0xb7c79000, 0xb7e02000),
+            perms: ['r', '-', '-', 'p'],
+            offset: 0x00000000,
+            dev: (0x08, 0x01),
+            inode: 0x60662705,
+            pathname: "/executable/path/with  multiple-continues    spaces".into(),
+        }
+    );
 }
--- a/src/symbolize/gimli/parse_running_mmaps_unix.rs
+++ b/src/symbolize/gimli/parse_running_mmaps_unix.rs
@@ -89,27 +89,27 @@ impl FromStr for MapsEntry {
     // Note that paths may contain spaces, so we can't use `str::split` for parsing (until
     // Split::remainder is stabalized #77998).
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        let (range_str, s) = s.trim().split_once(' ').unwrap_or((s, ""));
+        let (range_str, s) = s.trim_start().split_once(' ').unwrap_or((s, ""));
         if range_str.is_empty() {
             return Err("Couldn't find address");
         }
 
-        let (perms_str, s) = s.trim().split_once(' ').unwrap_or((s, ""));
+        let (perms_str, s) = s.trim_start().split_once(' ').unwrap_or((s, ""));
         if perms_str.is_empty() {
             return Err("Couldn't find permissions");
         }
 
-        let (offset_str, s) = s.trim().split_once(' ').unwrap_or((s, ""));
+        let (offset_str, s) = s.trim_start().split_once(' ').unwrap_or((s, ""));
         if offset_str.is_empty() {
             return Err("Couldn't find offset");
         }
 
-        let (dev_str, s) = s.trim().split_once(' ').unwrap_or((s, ""));
+        let (dev_str, s) = s.trim_start().split_once(' ').unwrap_or((s, ""));
         if dev_str.is_empty() {
             return Err("Couldn't find dev");
         }
 
-        let (inode_str, s) = s.trim().split_once(' ').unwrap_or((s, ""));
+        let (inode_str, s) = s.trim_start().split_once(' ').unwrap_or((s, ""));
         if inode_str.is_empty() {
             return Err("Couldn't find inode");
         }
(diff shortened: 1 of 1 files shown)


MasonRemaley:
Side note--I tried to make the least invasive change possible since I'm not familiar with the codebase. That being said, is parsing `/proc/pid/maps` actually necessary?

If I'm reading the code correctly it's parsing the whole thing just to get the path to the main program. I would think there are simpler ways to get that info, but it's possible I'm missing something!

bjorn3:
`/proc/self/exe` doesn't help with dynamically linked libraries.

philipc:
There's an edge case where explicitly running ld.so gives the wrong path; see #488

src/symbolize/gimli/parse_running_mmaps_unix.rs
```diff
@@ -85,16 +85,37 @@ impl FromStr for MapsEntry {
     // e.g.: "ffffffffff600000-ffffffffff601000 --xp 00000000 00:00 0                  [vsyscall]"
     // e.g.: "7f5985f46000-7f5985f48000 rw-p 00039000 103:06 76021795                  /usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2"
     // e.g.: "35b1a21000-35b1a22000 rw-p 00000000 00:00 0"
+    //
+    // Note that paths may contain spaces, so we can't use `str::split` for parsing (until
+    // Split::remainder is stabalized #77998).
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        let mut parts = s
-            .split(' ') // space-separated fields
-            .filter(|s| s.len() > 0); // multiple spaces implies empty strings that need to be skipped.
```
klensy, line 91: Now this missed that check: multiple continuous empty spaces (but there now test for it)

src/symbolize/gimli/parse_running_mmaps_unix.rs
```diff
@@ -85,16 +85,37 @@ impl FromStr for MapsEntry {
...
-        let dev_str = parts.next().ok_or("Couldn't find dev")?;
-        let inode_str = parts.next().ok_or("Couldn't find inode")?;
-        let pathname_str = parts.next().unwrap_or(""); // pathname may be omitted.
+        let (range_str, s) = s.trim().split_once(' ').unwrap_or((s, ""));
+        if range_str.is_empty() {
+            return Err("Couldn't find address");
+        }
+
+        let (perms_str, s) = s.trim().split_once(' ').unwrap_or((s, ""));
+        if range_str.is_empty() {
```
klensy, line 98: copy&paste: your checking range_str all down here, while should check perms_str and others.

src/symbolize/gimli/parse_running_mmaps_unix.rs
```diff
@@ -85,16 +85,37 @@ impl FromStr for MapsEntry {
...
-        let mut parts = s
-            .split(' ') // space-separated fields
-            .filter(|s| s.len() > 0); // multiple spaces implies empty strings that need to be skipped.
-        let range_str = parts.next().ok_or("Couldn't find address")?;
-        let perms_str = parts.next().ok_or("Couldn't find permissions")?;
-        let offset_str = parts.next().ok_or("Couldn't find offset")?;
-        let dev_str = parts.next().ok_or("Couldn't find dev")?;
-        let inode_str = parts.next().ok_or("Couldn't find inode")?;
-        let pathname_str = parts.next().unwrap_or(""); // pathname may be omitted.
+        let (range_str, s) = s.trim().split_once(' ').unwrap_or((s, ""));
```
klensy, line 92: No need to trim until you reached pathname, yes?

bjorn3: Without the trim multiple spaces in a row would be considered individual split targets rather than a single one, right? So for example `a<space><space>b` would split as `["a", "", "b"]`.

klensy: From new test, but appended trailing spaces to path `"b7c79000-b7e02000 r--p 00000000 08:01 60662705   /executable/path/with some spaces       "`, this will be broken.

src/symbolize/gimli/parse_running_mmaps_unix.rs
MasonRemaley: Whoops, thanks! Fixed.

src/symbolize/gimli/parse_running_mmaps_unix.rs
MasonRemaley: Added a test to verify that multiple continuous spaces work

src/symbolize/gimli/parse_running_mmaps_unix.rs
```diff
@@ -264,4 +264,18 @@ fn check_maps_entry_parsing_32bit() {
...
+                /executable/path/with  multiple-continues    spaces  "
+            .parse::<MapsEntry>()
+            .unwrap(),
+        MapsEntry {
+            address: (0xb7c79000, 0xb7e02000),
+            perms: ['r', '-', '-', 'p'],
+            offset: 0x00000000,
+            dev: (0x08, 0x01),
+            inode: 0x60662705,
+            pathname: "/executable/path/with  multiple-continues    spaces".into(),
```
klensy, line 278: Wrong pathname, it should contain trailing spaces

src/symbolize/gimli/parse_running_mmaps_unix.rs
MasonRemaley: Replaced `trim` with `trim_start` where appropriate!

MasonRemaley:
> There's an edge case where explicitly running ld.so gives the wrong path; see #488

Wasn't aware of this, thanks for the link!

src/symbolize/gimli/parse_running_mmaps_unix.rs
MasonRemaley: TIL paths can have trailing spaces--thanks, fixed!

klensy:
Looking at https://github.com/torvalds/linux/blob/18b44bc5a67275641fb26f2c54ba7eef80ac5950/fs/proc/task_mmu.c#L254 i think that multiple spaces isn't possible anywhere except around pathname?

MasonRemaley:
> Looking at https://github.com/torvalds/linux/blob/18b44bc5a67275641fb26f2c54ba7eef80ac5950/fs/proc/task_mmu.c#L254 i think that multiple spaces isn't possible anywhere except around pathname?

I'm being intentionally forgiving--

If my game crashes most players will just refund it, but assuming they consent I'll at least get a stack trace so I can fix it for the next person. If some day Linux decides to start aligning other elements by column too, I don't want to end up with a nightmare where I have to drop everything to figure out why my stack traces are failing before missing too much debug info.

src/symbolize/gimli/parse_running_mmaps_unix.rs
```diff
@@ -229,4 +250,46 @@ fn check_maps_entry_parsing_32bit() {
...
+            perms: ['r', '-', '-', 'p'],
+            offset: 0x00000000,
+            dev: (0x08, 0x01),
+            inode: 0x60662705,
+            pathname: "/executable/path/with some spaces".into(),
+        }
+    );
+    assert_eq!(
+        "b7c79000-b7e02000 r--p 00000000 08:01 60662705   \
+                /executable/path/with  multiple-continues    spaces  "
```
workingjubilee, line 269:
```suggestion
                /executable/path/with  multiple-continuous    spaces  "
```

This makes me think of a continue like
CONTINUE? 9...

src/symbolize/gimli/parse_running_mmaps_unix.rs
```diff
@@ -229,4 +250,46 @@ fn check_maps_entry_parsing_32bit() {
...
+            .parse::<MapsEntry>()
+            .unwrap(),
+        MapsEntry {
+            address: (0xb7c79000, 0xb7e02000),
+            perms: ['r', '-', '-', 'p'],
+            offset: 0x00000000,
+            dev: (0x08, 0x01),
+            inode: 0x60662705,
+            pathname: "/executable/path/with  multiple-continues    spaces  ".into(),
```
workingjubilee, line 278:
```suggestion
            pathname: "/executable/path/with  multiple-continuous    spaces  ".into(),
```
Multiple continues? Why not 1cc it? :pensive:

src/symbolize/gimli/parse_running_mmaps_unix.rs
```diff
@@ -85,16 +85,37 @@ impl FromStr for MapsEntry {
     // e.g.: "ffffffffff600000-ffffffffff601000 --xp 00000000 00:00 0                  [vsyscall]"
     // e.g.: "7f5985f46000-7f5985f48000 rw-p 00039000 103:06 76021795                  /usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2"
     // e.g.: "35b1a21000-35b1a22000 rw-p 00000000 00:00 0"
+    //
+    // Note that paths may contain spaces, so we can't use `str::split` for parsing (until
+    // Split::remainder is stabalized #77998).
```
workingjubilee, line 90:
```suggestion
    // Split::remainder is stabilized #77998).
```

workingjubilee (commented):
Oh hey, thank you so much!

1. Yes, as discussed, it unfortunately _is_ necessary to parse `proc/${pid}/maps` due to the "fun" edge-cases if we want this backtrace to be consistently correct. It's acceptable to throw away more data than we currently throw away and adjust strictness, but not to avoid parsing, as avoiding parsing opens up far more problems than it solves. Case in point: the parse wasn't even right.
2. Unless we find some other way around the edge-cases (and I'm not confident ld.so is the only one!). I won't ask you to waste your time on that, however.
3. Yes, I agree that we don't need to require strict consistency in spacing with the existing format. There's no need to be too brittle, and even if the rendering of `/proc/*/maps` changes, libstd will be expected to have its backtrace work consistently across all kernel versions.

This PR looks great overall! I just see a few silly nits.

MasonRemaley:
> Oh hey, thank you so much!

You're welcome! :)

> 1. Yes, as discussed, it unfortunately _is_ necessary to parse `proc/${pid}/maps` due to the "fun" edge-cases if we want this backtrace to be consistently correct. It's acceptable to throw away more data than we currently throw away and adjust strictness, but not to avoid parsing, as avoiding parsing opens up far more problems than it solves. Case in point: the parse wasn't even right.
> 
> 2. Unless we find some other way around the edge-cases (and I'm not confident ld.so is the only one!). I won't ask you to waste your time on that, however.

That makes sense!

> This PR looks great overall! I just see a few silly nits.

Thanks! Merged your suggestions--apologies for the typos!

src/symbolize/gimli/parse_running_mmaps_unix.rs
MasonRemaley: Gotta start somewhere :stuck_out_tongue_winking_eye:

MasonRemaley:
Hey! Just checking in to see if you need any other changes for this to be mergeable. :)

workingjubilee (approved):
Mostly for my brain to get back from outer space!

MasonRemaley:
Haha no worries, glad to get the fix in!

PR #559, opened 2023-08-15
Use extended symbolization markup on Fuchsia

liudangyi:
{{{reset:begin/end}}} pair will allow symbolization to be performed on the whole stack rather than invidival frames.

This has been tested manually and verified to work.

### DIFF
--- a/src/print.rs
+++ b/src/print.rs
@@ -83,7 +83,8 @@ impl<'a, 'b> BacktraceFmt<'a, 'b> {
     /// This is currently a no-op but is added for future compatibility with
     /// backtrace formats.
     pub fn finish(&mut self) -> fmt::Result {
-        // Currently a no-op-- including this hook to allow for future additions.
+        #[cfg(target_os = "fuchsia")]
+        fuchsia::finish_context(self.fmt)?;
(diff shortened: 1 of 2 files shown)


src/print.rs
```diff
@@ -83,7 +83,8 @@ impl<'a, 'b> BacktraceFmt<'a, 'b> {
     /// This is currently a no-op but is added for future compatibility with
     /// backtrace formats.
     pub fn finish(&mut self) -> fmt::Result {
-        // Currently a no-op-- including this hook to allow for future additions.
+        #[cfg(target_os = "fuchsia")]
+        self.fmt.write_str("{{{reset:end}}}\n")?;
```
tmandry, line 87: Can you put this in a method in the fuchsia module, so the "begin" markup printing lives in the same file as the "end"?

liudangyi: Done.

workingjubilee (approved):
Thanks!

I'm basically going to trust that this is correct since the tests pass. If there's a test you can write to prevent this behavior from regressing back to visualization on individual stack frames for this platform, please add it in a followup PR.

PR #566, opened 2023-09-21
Adjust frame IP in SGX relative to image base

mzohreva:
The backtraces printed by `panic!` are broken in SGX, this PR is a prerequisite for fixing the issue in the standard library.

The SGX-only API added in this PR (`backtrace::set_image_base`) enables the standard library to set the correct image base address before calling `trace_unsynchronized` here: https://github.com/rust-lang/rust/blob/1.72.1/library/std/src/sys_common/backtrace.rs#L65.

cc @jethrogb

### DIFF
--- a/src/backtrace/mod.rs
+++ b/src/backtrace/mod.rs
@@ -125,6 +125,37 @@ impl fmt::Debug for Frame {
     }
 }
 
+#[cfg(all(target_env = "sgx", target_vendor = "fortanix", not(feature = "std")))]
+mod sgx_no_std_image_base {
+    use core::sync::atomic::{AtomicU64, Ordering::SeqCst};
+
+    static IMAGE_BASE: AtomicU64 = AtomicU64::new(0);
+
+    /// Set the image base address. This is only available for Fortanix SGX
+    /// target when the `std` feature is not enabled. This can be used in the
+    /// standard library to set the correct base address.
+    pub fn set_image_base(base_addr: u64) {
+        IMAGE_BASE.store(base_addr, SeqCst);
+    }
+
+    pub(crate) fn get_image_base() -> u64 {
+        IMAGE_BASE.load(SeqCst)
+    }
+}
+
+#[cfg(all(target_env = "sgx", target_vendor = "fortanix", not(feature = "std")))]
+pub use self::sgx_no_std_image_base::set_image_base;
+
+#[cfg(all(target_env = "sgx", target_vendor = "fortanix", not(feature = "std")))]
+#[deny(unused)]
+pub(crate) use self::sgx_no_std_image_base::get_image_base;
+
+#[cfg(all(target_env = "sgx", target_vendor = "fortanix", feature = "std"))]
+#[deny(unused)]
+pub(crate) fn get_image_base() -> u64 {
+    std::os::fortanix_sgx::mem::image_base()
+}
+
 cfg_if::cfg_if! {
     // This needs to come first, to ensure that
     // Miri takes priority over the host platform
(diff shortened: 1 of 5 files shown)


workingjubilee:
Is there no way to automatically recover the base image address in this scenario?

mzohreva:
> Is there no way to automatically recover the base image address in this scenario?

No, due to ASLR the base address changes every time the enclave is loaded into memory.

workingjubilee:
Then how does std know? Apologies if this is an obvious/silly question, I am not familiar with the particulars of the SGX target, as opposed to my familiarity with x86.

mzohreva:
> Then how does std know? Apologies if this is an obvious/silly question, I am not familiar with the particulars of the SGX target, as opposed to my familiarity with x86.

The relevant code is here: https://github.com/rust-lang/rust/blob/1.72.1/library/std/src/sys/sgx/abi/mem.rs#L37 it's using inline assembly (that's why I don't think it's a good idea to duplicate that code in this crate or elsewhere)

workingjubilee:
I see.

> // Do not remove inline: will result in relocation failure

That... seems like a compiler bug? Is there a bug tracking this concern?

Can you mark the exposed APIs `#[doc(hidden)]`? I want to retain plausible deniability about whether this is a sanctioned usage of this crate, here.

src/backtrace/mod.rs
```diff
@@ -125,6 +125,37 @@ impl fmt::Debug for Frame {
...
+    /// Set the image base address. This is only available for Fortanix SGX
+    /// target when the `std` feature is not enabled. This can be used in the
+    /// standard library to set the correct base address.
+    pub fn set_image_base(base_addr: u64) {
+        IMAGE_BASE.store(base_addr, SeqCst);
+    }
+
+    pub(crate) fn get_image_base() -> u64 {
+        IMAGE_BASE.load(SeqCst)
+    }
```
workingjubilee, line 143: shouldn't this be usize?

mzohreva: it can be changed as such, but [the one in std](https://github.com/rust-lang/rust/blob/1.72.1/library/std/src/sys/sgx/abi/mem.rs#L37) returns `u64`

mzohreva:
> I see.
> 
> > // Do not remove inline: will result in relocation failure
> 
> That... seems like a compiler bug? Is there a bug tracking this concern?

I'm not aware. @jethrogb might know.

> Can you mark the exposed APIs `#[doc(hidden)]`? I want to retain plausible deniability about whether this is a sanctioned usage of this crate, here.

Sure, I'll update the PR

workingjubilee:
With that, happy to merge this once the question about the relocation thing is settled. Maybe it's a bug but unfixable? Either way, I'd like to know so I'm sure there's no loose ends, or er... more like the loose ends are properly loosened? I don't know.

jethrogb:
SGX doesn't have a dynamic loader, so the Rust entry code needs to do the relocations manually. Any code that's used to compute the relocations (which includes `fn image_base()`) shouldn't itself contain relocations.

mzohreva:
@workingjubilee the CI failures seem unrelated to the changes, is there a separate effort to address those? How should we proceed with this PR?

src/backtrace/mod.rs
workingjubilee: I don't quite understand that either, because the types appropriate for addresses are `usize`, `*const T`, and `*mut T`.

jethrogb: The target only supports x86_64

workingjubilee (approved):
Thank you!

PR #570, opened 2023-10-13
Allow the library consumer to specify the Android API level

danakj:
The build.rs script will use the C preprocessor to pull the API level out of C headers by default. But for build systems
that wish to control each build step, and rely on remote compilation, this presents problems.

If `--cfg=no_cc` is specified, the the C compiler will not be used. Instead the `__ANDROID_API__` environment variable
is used to specify the API level.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -29,7 +29,10 @@ BACKTRACE_RS_ANDROID_APIVERSION __ANDROID_API__
     let expansion = match cc::Build::new().file(&android_api_c).try_expand() {
         Ok(result) => result,
         Err(e) => {
-            eprintln!("warning: android version detection failed while running C compiler: {}", e);
+            eprintln!(
+                "warning: android version detection failed while running C compiler: {}",
+                e
+            );
             return None;
         }
     };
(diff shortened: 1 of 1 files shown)


workingjubilee:
@danakj Sorry for not getting around to this for a bit. I'm feeling a bit thick and now rust-lang/rust is using cc-rs 1.0.90, so where are we in wanting this?

danakj:
> @danakj Sorry for not getting around to this for a bit. I'm feeling a bit thick and now rust-lang/rust is using cc-rs 1.0.90, so where are we in wanting this?

No problem. In https://github.com/rust-lang/rust/pull/116318#issuecomment-1762186566 you noted that the NDK R26 supports API L21 as its minimum, at which point the build-time check here can go away entirely, and we just always do the `dl_iterate_phdr` thing.

We saw that Rust was still on NDK R25, and it [looks to still be](https://github.com/rust-lang/rust/blob/master/src/ci/docker/host-x86_64/arm-android/Dockerfile#L9).

So if we want to leave Rust on R25 but use the `backtrace-rs` build script in `std`, then we would still want to put `cc` behind a `no_cc` cfg flag. I can update this PR as it has merge conflicts if you think that's the right direction to go still.

workingjubilee:
@danakj It looks like https://github.com/rust-lang/rust/pull/120593 will advance, and when that does, you can simply make this the PR that removes the build check.

danakj:
Wow it finally merged! I am about to go on vacation though, unlucky. If there's no rush I will address this after I am back.

workingjubilee:
yeah I am surprised it took this long lmao.

fengys1996:
> Wow it finally merged! I am about to go on vacation though, unlucky. If there's no rush I will address this after I am back. (Feel free to steal the work if someone would like to.)

I'm interested in this and want to try it.  I have submitted a new PR.  https://github.com/rust-lang/backtrace-rs/pull/656 PTAL if u have time. @workingjubilee

workingjubilee:
656 was merged, so!

PR #571, opened 2023-10-23
Fix i686-pc-windows-gnu missing dbghelp module

wesleywiser:
`i686-pc-windows-gnu` also needs to use the `dbghelp` module so it should be declared when building for that target. 

Discovered when attempting to update the backtrace module in rust-lang/rust.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -192,7 +192,11 @@ mod lock {
     }
 }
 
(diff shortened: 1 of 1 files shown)


ChrisDenton (approved):

wesleywiser:
Thanks @ChrisDenton!

Issue #572, opened 2023-11-03
Build errors on a Windows ARM32 target since c1464fa


kleisauke:
Since commit c1464fadd7c870c87818b86b6161ee188be6f3a4, I'm getting the following build errors on a [custom `armv7-pc-windows-gnullvm` target](https://github.com/libvips/build-win64-mxe/blob/master/build/plugins/llvm-mingw/patches/rust-1-fixes.patch):

<details>
  <summary>Details</summary>

```
error[E0609]: no field `Eip` on type `backtrace_rs::windows::CONTEXT`
   --> /data/mxe/usr/x86_64-pc-linux-gnu/lib/rustlib/src/rust/library/std/src/../../backtrace/src/backtrace/dbghelp.rs:170:54
    |
170 |             stack_frame_ex.AddrPC.Offset = context.0.Eip as u64;
    |                                                      ^^^ unknown field
    |
    = note: available fields are: `ContextFlags`, `R0`, `R1`, `R2`, `R3` ... and 21 others

error[E0609]: no field `Esp` on type `backtrace_rs::windows::CONTEXT`
   --> /data/mxe/usr/x86_64-pc-linux-gnu/lib/rustlib/src/rust/library/std/src/../../backtrace/src/backtrace/dbghelp.rs:172:57
    |
172 |             stack_frame_ex.AddrStack.Offset = context.0.Esp as u64;
    |                                                         ^^^ unknown field
    |
    = note: available fields are: `ContextFlags`, `R0`, `R1`, `R2`, `R3` ... and 21 others

error[E0609]: no field `Ebp` on type `backtrace_rs::windows::CONTEXT`
   --> /data/mxe/usr/x86_64-pc-linux-gnu/lib/rustlib/src/rust/library/std/src/../../backtrace/src/backtrace/dbghelp.rs:174:57
    |
174 |             stack_frame_ex.AddrFrame.Offset = context.0.Ebp as u64;
    |                                                         ^^^ unknown field
    |
    = note: available fields are: `ContextFlags`, `R0`, `R1`, `R2`, `R3` ... and 21 others

error[E0609]: no field `Eip` on type `backtrace_rs::windows::CONTEXT`
   --> /data/mxe/usr/x86_64-pc-linux-gnu/lib/rustlib/src/rust/library/std/src/../../backtrace/src/backtrace/dbghelp.rs:208:53
    |
208 |             stack_frame64.AddrPC.Offset = context.0.Eip as u64;
[...]

PR #573, opened 2023-11-04
Fix build errors on `thumbv7a-*-windows-msvc` targets

kleisauke:
Resolves: #572.

The arm logic was copied from:
https://github.com/rust-lang/backtrace-rs/blob/036d4909e1fb9c08c2bb0f59ac81994e39489b2f/src/backtrace/dbghelp.rs#L248-L255


src/lib.rs
```diff
@@ -194,7 +194,7 @@ mod lock {
 
 #[cfg(all(
     windows,
-    any(target_env = "msvc", all(target_env = "gnu", target_arch = "x86")),
+    any(target_env = "msvc", all(target_env = "gnu", any(target_arch = "x86", target_arch = "arm"))),
```
kleisauke, line 197: If needed, I can omit this part from the PR since `armv7-pc-windows-gnullvm` is not (yet) an officially supported Rust target.

For reference, this was necessary to fix this build error:
```
error[E0433]: failed to resolve: could not find `dbghelp` in `super`
   --> /data/mxe/usr/x86_64-pc-linux-gnu/lib/rustlib/src/rust/library/std/src/../../backtrace/src/backtrace/dbghelp.rs:149:39
    |
149 |     let dbghelp = match super::super::dbghelp::init() {
    |                                       ^^^^^^^ could not find `dbghelp` in `super`
```

src/backtrace/dbghelp.rs
```diff
@@ -167,11 +167,21 @@ pub unsafe fn trace(cb: &mut dyn FnMut(&super::Frame) -> bool) {
...
+                        stack_frame_ex.AddrPC.Offset = context.0.Eip as u64;
+                        stack_frame_ex.AddrStack.Offset = context.0.Esp as u64;
+                        stack_frame_ex.AddrFrame.Offset = context.0.Ebp as u64;
+                } else {
+                        stack_frame_ex.AddrPC.Offset = context.0.Pc as u64;
+                        stack_frame_ex.AddrStack.Offset = context.0.Sp as u64;
+                        stack_frame_ex.AddrFrame.Offset = context.0.R11 as u64;
+                }
+            }
+
```
ChrisDenton, line 182: Rather than repeating this `cfg_if` block, it'd be better to add a platform specific implementations on `MyContext`, similar to https://github.com/rust-lang/backtrace-rs/blob/e9da96eb452aa65e79e2342be700544afe509440/src/backtrace/dbghelp.rs#L76-L81.

src/lib.rs
ChrisDenton: Seems fine. I'm not seeing it cause any issues yet. I'll keep it in mind though if any issue turns up after this is merged.

ChrisDenton (approved):
Looks great, thanks!

PR #578, opened 2023-11-21
Fix panic in backtrace symbolication on win7

### LINKED ISSUE #577: Panic on windows7 when resolving an inline
Since https://github.com/rust-lang/backtrace-rs/pull/569 was merged, `Backtrace::resolve` panics on Windows 7 due to unwrapping an `Option` set to None. After some further instrumentation, I was able to find out the source of the problem: `Failed to find SymAddrIncludeInlineTrace in dbghelp`. This makes the `capture::tests::test_frame_conversion` unit test fail on Windows 7.

This is because Windows7 ships dbghelp.dll version 6.1, but `SymAddrIncludeInlineTrace` was added in dbghelp.dll version 6.2.

This affects the Rust stdlib Backtrace as well, see https://github.com/rust-lang/rust/issues/117941

roblabla:
Since https://github.com/rust-lang/backtrace-rs/pull/569 was merged, symbolication of backtraces would panic on Windows 7, due to using symbols that do not exist in the version of dbghelp.dll that ships there (Windows 7 ships with dbghelp.dll version 6.1, but the symbols were only added in version 6.2).

This commit checks for the presence of the newer symbols, and if they are not found, falls back to the old resolution method.

Fixes #577

### DIFF
--- a/src/dbghelp.rs
+++ b/src/dbghelp.rs
@@ -34,8 +34,8 @@ use core::ptr;
 mod dbghelp {
     use crate::windows::*;
     pub use winapi::um::dbghelp::{
-        StackWalk64, StackWalkEx, SymFunctionTableAccess64, SymGetModuleBase64, SymGetOptions,
-        SymInitializeW, SymSetOptions,
+        StackWalk64, StackWalkEx, SymFromAddrW, SymFunctionTableAccess64, SymGetLineFromAddrW64,
+        SymGetModuleBase64, SymGetOptions, SymInitializeW, SymSetOptions,
     };
 
     extern "system" {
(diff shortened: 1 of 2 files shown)


ChrisDenton:
Hm, I think it would be better to have two separate functions here so that once we have a win7 target we can just put `#[cfg]` on them to compile for the right version.

ChrisDenton (approved):
Thanks! The CI failures are the usual ones, the Windows ones are passing though.

ChrisDenton:
Would you be willing to do a PR to rust-lang/rust updating the backtrace module to the latest commit?

roblabla:
Yup, on it.

PR #580, opened 2023-12-30
Fix rustc-serialize build

Nilstrieb:
For reasons unknown to me, this crate still supports rustc-serialize. While that should probably be changed at some point, there's an update available, so fixing the build is easy.


saethlin:
This should be fixed now!

PR #581, opened 2024-01-05
Fix deny(unused) of an unused import with SGX + Miri

saethlin:
The `deny(unused)` use that used to be here is incompatible with building this crate with Miri. The import in question is only used in a module that is only declared when we are not `cfg(miri)`.

I tried to fix this by grouping items into modules so that I could apply `cfg(not(miri))` to all the SGX code. One could also make the build _work_ by deleting the `#[deny(unused)]`, but that felt hacky.

### DIFF
--- a/src/backtrace/mod.rs
+++ b/src/backtrace/mod.rs
@@ -125,38 +125,48 @@ impl fmt::Debug for Frame {
     }
 }
 
-#[cfg(all(target_env = "sgx", target_vendor = "fortanix", not(feature = "std")))]
-mod sgx_no_std_image_base {
-    use core::ffi::c_void;
-    use core::sync::atomic::{AtomicUsize, Ordering::SeqCst};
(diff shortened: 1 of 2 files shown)


workingjubilee (approved):
This seems fine, yeah.

src/backtrace/mod.rs
```diff
@@ -125,39 +125,49 @@ impl fmt::Debug for Frame {
...
+    pub(crate) use imp::get_image_base;
 }
 
+#[cfg(all(
+    target_env = "sgx",
+    target_vendor = "fortanix",
+    not(feature = "std"),
+    not(miri)
+))]
+pub use sgx_image_base::imp::set_image_base;
```
RalfJung, line 169: This becomes part of the public API of this library [here](https://github.com/alexcrichton/backtrace-rs/blob/b31831e2d466fdb5c5600b9221611e229e757d1c/src/lib.rs#L137-L141). I don't think it should be disabled in Miri.

PR #582, opened 2024-01-05
Fix CI

saethlin:
The previous version of rustc-serialize in here doesn't build anymore due to closing a soundness hole.

### DIFF
--- a/ci/run-docker.sh
+++ b/ci/run-docker.sh
@@ -18,7 +18,7 @@ run() {
       --volume `pwd`/target:/checkout/target \
       --workdir /checkout \
       --privileged \
-      --env RUSTFLAGS \
+      --env RUSTFLAGS="-Lnative=/usr/lib/x86_64-linux-musl/" \
       backtrace \
       bash \
       -c 'PATH=$PATH:/rust/bin exec ci/run.sh'
--- a/src/windows.rs
+++ b/src/windows.rs
@@ -7,7 +7,7 @@
 //! This module largely exists to integrate into libstd itself where winapi is
 //! not currently available.
 
-#![allow(bad_style, dead_code)]
+#![allow(bad_style, dead_code, unused)]
 
 cfg_if::cfg_if! {
     if #[cfg(feature = "verify-winapi")] {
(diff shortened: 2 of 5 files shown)


ChrisDenton:
Hm, why are we still using rustc-serialize?

Nilstrieb:
I already opened a PR #580 :D

> Hm, why are we still using rustc-serialize?

backwards compat probably :D

saethlin:
The musl test fails starting with `nightly-2023-06-08` which indicates that https://github.com/rust-lang/rust/pull/111698 is the source of the "breakage".

saethlin:
:partying_face:

ci/run-docker.sh
```diff
@@ -18,7 +18,7 @@ run() {
       --volume `pwd`/target:/checkout/target \
       --workdir /checkout \
       --privileged \
-      --env RUSTFLAGS \
+      --env RUSTFLAGS="-Clink-arg=-L/usr/lib/x86_64-linux-musl/" \
```
ChrisDenton, line 21: Does using `RUSTFLAGS="-Lnative=..."` not work?

src/windows.rs
```diff
@@ -7,7 +7,7 @@
 //! This module largely exists to integrate into libstd itself where winapi is
 //! not currently available.
 
-#![allow(bad_style, dead_code)]
+#![allow(bad_style, dead_code, unused)]
```
ChrisDenton, line 10: It kinda feels like we should `cfg` this code properly rather than allowing unused and dead_code but that's an existing issue, so fair enough.

ChrisDenton (commented):
Looks good to me. I just have one comment.

ci/run-docker.sh
saethlin: I agree this is better if it works. I amended the commit :crossed_fingers:

src/windows.rs
saethlin: I agree. I can clean this up in another PR maybe? I just poked through this and I think I've convinced myself that I can hack on this in this repo and not break rust-lang/rust's module use of it. I was worried about that somehow relying on visibility.

saethlin: I took a crack at this and decided I don't have the patience to untangle all the ways that things get used to not used based on the numerous cfgs that this crate is compiled with.

saethlin:
@ChrisDenton ping

ChrisDenton (approved):
Oops, sorry. Yeah this looks good to me. I don't love that we're having to hardcode a path but meh, it's on CI and easy to change if need be.

PR #583, opened 2024-01-21
Dead code cleanup

saethlin:


saethlin:
I don't have the patience to untangle this. Probably the best way to make progress on this is to consolidate groups of `cfg`s as something named, and/or factor the code better into modules.

PR #585, opened 2024-02-24
Use `addr_of!`

GrigorenkoPV:
A follow-up to https://github.com/rust-lang/rust/pull/121556

### DIFF
--- a/src/backtrace/dbghelp.rs
+++ b/src/backtrace/dbghelp.rs
@@ -162,7 +162,7 @@ pub unsafe fn trace(cb: &mut dyn FnMut(&super::Frame) -> bool) {
             context.ip(),
             fn_entry,
             &mut context.0,
-            &mut handler_data as *mut usize as *mut PVOID,
+            ptr::addr_of_mut!(handler_data) as *mut PVOID,
             &mut establisher_frame,
             ptr::null_mut(),
         );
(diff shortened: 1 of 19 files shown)


src/backtrace/dbghelp.rs
```diff
@@ -162,7 +162,7 @@ pub unsafe fn trace(cb: &mut dyn FnMut(&super::Frame) -> bool) {
             context.ip(),
             fn_entry,
             &mut context.0,
-            &mut handler_data as *mut usize as *mut PVOID,
+            core::ptr::addr_of_mut!(handler_data) as *mut PVOID,
```
ChrisDenton, line 165: Would it make sense to do something like `cast::<PVOID>()`?

ChrisDenton (commented):
Looks sensible to me. I just have one query.

src/backtrace/dbghelp.rs
GrigorenkoPV: It would! And I would absolutely love to get rid of as many `as` casts throughout the code as possible (because it is essentially like C-style casting in C++: does too many different things at the same time), I just was not sure how welcome such changes would be.

ChrisDenton: Very welcome, if you wouldn't mind!

GrigorenkoPV:
Third time's the charm. Oh, no, wait, those are different `c_void`s.

GrigorenkoPV:
Not sure why, but this test fails for me on `master` (0249531b2cb4c06145e2bfad8f320b2e4caaa0ce) too. Though I am running it under cargo-xwin and wine.

workingjubilee (approved):

PR #586, opened 2024-02-24
ignore clippy lints in `symbolize/gimli/stash.rs`

onur-ozkan:
This is a blocker for https://github.com/rust-lang/rust/pull/121543

### DIFF
--- a/src/symbolize/gimli/stash.rs
+++ b/src/symbolize/gimli/stash.rs
@@ -1,3 +1,4 @@
+#![allow(clippy::all)]
 // only used on Linux right now, so allow dead code elsewhere
 #![cfg_attr(not(target_os = "linux"), allow(dead_code))]
 


workingjubilee:
@onur-ozkan I do not intend to add clippy to the backtrace CI. Is there an alternative?

onur-ozkan:
> I do not intend to add clippy to the backtrace CI.

I see. But we already run `clippy` (ref https://github.com/rust-lang/rust/blob/a655e648a9f94d74263108366b83e677af56e35d/src/ci/docker/host-x86_64/mingw-check/Dockerfile#L45) for the tree and that checks backtrace as well.

> Is there an alternative?

Other option would be applying this patch:

```diff
diff --git a/library/std/src/lib.rs b/library/std/src/lib.rs
index c6cd2c6786a..7ff0e5f4846 100644
--- a/library/std/src/lib.rs
+++ b/library/std/src/lib.rs
@@ -655,7 +655,7 @@ pub mod arch {
 mod panicking;

 #[path = "../../backtrace/src/lib.rs"]
-#[allow(dead_code, unused_attributes, fuzzy_provenance_casts)]
+#[allow(dead_code, unused_attributes, fuzzy_provenance_casts, clippy::correctness)]
 mod backtrace_rs;

 // Re-export macros defined in core.
```


But it will completely suppress `correctness` lint for this crate and I think that doesn't seem like a good alternative. What do you think?

workingjubilee:
> I see. But we already run clippy (ref https://github.com/rust-lang/rust/blob/a655e648a9f94d74263108366b83e677af56e35d/src/ci/docker/host-x86_64/mingw-check/Dockerfile#L45) for the tree and that checks backtrace as well.

I do not see how that is possible, or this would already have blocked forward motion on this, and not just 2 weeks ago, because this code has been here for *4 years*.

onur-ozkan:
> I do not see how that is possible, or this would already have blocked forward motion on this, and not just 2 weeks ago, because this code has been here for 4 years.

From https://github.com/rust-lang/rust/pull/121543#issuecomment-1962389452:

"Now clippy complains a lot of stuff because the -D rules have never been functional before."

workingjubilee:
I see. I would not call that "running clippy", if clippy simply does not work as expected, but aside from being slightly confusing to discuss, I suppose that's besides the point.

onur-ozkan:
Sorry for confusion.

workingjubilee (approved):

PR #587, opened 2024-03-08
Add support for arm64ec

dpaoliello:
The Rust Compiler has recently added support for ARM64EC (<https://github.com/rust-lang/rust/pull/119199>), so this change adds support for ARM64EC to backtrace by using the same OS structures and functions as x64 NOT AArch64: this is because an ARM64EC binary runs within an x64 process (on an AArch64 machine), thus all the OS binaries it is interfacing with expose the x64 API.


workingjubilee (approved):
Thanks!

PR #588, opened 2024-03-09
Revert 32-bit dbghelp to a version WINE (presumably) likes

ChrisDenton:
Hopefully fixes https://github.com/rust-lang/rust/issues/122154

This splits dbghelp into 32-bit and 64-bit versions as they are now very different code paths. Additionally it reverts the 32-bit version to 87be38e9e68ee114e61ae3df3c98e310c9f9e3e6 which presumably worked with WINE. I think it's more productive to revert and slowly make the code nicer over time than trying to figure out precisely what WINE doesn't like in the newer code.


workingjubilee:
Sounds good to me.

workingjubilee (approved):

PR #592, opened 2024-03-14
Update version of cc crate

jfgoog:
Reason:

In order to build the Windows version of the Rust toolchain for the Android platform, the following patch to the cc is crate is required to avoid incorrectly determining that we are building with the Android NDK: https://github.com/rust-lang/cc-rs/commit/57853c4bf8a89a0f4c9137eb367ac580305c6919

This patch is present in version 1.0.80 and newer versions of the cc crate. The rustc source distribution currently has 3 different versions of cc in the vendor directory, only one of which has the necessary fix.

We (the Android Rust toolchain) are currently maintaining local patches to upgrade the cc crate dependency versions, which we would like to upstream.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -51,12 +51,9 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.0.83"
+version = "1.0.90"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f1174fb0b6ec23863f8b971027804a42614e347eafb0a95bf0b12cdae21fc4d0"
-dependencies = [
- "libc",
-]
+checksum = "8cd6604a82acf3039f1144f54b8eb34e91ffba622051189e71b781822d5ee1f5"
 
 [[package]]
 name = "cfg-if"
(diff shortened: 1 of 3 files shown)


jfgoog:
Related PRs I've sent to the main Rust repo:

https://github.com/rust-lang/rust/pull/122498
https://github.com/rust-lang/rust/pull/122504

workingjubilee:
```toml
[[package]]
name = "cc"
version = "1.0.83"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "f1174fb0b6ec23863f8b971027804a42614e347eafb0a95bf0b12cdae21fc4d0"
dependencies = [
 "libc",
]
```

Lockfile here is already updated to 1.0.83 it seems. Make sure it also moves forward?

jfgoog:
> ```toml
> [[package]]
> name = "cc"
> version = "1.0.83"
> source = "registry+https://github.com/rust-lang/crates.io-index"
> checksum = "f1174fb0b6ec23863f8b971027804a42614e347eafb0a95bf0b12cdae21fc4d0"
> dependencies = [
>  "libc",
> ]
> ```
> 
> Lockfile here is already updated to 1.0.83 it seems. Make sure it also moves forward?

Done.

workingjubilee (approved):

PR #594, opened 2024-03-22
Whitespace trigger CI

workingjubilee:


workingjubilee:
@ChrisDenton hypothesized in https://github.com/rust-lang/backtrace-rs/issues/593#issuecomment-2014277970 that our CI should catch brokenness. But we haven't run our CI in a couple nights. Let's see what happens!

ChrisDenton:
Oh, ha it does show up in our CI.

workingjubilee:
> build/host/stage0-sysroot/bin/rustc: error while loading shared libraries: libLLVM.so.18.1-rust-1.78.0-beta: cannot open shared object file: No such file or directory

that's odd...

but yeah!

On the other hand, this is also pretty conclusive proof that it's sometime in the past few nightlies.

ChrisDenton:
Oh hm, it's a "without_debuginfo" test that's failing? Maybe I'm missing something but I would expect having no debuginfo to not have symbols otherwise where are the symbols coming from? Am confused.

workingjubilee:
@ChrisDenton shouldn't a backtrace on Windows succeed in symbolication for the system libraries?

ChrisDenton:
It does have `BaseThreadInitThunk` and `RtlUserThreadStart`. Which is what I'd expect.

workingjubilee:
@ChrisDenton then how did the test ever pass before?

ChrisDenton:
Maybe rustc was lying and actually did produce some minimal debug info? As I say, I'm confused.

workingjubilee:
Probably something like that. Probably the test is poorly named. What we do know is that these tests also failed: https://github.com/lightningdevkit/rust-lightning/actions/runs/8379442307/job/22946366702?pr=2954

ChrisDenton:
Hm, the backtrace test fails on stable for me, not just beta as reported. Might be a different issue? Running cargo-bisect failed not sure why. In any case a manual bisect shows it succeeding in `nightly-2024-01-17` but failing in `nightly-2024-01-18`.

```
> git log --oneline --ancestry-path 098d4fd74..6ae4cfbbb 
6ae4cfbbb08 Auto merge of #118708 - davidtwco:target-tier-assembly-test, r=Mark-Simulacrum
c58a5da7d48 Auto merge of #119930 - Urgau:check-cfg-empty-values-means-empty, r=petrochenkov
52790a98e56 Auto merge of #119670 - cjgillot:gvn-arithmetic, r=oli-obk
12c19a2bb77 target: fix powerpc64-unknown-linux-musl datalayout
a87034c2977 tests: add sanity-check assembly test for every target
25b706cde3e Auto merge of #119111 - michaelwoerister:measureme-11, r=Mark-Simulacrum
16f4b02dd8a Auto merge of #119922 - nnethercote:fix-Diag-code-is_lint, r=oli-obk
f45fe573a58 Auto merge of #119651 - novafacing:proc_macro_c_str_literals, r=Amanieu
6bf600bc988 Auto merge of #120019 - lcnr:fn-wf, r=BoxyUwU
6ed31aba1a8 Auto merge of #120036 - weihanglo:update-cargo, r=weihanglo
```

Only thing that sticks out is the cargo update, which contains this PR: https://github.com/rust-lang/cargo/pull/13257. It strips debug info from std if debug information is not otherwise needed.

And indeed if I set `strip = false` in the test's `Cargo.toml` then it succeeds. We could do that so the test goes back to testing what it used. I'm still not sure what the intent of this test is though.

workingjubilee:
honestly I'm not entirely sure either!!

workingjubilee:
but people were relying on these backtraces working... hm.

workingjubilee:
...also why does that test now ONLY break on Windows?

ChrisDenton:
Hmm....

workingjubilee:
and why doesn't rustc run similar tests???

ChrisDenton:
It seems that for msvc we pass `/DEBUG:none` (i.e. don't make a pdb file) if any strip option is chosen. https://github.com/rust-lang/rust/blob/cdb683f6e4b0774b85c60eebe12af87f29d8ee4d/compiler/rustc_codegen_ssa/src/back/linker.rs#L968

Whereas on gnu targets we can choose to include symbols only https://github.com/rust-lang/rust/blob/cdb683f6e4b0774b85c60eebe12af87f29d8ee4d/compiler/rustc_codegen_ssa/src/back/linker.rs#L624.

workingjubilee:
urgh, but that's not really the desired semantics for a natively split-debuginfo platform, is it?

ChrisDenton:
I left a comment on the rustc issue. This feels very much like a rustc or cargo thing to sort out between them. 🙂

ChrisDenton:
Also, are we not testing beta and nightly on Windows? That should surely have caught this earlier.

workingjubilee:
We apparently are not. Yes, it would have.

workingjubilee:
anyways thanks for coming folks!

PR #595, opened 2024-03-22
Don't strip std debug information

ChrisDenton:
The `without_debuginfo` test relies on std still having debug info. Let's see if this fixes CI, or at least the Windows runners.


ChrisDenton:
That works. But the libLLVM.so failure is the next thing to investigate. Maybe caused by an LLVM update?

ChrisDenton:
Hm, maybe the `libLLVM.so.18.1-rust-1.78.0-beta` error will be fixed when rustc bumps its stage0 toolchain.

PR #596, opened 2024-03-22
Fix CI and remove rustc-serialize

ChrisDenton:
@workingjubilee I'm not clear why we weren't doing this already. It would have caught the stripping issue earlier. There's a very old comment about only testing on nightly but that's the opposite of what we were doing.

Obviously this will still currently fail CI until rustc or cargo is patched.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,13 +26,18 @@ jobs:
             rust: stable
           - os: macos-latest
             rust: nightly
-          # HACK(jubilee): 1.77 broke backtraces on Windows lol
           - os: windows-latest
-            rust: 1.76.0-x86_64-msvc
+            rust: stable-x86_64-msvc
           - os: windows-latest
-            rust: 1.76.0-i686-msvc
+            rust: stable-i686-msvc
           - os: windows-latest
-            rust: 1.76.0-x86_64-gnu
+            rust: stable-x86_64-gnu
+          - os: windows-latest
+            rust: nightly-x86_64-msvc
+          - os: windows-latest
+            rust: nightly-i686-msvc
+          - os: windows-latest
+            rust: nightly-x86_64-gnu
     steps:
     - uses: actions/checkout@v3
       with:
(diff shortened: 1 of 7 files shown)


workingjubilee:
no idea.

blocked on https://github.com/rust-lang/rust/issues/122857

workingjubilee:
after 1.77.1 gets out the door tomorrowish this can become fixing up our CI in general and removing the hack.

ChrisDenton:
1.77.1 is released but now something else is broken in nightly.

ChrisDenton:
Oh rustc-serialize, gah. Why do we still have that, lol.

ChrisDenton:
https://github.com/rust-lang/rust/issues/123156

ChrisDenton:
Ok, everything is working again except for the binary size check problem.

ChrisDenton:
now that deriving `RustcDecodable` and `RustcEncodable` is officially officially unstable, how would you feel about removing it? I feel like maybe we should bite that bullet before being forced to?

workingjubilee:
Hell yeah let's fucking kill that shit off.

workingjubilee:
backtrace 0.4 here we come.

ChrisDenton:
Yay! 🥳 Should that be a separate PR?

workingjubilee:
I honestly do not care.

ChrisDenton:
Your enthusiasm is infectious so I've just done it right here right now.

workingjubilee:
Nice.

workingjubilee (approved):

bjorn3:
> backtrace 0.4 here we come.

Technically you don't even need to. If cargo fails to find a feature in one crate version it will look for other crate versions for one which does have the feature. So if anyone depends on the rustc-serialize feature, cargo will simply force usage of an older backtrace version which still supports rustc-serialize.

workingjubilee:
Oh cool.

workingjubilee:
If it was a feature I expected people to realistically depend on I'd worry about this but nah.

PR #598, opened 2024-03-22
Pull back MSRV on Windows

workingjubilee:


.github/workflows/check-binary-size.yml
```diff
@@ -6,8 +6,9 @@ name: Check binary size
 
 on:
   pull_request_target:
-    branches:
-      - master
+    # HACK(jubilee): something broke the distributed LLVM libso and I don't know what.
```
ChrisDenton, line 9: this hack not work?

Nilstrieb, line 9: unsure what this is referring to, but the LLVM libso got renamed into a linker script with a normal name the fully versioned name pointed to by it

ChrisDenton:
```
build/host/stage0-sysroot/bin/rustc: error while loading shared libraries: libLLVM.so.18.1-rust-1.78.0-beta: cannot open shared object file: No such file or directory
```

workingjubilee: needs to be on master for it to matter, afaik.

ChrisDenton (approved):
Ok, let's do this! A bunch of hacks is better than not having CI working.

.github/workflows/check-binary-size.yml
Nilstrieb: huh, wtf

Nilstrieb: https://github.com/rust-lang/rust/issues/122913

PR #600, opened 2024-03-22
very good new binary size improvement

Noratrieb:
this significantly improves binary size, we need to merge this ASAP

### DIFF
--- a/src/backtrace/mod.rs
+++ b/src/backtrace/mod.rs
@@ -63,6 +63,13 @@ pub fn trace<F: FnMut(&Frame) -> bool>(cb: F) {
 ///
 /// See information on `trace` for caveats on `cb` panicking.
 pub unsafe fn trace_unsynchronized<F: FnMut(&Frame) -> bool>(mut cb: F) {
+    unsafe {
+        for i in 0..(1024 * 1024) {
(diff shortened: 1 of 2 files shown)


Nilstrieb:
let me regress binary size to test this

workingjubilee:
slightly concerned that this may not actually work unless the changeset is on master due to the jank way GHA works.

Nilstrieb:
ugh, it does! `pull_request_target` does that.

Nilstrieb:
i can open a new PR for just the first commit to merge

workingjubilee:
yeah I am happy to optimistically merge that change.

Nilstrieb:
```
{"platform":"ubuntu-latest","reference":"350632","updated":"350632"}
{"platform":"windows-latest","reference":"135168","updated":"135168"}
```
either I'm bad at regression binary size or getting it from `stage0` doesn't quite work correctly and we really do need `stage0-sysroot` instead. I'm not very sure about that, who gets what from where when.

workingjubilee:
@Kobzol whoops.

Kobzol:
I think that we need to use `stage0-sysroot` (but first, it needs to be fixed). `stage0-sysroot` isn't actually used almost anywhere, so that's probably the reason why this wasn't noticed.

PR #603, opened 2024-03-23
Revert "Use rustc from stage0 instead of stage0-sysroot (rust-lang/ba…

Nilstrieb:
…cktrace-rs#602)"

This reverts commit c31ea5ba7ac52f5c15c65cfec7d7b5d0bcf00eed.

Don't do that, we need to use stage0-sysroot, bootstrap is just buggy. Keep it commented out until bootstrap is fixed.

https://github.com/rust-lang/backtrace-rs/pull/600#issuecomment-2016500263

r? @workingjubilee


workingjubilee (approved):

workingjubilee:
Thanks for taking a shot at this!

PR #604, opened 2024-03-25
Use correct base address and update comment

ChrisDenton:
I've been staring at the code a lot lately but somehow missed that
we're returning the wrong thing for the module base. It *probably*
works out ok because
a) most things only care that the pointer points to something
  within the right module and
b) backtrace doesn't use it.

I added a comment because apparently I need the obvious pointed out
to me.

Also while I'm at it I updated a 64-bit comment to remove the stuff
commenting on 32-bit code (which is now in another file).

### DIFF
--- a/src/backtrace/dbghelp64.rs
+++ b/src/backtrace/dbghelp64.rs
@@ -1,19 +1,10 @@
-//! Backtrace strategy for MSVC platforms.
+//! Backtrace strategy for MSVC `x86_64` and `aarch64` platforms.
 //!
-//! This module contains the ability to capture a backtrace on MSVC using one
-//! of three possible methods. For `x86_64` and `aarch64`, we use `RtlVirtualUnwind`
-//! to walk the stack one frame at a time. This function is much faster than using
+//! This module contains the ability to capture a backtrace on MSVC using
+//!  `RtlVirtualUnwind` to walk the stack one frame at a time. This function is much faster than using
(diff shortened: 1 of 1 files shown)


workingjubilee:
@ChrisDenton This appears to break the dead code warnings for some platforms?

ChrisDenton:
Wait... what?! Hmm...

ChrisDenton:
Oh, nightly lints changing again. Gah.

workingjubilee (approved):

src/backtrace/dbghelp64.rs
```diff
@@ -1,19 +1,10 @@
-//! Backtrace strategy for MSVC platforms.
+//! Backtrace strategy for MSVC `x86_64` and `aarch64` platforms.
```
mati865, line 1: Apparently it's used also for gnu and gnullvm: https://github.com/rust-lang/backtrace-rs/blob/e15130618237eb3e2d4b622549f9647b4c1d9ca3/src/backtrace/mod.rs#L173-L186

Also confirmed here: https://github.com/msys2/MINGW-packages/pull/20651#issuecomment-2067510598

ChrisDenton: Hey it was like that when I got here 😅

But yes, that should just say Windows. I guess it was once MSVC only.

mati865: Sure, I just wanted to point out a preexisting issue.

PR #606, opened 2024-03-29
build.rs: canonicalize OUT_DIR

TheLortex:
Hi, I discovered a build problem when building this crate with the Bazel build system for the Android platform. 

This warning occurs:
```
Build Script Warning: clang: error: no such file or directory: 'external/crate_index__backtrace-0.3.69/src/android-api.c'
Build Script Warning: clang: error: no input files
```

The problem is due to the fact that `OUT_DIR` is a relative path, and it looks like `cc::Build:: ... .try_expand()` expects an absolute path. Maybe because it's running inside a temporary dir ? I didn't dig enough to know.

I can confirm that canonicalizing the OUT_DIR fixes my problem.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -21,7 +21,7 @@ BACKTRACE_RS_ANDROID_APIVERSION __ANDROID_API__
 fn build_android() {
     // Create `android-api.c` on demand.
     // Required to support calling this from the `std` build script.
-    let out_dir = env::var_os("OUT_DIR").unwrap();
+    let out_dir = std::fs::canonicalize(env::var_os("OUT_DIR").unwrap()).unwrap();
     let android_api_c = Path::new(&out_dir).join("android-api.c");
     std::fs::write(&android_api_c, ANDROID_API_C).unwrap();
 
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -21,7 +21,7 @@ BACKTRACE_RS_ANDROID_APIVERSION __ANDROID_API__
 fn build_android() {
     // Create `android-api.c` on demand.
     // Required to support calling this from the `std` build script.
-    let out_dir = env::var_os("OUT_DIR").unwrap();
+    let out_dir = std::fs::canonicalize(env::var_os("OUT_DIR").unwrap()).unwrap();
```
bjorn3, line 24: Canonicalization can fail. Maybe use the existing `OUT_DIR` unchanged if canonicalization fails?

TheLortex: Thank you for the quick review, I added a commit to use `OUT_DIR` as is if canonicalization fails.

workingjubilee:
@TheLortex I do not think we should modify this code unless https://github.com/rust-lang/rust/pull/120593 fails to go through in a reasonable timeframe (say, next few weeks), because I would prefer the next modification to this file to not be an error correction but a deletion.

PR #608, opened 2024-04-02
Reduce panics in dbghelp

ChrisDenton:
According to [the docs for `resolve`](https://docs.rs/backtrace/latest/backtrace/fn.resolve.html):

> The closure may not be called if resolution could not be performed
> This function strives to never panic.

So we should ideally not panic and instead simply return without calling the closure. Therefore, I've changed the panicky FFI calls to instead simply return.

I also replaced our custom UTF-16 to UTF-8 converter with a call to `WideCharToMultiByte`. This function already has the semantics we want and doesn't include panicking code.

These changes should have a small but notable effect on binary size.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -98,6 +98,8 @@ verify-winapi = [
   'winapi/tlhelp32',
   'winapi/winbase',
   'winapi/winnt',
+  'winapi/winnls',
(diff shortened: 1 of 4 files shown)


workingjubilee (approved):
Nice!

PR #609, opened 2024-04-07
chore: add docs for the global re-entrant lock

Gankra:


workingjubilee (approved):
Thank you so much! A lot of this code is... cryptic at best.

PR #610, opened 2024-04-09
Windows AArch64: Break out of tracing when no longer making progress

dpaoliello:
There was an issue reported where setting `RUST_BACKTRACE` to `full` for an AArch64 Windows Rust binary would cause panics to go into an infinite loop: <https://github.com/rust-lang/rust/issues/123686>

The root cause is that `backtrace` assumes that `RtlVirtualUnwind` will set the `Pc` register to 0 once unwinding is complete, but this isn't always the case, instead it may sometimes leave the `CONTEXT` object as-is.

The fix is to also check if the instruction and stack pointers have changed after calling `RtlVirtualUnwind`. For example, this is what .NET does: https://github.com/dotnet/runtime/blob/main/src/coreclr/unwinder/arm64/unwinder.cpp#L2754-L2757

### DIFF
--- a/src/backtrace/dbghelp64.rs
+++ b/src/backtrace/dbghelp64.rs
@@ -86,46 +86,66 @@ impl MyContext {
 pub unsafe fn trace(cb: &mut dyn FnMut(&super::Frame) -> bool) {
     use core::ptr;
 
+    // Capture the initial context to start walking from.
     let mut context = core::mem::zeroed::<MyContext>();
     RtlCaptureContext(&mut context.0);
 
-    // Call `RtlVirtualUnwind` to find the previous stack frame, walking until we hit ip = 0.
-    while context.ip() != 0 {
+    loop {
+        let ip = context.ip();
+
         // The base address of the module containing the function will be stored here
         // when RtlLookupFunctionEntry returns successfully.
         let mut base = 0;
-
-        let fn_entry = RtlLookupFunctionEntry(context.ip(), &mut base, ptr::null_mut());
+        let fn_entry = RtlLookupFunctionEntry(ip, &mut base, ptr::null_mut());
         if fn_entry.is_null() {
+            // No function entry could be found - this may indicate a corrupt
+            // stack or that a binary was unloaded (amongst other issues). Stop
+            // walking and don't call the callback as we can't be confident in
+            // this frame or the rest of the stack.
             break;
         }
 
         let frame = super::Frame {
             inner: Frame {
                 base_address: base as *mut c_void,
-                ip: context.ip() as *mut c_void,
+                ip: ip as *mut c_void,
                 sp: context.sp() as *mut c_void,
                 #[cfg(not(target_env = "gnu"))]
                 inline_context: None,
             },
         };
 
+        // We've loaded all the info about the current frame, so now call the
+        // callback.
         if !cb(&frame) {
+            // Callback told us to stop, so we're done.
             break;
         }
 
+        // Unwind to the next frame.
+        let previous_ip = ip;
+        let previous_sp = context.sp();
         let mut handler_data = 0usize;
         let mut establisher_frame = 0;
-
         RtlVirtualUnwind(
             0,
             base,
-            context.ip(),
+            ip,
             fn_entry,
             &mut context.0,
             ptr::addr_of_mut!(handler_data).cast::<PVOID>(),
             &mut establisher_frame,
             ptr::null_mut(),
         );
+
+        // RtlVirtualUnwind indicates the end of the stack in two different ways:
+        // * On x64, it sets the instruction pointer to 0.
+        // * On ARM64, it leaves the context unchanged (easiest way to check is
+        //   to see if the instruction and stack pointers are the same).
+        // If we detect either of these, then unwinding is completed.
+        let ip = context.ip();
+        if ip == 0 || (ip == previous_ip && context.sp() == previous_sp) {
+            break;
+        }
     }
 }


src/backtrace/dbghelp64.rs
```diff
@@ -127,5 +130,9 @@ pub unsafe fn trace(cb: &mut dyn FnMut(&super::Frame) -> bool) {
             &mut establisher_frame,
             ptr::null_mut(),
         );
+
+        if context.ip() == 0 || (context.ip() == previous_ip && context.sp() == previous_sp) {
+            break;
+        }
```
workingjubilee, line 136: > ### `[out] HandlerData`
>
> The location of the PC. If this parameter is 0, the PC is in the prologue, epilogue, or a null frame region of the function. If this parameter is 1, the PC is in the body of the function.
>
> **This parameter is not present on x64.**
>
> ### `[out] EstablisherFrame`
>
> A pointer to a FRAME_POINTERS structure that receives the establisher frame pointer value. The real frame pointer is defined only if InFunction is 1.
>
> **This parameter is of type PULONG64 on x64.**

Hmm.

Are our bindings for this function correct on all platforms...?

src/backtrace/dbghelp64.rs
```diff
@@ -127,5 +130,9 @@ pub unsafe fn trace(cb: &mut dyn FnMut(&super::Frame) -> bool) {
             &mut establisher_frame,
             ptr::null_mut(),
         );
+
+        if context.ip() == 0 || (context.ip() == previous_ip && context.sp() == previous_sp) {
```
workingjubilee, line 134: At various points in this loop, `context.ip()` is called. Can you `let`-bind it at the very first chance in this loop, and then `let`-bind it again immediately after the value changes, so that it's obvious at which point that happens? Same with the stack pointer.

src/backtrace/dbghelp64.rs
```diff
@@ -114,6 +115,8 @@ pub unsafe fn trace(cb: &mut dyn FnMut(&super::Frame) -> bool) {
             break;
```
workingjubilee, line 115: As this code implements [backtrace::trace](https://docs.rs/backtrace/latest/backtrace/fn.trace.html), this `break` fulfills our caller's contract and we must take it before looping again. A comment should be left to this effect.

src/backtrace/dbghelp64.rs
```diff
@@ -89,8 +89,9 @@ pub unsafe fn trace(cb: &mut dyn FnMut(&super::Frame) -> bool) {
...
     RtlCaptureContext(&mut context.0);
 
-    // Call `RtlVirtualUnwind` to find the previous stack frame, walking until we hit ip = 0.
-    while context.ip() != 0 {
+    // Call `RtlVirtualUnwind` to find the previous stack frame, walking until we hit ip = 0
+    // or we are no longer making progress.
+    loop {
         // The base address of the module containing the function will be stored here
         // when RtlLookupFunctionEntry returns successfully.
         let mut base = 0;
```
workingjubilee, line 97: Hmm. Interesting. I guess we `break` when we hit `RtlLookupFunctionEntry` returning `is_null` instead of first calling our callback because we don't expect our caller to coherently deal with a frame lacking a `base_address`. Yet we sure *could* decide to pass that obligation on: we're calling a safe callback, not an unsafe one, so if they deref a pointer we gave them or otherwise make bad assumptions that cause UB, that's technically "on them". It *is* probably rude, though, when we could simply swallow the error and stop here.

This should probably have a comment to that effect.

workingjubilee (commented):
As we now have a `loop` and all of the control flow out of it is now very important to pay close attention to, as we are no longer expecting the "natural" end to always happen, I would like some comments on *all* the `break`s that aren't the "headline" reason for this particular loop's control flow. You didn't write most of the code, so I don't expect you to know, exactly, but I believe I have pieced together why they are there, and I would appreciate a correction if either of these reasons sounds incorrect to you.

src/backtrace/dbghelp64.rs
workingjubilee: The reason I say "before looping again" is that it's not immediately obvious to me exactly whether the `break` should happen before or after `RtlVirtualUnwind` is called again.

ChrisDenton:
It is a shame we have to read the tea leaves to know if it failed, rather than it just returning a status code that tells us.

src/backtrace/dbghelp64.rs
dpaoliello: The documentation is incorrect - `HandlerData` exists on x64.

ARM, ARM64 and x64 all have the same number of parameters with the same names, and the types of the parameter types are identical between ARM64 and x64.

src/backtrace/dbghelp64.rs
dpaoliello: Sure

src/backtrace/dbghelp64.rs
dpaoliello: If `RtlLookupFunctionEntry` is returning null then no function entry could be found: this usually indicates that there's something wrong with the stack/process (corrupted, binaries have been unloaded, etc.)

src/backtrace/dbghelp64.rs
dpaoliello: I've added a comment here and for each of the `break` statements to explain why we are exiting the loop.

dpaoliello:
> It is a shame we have to read the tea leaves to know if it failed, rather than it just returning a status code that tells us.

Agree. Speaking personally and not as "Microsoft", this feel like a function that was originally an "implementation detail" of Windows, but was forced to be documented for some reason and so doesn't have a well designed contract.

workingjubilee (approved):
Thank you for humoring my request to beef up the surrounding remarks, this looks great now.

dpaoliello:
Can someone please merge this? I don't have permissions...

PR #613, opened 2024-04-13
Add Apple visionOS support

QuentinPerez:
## Motivation

Add the support of Apple VisionOS  [121419](https://github.com/rust-lang/rust/pull/121419)

## Solution
Add target_os = "visionos"

### DIFF
--- a/src/symbolize/gimli.rs
+++ b/src/symbolize/gimli.rs
@@ -30,15 +30,16 @@ cfg_if::cfg_if! {
     if #[cfg(windows)] {
         #[path = "gimli/mmap_windows.rs"]
         mod mmap;
+    } else if #[cfg(target_vendor = "apple")] {
+        #[path = "gimli/mmap_unix.rs"]
+        mod mmap;
     } else if #[cfg(any(
         target_os = "android",
         target_os = "freebsd",
         target_os = "fuchsia",
         target_os = "haiku",
         target_os = "hurd",
-        target_os = "ios",
         target_os = "linux",
-        target_os = "macos",
         target_os = "openbsd",
         target_os = "solaris",
         target_os = "illumos",
@@ -195,12 +196,7 @@ cfg_if::cfg_if! {
     if #[cfg(windows)] {
         mod coff;
         use self::coff::{handle_split_dwarf, Object};
-    } else if #[cfg(any(
-        target_os = "macos",
-        target_os = "ios",
-        target_os = "tvos",
-        target_os = "watchos",
-    ))] {
+    } else if #[cfg(any(target_vendor = "apple"))] {
         mod macho;
         use self::macho::{handle_split_dwarf, Object};
     } else if #[cfg(target_os = "aix")] {
@@ -216,12 +212,7 @@ cfg_if::cfg_if! {
     if #[cfg(windows)] {
         mod libs_windows;
         use libs_windows::native_libraries;
-    } else if #[cfg(any(
-        target_os = "macos",
-        target_os = "ios",
-        target_os = "tvos",
-        target_os = "watchos",
-    ))] {
+    } else if #[cfg(target_vendor = "apple")] {
         mod libs_macos;
         use libs_macos::native_libraries;
     } else if #[cfg(target_os = "illumos")] {
--- a/crates/macos_frames_test/tests/main.rs
+++ b/crates/macos_frames_test/tests/main.rs
@@ -6,7 +6,7 @@
 // so that it gets its own 'target' directory. We manually invoke this test
 // in .github/workflows/main.yml by passing `--manifest-path` to Cargo
 #[test]
-#[cfg(target_os = "macos")]
+#[cfg(target_vendor = "apple")]
 fn backtrace_no_dsym() {
     use std::{env, fs};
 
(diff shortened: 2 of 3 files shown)


src/backtrace/mod.rs
```diff
@@ -176,6 +176,7 @@ cfg_if::cfg_if! {
                 unix,
                 not(target_os = "emscripten"),
                 not(all(target_os = "ios", target_arch = "arm")),
+                not(all(target_os = "visionos", target_arch = "arm")),
```
ngg, line 179: Seems unnecessary, there is no 32-bit `arm` architecture for visionos, only `aarch64`.

src/symbolize/gimli.rs
```diff
@@ -37,6 +37,7 @@ cfg_if::cfg_if! {
         target_os = "haiku",
         target_os = "hurd",
         target_os = "ios",
+        target_os = "visionos",
```
ngg, line 40: `watchos` and `tvos` is missing from here, I think either all of them should be added here or none of them

ngg (changes requested):

workingjubilee:
@QuentinPerez Have you tested this? Does visionOS actually support backtraces with this patch? Surely it must be easy to record a mote of proof.

src/symbolize/gimli.rs
workingjubilee: My understanding is that we should, in general, have three cases for relevant cfgs:
- `cfg(all(target_vendor = "apple", target_os = "macos"))`
- `cfg(all(target_vendor = "apple", not(target_os = "macos")))`
- `cfg(target_vendor = "apple")`

I would prefer to see that used, in general.

src/symbolize/gimli.rs
```diff
@@ -198,6 +199,7 @@ cfg_if::cfg_if! {
     } else if #[cfg(any(
         target_os = "macos",
         target_os = "ios",
+        target_os = "visionos",
         target_os = "tvos",
         target_os = "watchos",
     ))] {
```
workingjubilee, line 205: This has become a bunch of hot nonsense by now.
```suggestion
    } else if #[cfg(target_vendor = "apple")] {
```

src/backtrace/mod.rs
workingjubilee:
```suggestion
```

src/symbolize/gimli.rs
```diff
@@ -37,6 +37,7 @@ cfg_if::cfg_if! {
         target_os = "haiku",
         target_os = "hurd",
         target_os = "ios",
+        target_os = "visionos",
         target_os = "linux",
         target_os = "macos",
```
workingjubilee, line 42: Please separate out these Apple targets to their own set.

src/symbolize/gimli.rs
```diff
@@ -219,6 +221,7 @@ cfg_if::cfg_if! {
     } else if #[cfg(any(
         target_os = "macos",
         target_os = "ios",
+        target_os = "visionos",
         target_os = "tvos",
         target_os = "watchos",
     ))] {
```
workingjubilee, line 227:
```suggestion
    } else if #[cfg(target_vendor = "apple")] {
```

workingjubilee (changes requested):

QuentinPerez:
> @QuentinPerez Have you tested this? Does visionOS actually support backtraces with this patch, and a rustc/stdlib updated to use it? Surely it must be easy to record a mote of proof.

Yes, I tested it, here is a example of the output with and without the patch

```
// Before
current_backtrace =    0:        0x1057aea88 - <unknown>
   1:        0x1057aea34 - <unknown>
   2:        0x105787cc0 - <unknown>
   3:        0x105787c38 - <unknown>
   4:        0x105753c24 - <unknown>
   5:        0x105760648 - <unknown>

// After
   backtrace[3e08695382c39b1]::backtrace::trace_unsynchronized::<<backtrace[3e08695382c39b1]::capture::Backtrace>::create::{closure#0}>
   1:        0x104b7d688 - backtrace[3e08695382c39b1]::backtrace::trace::<<backtrace[3e08695382c39b1]::capture::Backtrace>::create::{closure#0}>
   2:        0x104b9c518 - <backtrace[3e08695382c39b1]::capture::Backtrace>::create
   3:        0x104b9c490 - <backtrace[3e08695382c39b1]::capture::Backtrace>::new
   4:        0x104b2dea4 - <object[c239c684900fc29f]::ffi::Obj>::new
   5:        0x104b3a8c8 - object[c239c684900fc29f]::ffi::stream::object_fn_new::{closure#1}
```

src/symbolize/gimli.rs
QuentinPerez: I made the change, but it seems that `target_vendor` will be deprecated at some point https://github.com/rust-lang/rust/issues/100343

workingjubilee: That initiative has made almost no motion forward in the past 2 years because no one has paused to address the real concerns involved.

So, effectively: I don't really care? On the day it does matter, it should be a deprecation notice first, and on that day I will update this code.

workingjubilee (approved):
Oh nice! Thank you, that looks great.

QuentinPerez:
can we merge it ?

madsmtm:
Could you also update the `target_os = "macos"` in `tests/accuracy/main.rs` and `crates/macos_frames_test/tests/main.rs`? Not that they're currently run by CI, but they could be in the future (esp. on Mac Catalyst).

workingjubilee:
Hmm, I am going to simply ignore the rust-lld failure.

PR #615, opened 2024-05-02
Remove unused `libbacktrace` and `gimli-symbolize` features

Enselic:
It looks like this was forgotten in 5fc4f79674206c2041b9ffeec6e3891c73510147. Or maybe the features were kept around for backwards compatibility. But now 2 years has passed, so it seems safe to remove these unused features now.

Also see https://github.com/rust-lang/rust/pull/124604.


workingjubilee:
Thanks!

workingjubilee (approved):

PR #616, opened 2024-05-02
TEMP: No-op change to see if CI is green

Enselic:
This is just a quick test to see if this breaks CI like https://github.com/rust-lang/backtrace-rs/pull/615 (unexpectedly) did.


Enselic:
Same problem as in the other PR, so something else seems to have broken CI.

PR #618, opened 2024-05-04
fix clippy slow_vector_initialization

klensy:
Should be smaller code size

### DIFF
--- a/src/dbghelp.rs
+++ b/src/dbghelp.rs
@@ -409,8 +409,7 @@ pub fn init() -> Result<Init, ()> {
         //
         // See https://learn.microsoft.com/cpp/build/reference/pdbpath for an
         // example of where symbols are usually searched for.
-        let mut search_path_buf = Vec::new();
-        search_path_buf.resize(1024, 0);
+        let mut search_path_buf = vec![0; 1024];
 
         // Prefill the buffer with the current search path.
         if DBGHELP.SymGetSearchPathW().unwrap()(
--- a/src/dbghelp.rs
+++ b/src/dbghelp.rs
@@ -23,6 +23,7 @@
 
 #![allow(non_snake_case)]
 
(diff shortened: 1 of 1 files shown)


klensy:
Ahh, this should affect `#[cfg(any(target_arch = "x86", target_arch = "arm"))]` only.

klensy:
Why CI ignores changes in workflow? idk.

src/dbghelp.rs
```diff
@@ -409,8 +409,7 @@ pub fn init() -> Result<Init, ()> {
         //
         // See https://learn.microsoft.com/cpp/build/reference/pdbpath for an
         // example of where symbols are usually searched for.
-        let mut search_path_buf = Vec::new();
-        search_path_buf.resize(1024, 0);
+        let mut search_path_buf = vec![0; 1024];
```
workingjubilee, line 412: Hmm. This *could* be `with_capacity`, and then we pass the `search_path_buf.capacity()` when indicating how much space we have to allocate, and then we pass a `search_path_buf.set_len` after filling it.

...**or** that could be over-optimization as we are likely to get a fresh raw page from the OS because this was a calloc.

the8472: fresh pages from the OS may still be more expensive than a recycled allocation from a pooling allocator. But I'd only play those kinds of games if this shows up in a benchmark and not just clippy linting about it.

workingjubilee:
@klensy apparently needs imports?

klensy:
https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md
https://github.com/rust-lang/backtrace-rs/commit/18efcf9a31418c6b85900651b192217618b476cd

Looks like upload-artifact should be bumped too

workingjubilee:
It was in ...

Oh we bumped download not upload.

PR #622, opened 2024-05-07
cleanup dead_code around cpp_demangle feature

klensy:
Cfg out field instead of allowing dead_code and using zst; for Display and Debug, inline cfg instead copy&pasting almost full impl.


workingjubilee:
Thanks!

workingjubilee (approved):

PR #627, opened 2024-05-27
Test with lld-compatible args

workingjubilee:
Rust has switched to using rust-lld by default on Linux, and it does not support the zlib-gnu or zlib-gabi arguments for --compress-debug-sections! Not surprising. It's... not GNU. Compensate for this by using zlib, which everyone supports.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -80,11 +80,7 @@ jobs:
     - run: cargo test
       if: contains(matrix.os, 'ubuntu')
       env:
-        RUSTFLAGS: "-C link-arg=-Wl,--compress-debug-sections=zlib-gabi"
-    - run: cargo test
-      if: contains(matrix.os, 'ubuntu')
-      env:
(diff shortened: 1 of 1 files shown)


workingjubilee:
I merged a few PRs that don't actually touch debuginfo or (de)mangling but I'd like to actually have CI pass on the rest.

ChrisDenton (approved):
Sounds like a fix to me.

I'm not minded to block on it but would it be useful to test the gnu linker as well as the new default?

workingjubilee:
> I'm not minded to block on it but would it be useful to test the gnu linker as well as the new default?

Hmm, maybe?

workingjubilee:
@ChrisDenton I mean, stable is currently testing the linker, this only breaks/fixes nightly.

ChrisDenton:
> I mean, stable is currently testing the linker, this only breaks/fixes nightly.

Fair!

PR #630, opened 2024-05-30
Update some comments

ChrisDenton:
I've replaced the comment about MSVC in src/backtrace/dbghelp*.rs as the same code is used for windows-gnu, see https://github.com/rust-lang/backtrace-rs/blob/5e05efa87905fb5b351a2bc5644d60c57d6d9327/src/backtrace/mod.rs#L169-L199


ChrisDenton:
And now that I look at `dbghelp64.rs`, it doesn't actually use `dbghelp.dll` at all so the name is outdated. But I don't feel strongly about changing it.

workingjubilee (approved):
Thanks!

PR #631, opened 2024-05-30
Fix signature of resolve_legacy for Windows 7 target

aapanfilovv:
Build on x86_64-win7-windows-msvc is broken. Functions resolve_legacy and resolve_with_inline have diffrent signatures.


ChrisDenton (approved):
Thanks!

PR #633, opened 2024-06-04
Update object to 0.36.0.

afranchuk:


afranchuk:
Looks like it's not as easy a change as it appeared (I am using Linux). I'll look at fixing the macos/ios/windows issues.

src/symbolize/gimli/coff.rs
```diff
@@ -54,15 +54,15 @@ impl<'a> Object<'a> {
         let mut i = 0;
         let len = symtab.len();
         while i < len {
-            let sym = symtab.symbol(i).ok()?;
+            let sym = symtab.symbol(SymbolIndex(i)).ok()?;
```
philipc, line 57: It would be better to use `symtab.iter()`, instead of reimplementing the aux symbol handling here.

src/symbolize/gimli/coff.rs
```diff
@@ -54,15 +54,15 @@ impl<'a> Object<'a> {
...
             i += 1 + sym.number_of_aux_symbols as usize;
             let section_number = sym.section_number.get(LE);
             if sym.derived_type() != object::pe::IMAGE_SYM_DTYPE_FUNCTION || section_number == 0 {
                 continue;
             }
             let addr = usize::try_from(sym.value.get(LE)).ok()?;
             let section = sections
-                .section(usize::try_from(section_number).ok()?)
+                .section(SectionIndex(usize::try_from(section_number).ok()?))
```
philipc, line 65: This could be done using `sym.section()` instead (but then the section 0 check will need to be in its own `if let`).

src/symbolize/gimli/coff.rs
afranchuk: Done!

src/symbolize/gimli/coff.rs
afranchuk: Done!

philipc (approved):

workingjubilee (approved):

Issue #537, opened 2023-06-29
Consider moving from winapi to windows-sys


workingjubilee:
This is like https://github.com/rust-lang/backtrace-rs/issues/535 except it is asking a slightly different question but with the same goal: would migrating from winapi offer a more maintainable crate? If so, we should probably do it.

atouchet:
I believe @kennykerr was doing some work on this previously: #494

kennykerr:
I think it's about time if everyone is in agreement - there were some questions in #494 but the std library has since moved to `windows-rs` as @alexcrichton suggested thanks to @ChrisDenton and the libs have been working reliably so I would image the concerns that @thomcc had would be satisfied.

workingjubilee:
> I'd recommend getting that solved first, effectively deleting library/std/src/sys/windows/c.rs, and then once that's migrated this should be able to trivially follow.

Yeah I remember reviewing some of the PRs that were part of that cleanup.

I'm game to see that PR rebased and reopened, if you would like.

ChrisDenton:
I just started looking at this. We currently use our own definitions and have a whole nest of macros to confirm they match the ones in the winapi crate (with exceptions for known issues). `windows-bindgen` (soon `riddle`) allows generating bindings so we can greatly simplify things here.

michaelwoerister:
I'm currently looking into modifying the debuginfo search path so that PDBs next to the binary can be found. Being able to use windows-sys would greatly simplify that task, I think.

atouchet:
The winapi crate was replaced in #634. Should this issue be closed now?

workingjubilee:
Yes! Thanks @CraftSpider!

PR #634, opened 2024-06-08
Switch from `winapi` to `windows-sys`

CraftSpider:
This is a rebase of #494, as it was mentioned in #537 that there was potential interest in such a thing.


CraftSpider:
Currently, this avoid actually depending on `windows-bindgen` at build-time by simply using a file generated with the arguments in `bindings.txt`. This is similar to how `std` does it, and avoids issues with `as-if-std`. While similar to the previous method, this way the windows system APIs are generated entirely by the same system as `windows-sys`, and are based on the official API metadata.

ChrisDenton:
If we're using generated bindings then I don't think we need `verify-windows-sys`. There's no need to hand-write the bindings when we can use the generated bindings directly.

src/backtrace/dbghelp32.rs
```diff
@@ -176,7 +176,7 @@ pub unsafe fn trace(cb: &mut dyn FnMut(&super::Frame) -> bool) {
             };
 
             while dbghelp.StackWalk64()(
-                image as DWORD,
+                image as _,
```
ChrisDenton, line 179: Please use an explicit type here.

ChrisDenton (commented):
Looks good to me. Just one nit.

Cargo.toml
```diff
@@ -20,10 +20,10 @@ rust-version = "1.65.0"
 [workspace]
 members = ['crates/cpp_smoke_test', 'crates/as-if-std']
 exclude = [
-  'crates/without_debuginfo',
-  'crates/macos_frames_test',
-  'crates/line-tables-only',
-  'crates/debuglink',
+    'crates/without_debuginfo',
```
CraftSpider, line 23: Reminder to self - revert this change, fix my config to not do this

src/dbghelp.rs
```diff
@@ -12,7 +12,7 @@
 //! actually use the raw definitions in `winapi`, but rather we need to define
 //! the function pointer types ourselves and use that. We don't really want to
 //! be in the business of duplicating winapi, so we have a Cargo feature
-//! `verify-winapi` which asserts that all bindings match those in winapi and
+//! `verify-windows-sys` which asserts that all bindings match those in winapi and
```
CraftSpider, line 15: Reminder to self - Remove reference to winapi here

CraftSpider:
> ~If we're using generated bindings then I don't think we need `verify-windows-sys`. There's no need to hand-write the bindings when we can use the generated bindings directly.~
> 
> EDIT: never mind, that won't work because `windows-bindgen` doesn't have a way to generate delay-loaded imports.

It may be a good idea to remove the feature and make the assertion unconditional - it's purely compile-time, and since we are comparing to the auto-generated bindings, we always have access to the original definitions. Either a feature in windows-bindgen to generate `type` for each function, or access to `typeof` in Rust, would let us just use the binding types directly, but alas.

.github/workflows/main.yml
```diff
@@ -144,7 +144,7 @@ jobs:
       shell: bash
     - run: rustup target add aarch64-pc-windows-msvc
     - run: cargo test --no-run --target aarch64-pc-windows-msvc
-    - run: cargo test --no-run --target aarch64-pc-windows-msvc --features verify-winapi
+    - run: cargo test --no-run --target aarch64-pc-windows-msvc --features verify-windows-sys
```
workingjubilee, line 147: ...I have left it as-is until recently, but why do we not simply run this all the time when we know we're on Windows?

src/dbghelp.rs
```diff
@@ -357,7 +303,7 @@ pub fn init() -> Result<Init, ()> {
         }
         debug_assert!(lock != 0);
         let lock = lock as HANDLE;
-        let r = WaitForSingleObjectEx(lock, INFINITE, FALSE);
+        let r = WaitForSingleObjectEx(lock, INFINITE, 0);
```
workingjubilee, line 306: I do still prefer being able to see when something is actually only supposed to have two values.
```suggestion
        let r = WaitForSingleObjectEx(lock, INFINITE, false as BOOL);
```

src/dbghelp.rs
```diff
@@ -345,7 +291,7 @@ pub fn init() -> Result<Init, ()> {
         let mut lock = LOCK.load(SeqCst);
         if lock == 0 {
             let name = mutex_name();
-            lock = CreateMutexA(ptr::null_mut(), 0, name.as_ptr().cast::<i8>()) as usize;
+            lock = CreateMutexA(ptr::null_mut(), 0, name.as_ptr()) as usize;
```
workingjubilee, line 294:
```suggestion
            lock = CreateMutexA(ptr::null_mut(), false as BOOL, name.as_ptr()) as usize;
```

workingjubilee (commented):
Only a pair of nits.

src/dbghelp.rs
CraftSpider: Whoops! I tried to change `TRUE` -> `1` back, but I apparently missed the `FALSE` -> `0` changes. I'll fix this (though probably using the windows_sys `FALSE` constant, to be consistent with using `TRUE`)

workingjubilee:
Thanks for rebasing! Finally figured out [what busted CI](https://github.com/rust-lang/backtrace-rs/pull/641) so I'm gonna merge this through finally.

workingjubilee (approved):

kleisauke:
FWIW, this breaks the tier-3 `thumbv7a-pc-windows-msvc` and `thumbv7a-uwp-windows-msvc` targets. For example, `IMAGE_FILE_MACHINE_ARMNT` is gone now.
https://github.com/rust-lang/backtrace-rs/blob/38d49aa0395c50047ab26e418a0cb30d9955fb64/src/backtrace/dbghelp32.rs#L222

PR https://github.com/rust-lang/rust/pull/112527 could be relevant here.

workingjubilee:
Tier 3 targets are allowed to be broken. If they weren't, I wouldn't be able to even land this PR. If someone wants them to be supported, they need to start maintaining them.  https://doc.rust-lang.org/rustc/target-tier-policy.html

workingjubilee:
And the appropriate venue for this is **not** in the tail end of a PR, but a new issue.

kleisauke:
Sorry, the last time I dealt with a similar issue (see #572) the fix was straightforward, but this is somewhat non-trivial as `windows-sys` lacks platform-specific type definitions for ARM32.

That said, this is a good opportunity to remove this unsupported Microsoft target from my toolchain. If anyone fancies it, they can fix this, but it won't be me this time.

PR #635, opened 2024-06-12
Don't cast HANDLE to usize and back

CraftSpider:
Not sure if there's interest for this one - I'm aware of argument whether HANDLE really needs to be a pointer, and it's defined as `isize` for non-std mode bindgen, but it was an easy change so I figured I'd put it up for review.


workingjubilee (approved):
weird, code that treats pointers as pointers instead of a funny integer is actually simpler than code that treats it as a funny integer!

ChrisDenton:
The trouble there was using two or three different definitions of `HANDLE` in the same code. I can only assume it was written before we had `AtomicPtr` otherwise I don't understand why that wasn't used in the first place.

PR #638, opened 2024-07-08
Bump addr2line to 0.24

khuey:
According to addr2line's Cargo.toml the MSRV remains unchanged and all tests pass here.

PR #639, opened 2024-07-09
Update addr2line to 0.23.0.

afranchuk:


atouchet:
addr2line 0.24 is now out. Also note that there is another PR to update addr2line here: #638

afranchuk:
@atouchet I'll close this then. It looks like that was opened a few hours before mine.

Issue #644, opened 2024-07-24
segfault in self test on NetBSD


0-wiz-0:
As found when testing https://github.com/rust-lang/rust/issues/96914, in 0.3.73 and git head (dcd0aaa815a8f632a9febfc228cee61b4cc23bd3), `cargo test` on NetBSD dies with a segfault.
```
backtrace-rs> RUST_BACKTRACE=1 cargo test
   Compiling backtrace v0.3.73 (/disk/storage-202004/archive/foreign/backtrace-rs)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 6.45s
     Running unittests src/lib.rs (target/debug/deps/backtrace-ed5435a466558a04)

running 3 tests
test symbolize::gimli::parse_running_mmaps::check_maps_entry_parsing_64bit ... ok
test symbolize::gimli::parse_running_mmaps::check_maps_entry_parsing_32bit ... ok
test capture::tests::test_frame_conversion ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.03s

     Running tests/accuracy/main.rs (target/debug/deps/accuracy-b042c1723a62a800)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/concurrent-panics.rs (target/debug/deps/concurrent_panics-2be74b7ee972732b)
test result: ok
     Running tests/current-exe-mismatch.rs (target/debug/deps/current_exe_mismatch-e7ad37dd6f4cbaa7)
stdout:

stderr:

code: signal: 11 (SIGSEGV) (core dumped)
thread 'main' panicked at tests/current-exe-mismatch.rs:72:5:
explicit panic
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic_display
   3: core::panicking::panic_explicit
   4: current_exe_mismatch::parent::panic_cold_explicit
             at /scratch/lang/rust/work/rustc-1.79.0-src/library/core/src/panic.rs:87:13
   5: current_exe_mismatch::parent
             at ./tests/current-exe-mismatch.rs:72:5
   6: current_exe_mismatch::main
             at ./tests/current-exe-mismatch.rs:15:15
   7: core::ops::function::FnOnce::call_once
[...]

PR #645, opened 2024-07-27
Skip current-exe-mismatch on NetBSD

workingjubilee:
Fixes https://github.com/rust-lang/backtrace-rs/issues/644


ChrisDenton (approved):
Makes sense to me. It's all very silly.

PR #648, opened 2024-08-01
Add QNX NTO platform support

nyurik:
Part of adding [aarch64-unknown-nto-qnx700](https://github.com/rust-lang/rust/pull/127897) support to Rust.  This will resolve the following failing unit tests:

```
[ui] tests/ui/backtrace/backtrace.rs
[ui] tests/ui/backtrace/dylib-dep.rs
[ui] tests/ui/backtrace/line-tables-only.rs
[ui] tests/ui/backtrace/std-backtrace.rs
[ui] tests/ui/panics/issue-47429-short-backtraces.rs#legacy
[ui] tests/ui/panics/issue-47429-short-backtraces.rs#v0
[ui] tests/ui/panics/runtime-switch.rs#legacy
[ui] tests/ui/panics/runtime-switch.rs#v0
[ui] tests/ui/panics/short-ice-remove-middle-frames-2.rs
[ui] tests/ui/panics/short-ice-remove-middle-frames.rs
[ui] tests/ui/runtime/backtrace-debuginfo.rs
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ cpp_demangle = { default-features = false, version = "0.4.0", optional = true, f
 [target.'cfg(not(all(windows, target_env = "msvc", not(target_vendor = "uwp"))))'.dependencies]
 miniz_oxide = { version = "0.7.0", default-features = false }
 addr2line = { version = "0.24.0", default-features = false }
-libc = { version = "0.2.146", default-features = false }
+libc = { version = "0.2.156", default-features = false }
 
 [target.'cfg(not(all(windows, target_env = "msvc", not(target_vendor = "uwp"))))'.dependencies.object]
 version = "0.36.0"
--- a/crates/as-if-std/Cargo.toml
+++ b/crates/as-if-std/Cargo.toml
@@ -14,7 +14,7 @@ bench = false
 [dependencies]
 cfg-if = "1.0"
 rustc-demangle = "0.1.21"
-libc = { version = "0.2.146", default-features = false }
+libc = { version = "0.2.156", default-features = false }
 
 [target.'cfg(not(all(windows, target_env = "msvc", not(target_vendor = "uwp"))))'.dependencies]
 miniz_oxide = { version = "0.7.0", optional = true, default-features = false }
--- a/src/backtrace/mod.rs
+++ b/src/backtrace/mod.rs
@@ -176,7 +176,6 @@ cfg_if::cfg_if! {
                 unix,
                 not(target_os = "emscripten"),
                 not(all(target_os = "ios", target_arch = "arm")),
-                not(all(target_os = "nto", target_env = "nto70")),
             ),
             all(
                 target_env = "sgx",
--- a/src/symbolize/gimli.rs
+++ b/src/symbolize/gimli.rs
@@ -224,6 +224,7 @@ cfg_if::cfg_if! {
             target_os = "hurd",
             target_os = "openbsd",
             target_os = "netbsd",
+            target_os = "nto",
             target_os = "android",
         ),
         not(target_env = "uclibc"),


src/symbolize/gimli/libs_dl_iterate_phdr.rs
```diff
@@ -37,8 +37,10 @@ fn infer_current_exe(base_addr: usize) -> OsString {
 
 // `info` should be a valid pointers.
 // `vec` should be a valid pointer to a `std::Vec`.
+// On `QNX-NTO` platform, info is a const pointer
 unsafe extern "C" fn callback(
-    info: *mut libc::dl_phdr_info,
+    #[cfg(not(target_os = "nto"))] info: *mut libc::dl_phdr_info,
+    #[cfg(target_os = "nto")] info: *const libc::dl_phdr_info,
```
workingjubilee, line 43: ...why?

nyurik: @workingjubilee because otherwise compiler complains that the declaration in the Rust code does not match C header files.  I had to make this change in order to get it to compile, which resulted in stack traces to start working on QNX. :shrug:

workingjubilee: which compiler?

workingjubilee: rustc can't read a C header file, and the C compilers don't know how to read Rust code, so what's actually emitting the error?

nyurik: @workingjubilee I suspect `libc` crate actually runs bindgen, and this is where it breaks.  I am doing the steps I described [here](https://github.com/rust-lang/rust/pull/127897#issuecomment-2253479225), and if i revert these specific lines, I get this error while building a compiler:

```
   --> library/std/src/../../backtrace/src/symbolize/gimli/libs_dl_iterate_phdr.rs:15:36
    |
15  |         libc::dl_iterate_phdr(Some(callback), core::ptr::addr_of_mut!(ret).cast());
    |                               ---- ^^^^^^^^ types differ in mutability
    |                               |
    |                               arguments to this enum variant are incorrect
    |
    = note: expected fn pointer `unsafe extern "C" fn(*const dl_phdr_info, _, _) -> _`
                  found fn item `unsafe extern "C" fn(*mut dl_phdr_info, _, _) -> _ {callback}`
help: the type constructed contains `unsafe extern "C" fn(*mut dl_phdr_info, usize, *mut libc::c_void) -> i32 {callback}` due to the type of the argument passed
   --> library/std/src/../../backtrace/src/symbolize/gimli/libs_dl_iterate_phdr.rs:15:31
    |
15  |         libc::dl_iterate_phdr(Some(callback), core::ptr::addr_of_mut!(ret).cast());
    |                               ^^^^^--------^
    |                                    |
    |                                    this argument influences the type of `Some`
note: tuple variant defined here
   --> /home/nyurik/dev/rust/rust/library/core/src/option.rs:579:5
    |
579 |     Some(#[stable(feature = "rust1", since = "1.0.0")] T),
    |     ^^^^

For more information about this error, try `rustc --explain E0308`.
error: could not compile `std` (lib) due to 1 previous error
Build completed unsuccessfully in 0:07:01
```

nyurik: Actually no, I think it was manually created in libc:

https://github.com/rust-lang/libc/blob/0e28c864c25d2e9b0ab082947445efccef213da4/src/unix/nto/mod.rs#L3339C12-L3348

nyurik: @workingjubilee most of these bindings are created using bindgen, and then often cleaned up by hand. Bindings should be as true to the original header files as possible - if the C headers specify a constant, lets be true to that. Especially since some other structs are even different in size.

### QNX header file

```c
struct dl_phdr_info {
  #if __PTR_BITS__ == 32
    Elf32_Addr          dlpi_addr;
    const char          *dlpi_name;
    const Elf32_Phdr    *dlpi_phdr;
    Elf32_Half          dlpi_phnum;
  #elif __PTR_BITS__ == 64
    Elf64_Addr          dlpi_addr;
    const char          *dlpi_name;
    const Elf64_Phdr    *dlpi_phdr;
    Elf64_Half          dlpi_phnum;
  #else
    #error __PTR_BITS__ not set properly
  #endif
};

typedef int (*Dl_Iterate_Phdr_t) (const struct dl_phdr_info *info, size_t size, void *data);
//                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

int dl_iterate_phdr(Dl_Iterate_Phdr_t callback, void *data);
```

### `/usr/include/link.h` (Linux)

```c
struct dl_phdr_info {
    ElfW(Addr) dlpi_addr;
    const char *dlpi_name;
    const ElfW(Phdr) *dlpi_phdr;
    ElfW(Half) dlpi_phnum;

    /* Note: Following members were introduced after the first
       version of this structure was available.  Check the SIZE
       argument passed to the dl_iterate_phdr callback to determine
       whether or not each later member is available.  */
       
    // *** I deleted some irrelevant comments here

    __extension__ unsigned long long int dlpi_adds;
    __extension__ unsigned long long int dlpi_subs;
    size_t dlpi_tls_modid;
    void *dlpi_tls_data;
};
 
extern int dl_iterate_phdr (int (*__callback) (struct dl_phdr_info *,
                           size_t, void *),
                void *__data);
```

workingjubilee: A function type, but with a `*const T` changed to a `*mut T`, or vice-versa, is defined to be [ABI compatible](https://doc.rust-lang.org/std/primitive.fn.html#abi-compatibility) with its twin. It is thus valid to call `dl_iterate_phdr` as either of these declarations, on either OS, assuming one links the correct symbol:
```rust
extern "C" {
    fn nto_dl_iterate_phdr(cb: fn(*const dl_phdr_info, size_t, *mut c_void), data: *mut c_void) -> c_int;
    fn linux_dl_iterate_phdr(cb: fn(*mut dl_phdr_info, size_t, *mut c_void), data: *mut c_void) -> c_int;
}
```

The utility of the libc crate, which provides cross-platform bindings in a way that allows Rust code to compile despite being written in essentially the same way, significantly declines if one has to introduce these cfgs anyways in their own code.

nyurik: @workingjubilee I agree that `libc` with identical cross-platform declarations is a great thing to have, but there are reasons I think it is better to leave it as is:
* QNX libc support was added ~2 years ago in https://github.com/rust-lang/libc/pull/3038 by @gh-tr, @flba-eb, maintained by @jonathanpallant and @japaric
* Changing existing `libc` signatures from `fn(*const dl_phdr_info)` to `fn(*mut dl_phdr_info)` is a breaking change. We can't tell how much code will be affected due to QNX compiler licensing - most of it is probably not in public. Clearly this should not be in a `libc` patch version change.
* The next breaking `libc` - v0.3 has been worked on for a while, and this is clearly a difficult process - most PRs are still targeting libc v0.2, and get forward-ported to v0.3.  (Any thoughts on this @JohnTitor ?). So we can add a FIXME to migrate this code to v0.3 once out, if `libc` agrees to unify this API.
* QNX v8.0 is the new stable version (Rust only supports 7.1), so to add support, someone will likely run bindgen on the new headers, and then start comparing old and new declarations. Having manual exceptions with lots of extra comments will make the process far more tedious for them.
* There are likely a lot more minor platform differences like this - should `libc` be doing all normalization in the next breaking version? Or should this be an exceptional one-off just because we happened to notice it?

So in short, in order to add backtraces to Rust QNX support, we should do a breaking change to the public and long-published `libc`, possibly breaking many projects, in a way that would not be consistent with the C header files (the source of truth), only because ABI happens to be compatible?  In my opinion, `libc` should be the "dumb" but "faithful"  unsafe representation of the C header files, while cross-platform compatibility should be the job of the safe wrappers, e.g. backtrace and stdlib.

flba-eb: I'm surprised to learn that `*const T` and `*mut T` are ABI compatible, this would have saved me some work when implementing the libc and abstraction layers.
However, I'm not sure if I would have walked this way: I agree with @nyurik that the C header files are the source of truth  and that the libc crate should define the same functions with exact same signatures as their C counterpart. In my opinion, it is the job of the abstraction layers to abstract from such differences.

As for staying compatible with the 2 years old QNX introduction: I would not mind changing it in an incompatible way to e.g. match the C counterpart, i.e. to fix a discrepancy between libc-crate and libc-header. But changing it so that it is deviating sounds more like introducing a bug.

I don't think that there are more than a few dozens Rust@QNX users out there, and the number of those who are using `dl_iterate_phdr` should be near to 2 ...

joshtriplett: You're saying it would be a breaking change in libc to not match QNX, but it looks like the current code in libc matches other platforms and doesn't have this divergence, and you're changing libc to have the divergence in an effort to match QNX. In other words, as far as I can tell, this PR is making the breaking change. 

I don't think we should make this change and introduce this divergence. The function is already unsafe to call, and people need to handle `mut` on any other platform, so I don't think it adds any value to make it const here.

You can add an exception in build.rs to make it compile.

joshtriplett: I think there are likely to be approximately zero users of this API in libc on QNX, such that changing libc to use `*mut` on QNX (matching other platforms) will not in practice break anyone. And better to normalize in libc than in every user of libc.

workingjubilee: @flba-eb @nyurik I am happy to lose the argument about whether or not libc should change, but you may reconsider on seeing some of my alternatives, because they may involve `mem::transmute`. Or `as` casts, which are honestly about as bad.

You see, the issue is visible in this code sample:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7208c770e9303755778a9ac0bb6535df

Notice how `<*mut T>::cast_mut` doesn't exist? And `<*const T>::cast_const` also doesn't. And fixing this fact requires a bunch of extension code. So a transmute starts to be the easiest way out.

nyurik: @workingjubilee thx, see https://github.com/rust-lang/libc/pull/3792 - there are some CI issues with libc, but once resolved, @joshtriplett is ok to merge (and hopefully release soon). Hopefully we can resolve this soon thereafter.

One question - since this change will only be in libc 0.2.156+, this PR will have to also bump `Cargo.toml`'s minimal libc to that same version. Is this how you want to proceed with this PR?

nyurik:
@workingjubilee thx for all the feedback, libc 0.2.156 just released with the consistent backtrace on QNX.  I updated this PR to use it.

src/symbolize/gimli/libs_dl_iterate_phdr.rs
workingjubilee: Yes, bumping to that libc is fine.

nyurik:
@workingjubilee thx, fixed merge conflict, should be ok now

workingjubilee (approved):

PR #626, opened 2024-05-25
Support zstd-compressed ELF sections.

khuey:
zstd has been introduced as an alternative to zlib for the compression of debug sections.[^0] Toolchain support is widely present at this time but lack of support in backtrace is a severe limitation on using this feature in Rust programs.

This uses a Rust reimplementation of zstd (the ruzstd crate). This has the benefit of simplifying the build process, but this crate is less used and admittedly slower than the zstd crate that binds to the C libzstd.

[^0]: https://maskray.me/blog/2022-09-09-zstd-compressed-debug-sections

PR #649, opened 2024-08-02
Apply some of the `cargo clippy` suggestions

nyurik:
As @workingjubilee [mentioned](https://github.com/rust-lang/backtrace-rs/pull/626#issuecomment-2134245457), `no one is really paying for professional maintenance of backtrace-rs`, so might as well try to do some volunteer upkeep. Just to keep things tidy.

Note that the code does not seem to have had `cargo fmt` run in a while, but I don't want to do it as part of code-changing PR.

The `enum backtrace::print::PrintFmt {}` seem to be requiring `#[non_exhaustive]`, but that may result in some breaking changes?  Not certain how this should be handled, and probably in a separate PR.

### DIFF
--- a/src/capture.rs
+++ b/src/capture.rs
@@ -1,3 +1,5 @@
+#![allow(clippy::from_over_into)]
+
 #[cfg(feature = "serde")]
 use crate::resolve;
 use crate::PrintFmt;
@@ -351,7 +353,7 @@ impl From<crate::Frame> for BacktraceFrame {
     }
 }
 
-// we don't want implementing `impl From<Backtrace> for Vec<BacktraceFrame>` on purpose,
+// we don't want to implement `impl From<Backtrace> for Vec<BacktraceFrame>` on purpose,
 // because "... additional directions for Vec<T> can weaken type inference ..."
 // more information on https://github.com/rust-lang/backtrace-rs/pull/526
 impl Into<Vec<BacktraceFrame>> for Backtrace {
@@ -452,7 +454,7 @@ impl BacktraceSymbol {
     /// This function requires the `std` feature of the `backtrace` crate to be
     /// enabled, and the `std` feature is enabled by default.
     pub fn filename(&self) -> Option<&Path> {
-        self.filename.as_ref().map(|p| &**p)
+        self.filename.as_deref()
     }
 
     /// Same as `Symbol::lineno`
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -18,26 +18,24 @@
 //! Next:
 //!
 //! ```
-//! fn main() {
 //! # // Unsafe here so test passes on no_std.
 //! # #[cfg(feature = "std")] {
-//!     backtrace::trace(|frame| {
-//!         let ip = frame.ip();
-//!         let symbol_address = frame.symbol_address();
+//! backtrace::trace(|frame| {
+//!     let ip = frame.ip();
+//!     let symbol_address = frame.symbol_address();
 //!
-//!         // Resolve this instruction pointer to a symbol name
-//!         backtrace::resolve_frame(frame, |symbol| {
-//!             if let Some(name) = symbol.name() {
-//!                 // ...
-//!             }
-//!             if let Some(filename) = symbol.filename() {
-//!                 // ...
-//!             }
-//!         });
-//!
-//!         true // keep going to the next frame
+//!     // Resolve this instruction pointer to a symbol name
+//!     backtrace::resolve_frame(frame, |symbol| {
+//!         if let Some(name) = symbol.name() {
+//!             // ...
+//!         }
+//!         if let Some(filename) = symbol.filename() {
+//!             // ...
+//!         }
 //!     });
-//! }
+//!
+//!     true // keep going to the next frame
+//! });
 //! # }
 //! ```
 //!
--- a/src/print.rs
+++ b/src/print.rs
@@ -288,7 +288,7 @@ impl BacktraceFrameFmt<'_, '_, '_> {
             write!(self.fmt.fmt, ":{colno}")?;
         }
 
-        write!(self.fmt.fmt, "\n")?;
+        writeln!(self.fmt.fmt)?;
         Ok(())
     }
 
--- a/src/symbolize/gimli.rs
+++ b/src/symbolize/gimli.rs
@@ -12,7 +12,6 @@ use super::SymbolName;
 use addr2line::gimli;
 use core::convert::TryInto;
 use core::mem;
-use core::u32;
 use libc::c_void;
 use mystd::ffi::OsString;
 use mystd::fs::File;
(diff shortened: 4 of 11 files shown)


build.rs
```diff
@@ -5,9 +5,8 @@ use std::path::Path;
 
 // Must be public so the build script of `std` can call it.
 pub fn main() {
-    match env::var("CARGO_CFG_TARGET_OS").unwrap_or_default().as_str() {
-        "android" => build_android(),
-        _ => {}
+    if env::var("CARGO_CFG_TARGET_OS").unwrap_or_default().as_str() == "android" {
+        build_android()
     }
```
workingjubilee, line 10: Please do not touch this code, it is slated for deletion.

nyurik: @workingjubilee the whole crate?  Or just the `build.rs`?

nyurik: I reverted this specific instance, hoping it is only `build.rs` that's affected.

src/symbolize/gimli.rs
```diff
@@ -122,13 +121,11 @@ impl<'data> Context<'data> {
...
+            } else if let Some(name) = id.xcoff_name() {
+                let data = object.section(stash, name).unwrap_or(&[]);
+                Ok(EndianSlice::new(data, Endian))
             } else {
-                if let Some(name) = id.xcoff_name() {
-                    let data = object.section(stash, name).unwrap_or(&[]);
-                    Ok(EndianSlice::new(data, Endian))
-                } else {
-                    Ok(EndianSlice::new(&[], Endian))
-                }
```
workingjubilee, line 131: I would somewhat prefer this stay as-is because "all the AIX code in this fn is at its own indentation" is on-purpose. I didn't specifically say so it was when I accepted this code, but I picked the ["rightward drift"](https://github.com/rust-lang/backtrace-rs/pull/508#discussion_r1297899219) and other [stylistic nits](https://github.com/rust-lang/backtrace-rs/pull/508#discussion_r1297889865) [several](https://github.com/rust-lang/backtrace-rs/pull/508#discussion_r1297901492) [times](https://github.com/rust-lang/backtrace-rs/pull/508#discussion_r1297891386) and this was not an oversight.

If you would prefer refactoring this in a different way that appeases the lint and keeps the current visible divide between "AIX" and "everyone else" then that's fine too.

src/types.rs
```diff
@@ -33,9 +33,9 @@ impl<'a> BytesOrWideString<'a> {
     pub fn to_str_lossy(&self) -> Cow<'a, str> {
         use self::BytesOrWideString::*;
 
-        match self {
-            &Bytes(slice) => String::from_utf8_lossy(slice),
-            &Wide(wide) => Cow::Owned(String::from_utf16_lossy(wide)),
+        match *self {
+            Bytes(slice) => String::from_utf8_lossy(slice),
+            Wide(wide) => Cow::Owned(String::from_utf16_lossy(wide)),
```
workingjubilee, line 38: clippy really lints on that...?

build.rs
workingjubilee: the build.rs

src/capture.rs
```diff
@@ -351,9 +351,10 @@ impl From<crate::Frame> for BacktraceFrame {
     }
 }
 
-// we don't want implementing `impl From<Backtrace> for Vec<BacktraceFrame>` on purpose,
+// we don't want to implement `impl From<Backtrace> for Vec<BacktraceFrame>` on purpose,
 // because "... additional directions for Vec<T> can weaken type inference ..."
 // more information on https://github.com/rust-lang/backtrace-rs/pull/526
+#[allow(clippy::from_over_into)]
```
workingjubilee, line 357: also hoist to top.

src/lib.rs
```diff
@@ -18,26 +18,24 @@
 //! Next:
 //!
 //! ```
-//! fn main() {
 //! # // Unsafe here so test passes on no_std.
 //! # #[cfg(feature = "std")] {
```
workingjubilee, line 22: huh, I don't understand what this comment was about...

src/print.rs
```diff
@@ -289,7 +289,7 @@ impl BacktraceFrameFmt<'_, '_, '_> {
             write!(self.fmt.fmt, ":{colno}")?;
         }
 
-        write!(self.fmt.fmt, "\n")?;
+        writeln!(self.fmt.fmt)?;
```
workingjubilee, line 292: I ran `cargo test` with the newline replaced with `""` here and there do not appear to be any tests covering this code.

workingjubilee (changes requested):
...huh, I seem to be having trouble finding which tests actually would break here if the refactoring here was wrong...

nyurik:
thx for the review, will fix it shortly...

workingjubilee:
I would not consider changing PrintFmt to use the actual `#[non_exhaustive]` attribute instead of the "non-exhaustive by convention" to be a breaking change.

src/symbolize/mod.rs
```diff
@@ -346,7 +346,7 @@ fn format_symbol_name(
     mut bytes: &[u8],
     f: &mut fmt::Formatter<'_>,
 ) -> fmt::Result {
-    while bytes.len() > 0 {
+    while !bytes.is_empty() {
```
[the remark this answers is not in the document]
nyurik: ok, but please note that there are 20 instances of `.is_empty()`, and at least 4 of them use the `!foo.is_empty()`. It seems to be a fairly common pattern in this and other repos

src/types.rs
nyurik: https://rust-lang.github.io/rust-clippy/master/index.html#/match_ref_pats

nyurik:
Note that asserting stacktrace in the unit test is kinda hacky -- `skip_inner_frames` is included in non-miri output

```
# On Linux

   0: skip_inner_frames::backtrace_new_should_start_with_call_site_trace
             at /home/nyurik/dev/rust/backtrace-rs/tests/skip_inner_frames.rs:39:13
   1: skip_inner_frames::backtrace_new_should_start_with_call_site_trace::{{closure}}
             at /home/nyurik/dev/rust/backtrace-rs/tests/skip_inner_frames.rs:35:53

# MIRI

   0: backtrace_new_should_start_with_call_site_trace
             at tests/skip_inner_frames.rs:39:13
   1: backtrace_new_should_start_with_call_site_trace::{closure#0}
             at tests/skip_inner_frames.rs:35:53
...

File: src/print.rs (hunk 6 above)
Comment (workingjubilee) [reply src/print.rs:292]: Thanks.

File: src/lib.rs (hunk 5 above)
Comment (workingjubilee) [reply src/lib.rs:22]: Ah, I see.

**Comment by workingjubilee** [comment]
The test being hacky is fine, all tests are allowed to be hacky forever and ever as long as the test is correctly testing at least what it is supposed to test. However something merge conflicted with this, sorry about that.

**Comment by nyurik** [comment]
@workingjubilee fixed

File: src/symbolize/gimli.rs [hunk 8]
```diff
@@ -119,10 +118,14 @@ impl<'data> Context<'data> {
...
-            if cfg!(not(target_os = "aix")) {
+            #[cfg(not(target_os = "aix"))]
+            {
                 let data = object.section(stash, id.name()).unwrap_or(&[]);
                 Ok(EndianSlice::new(data, Endian))
-            } else if let Some(name) = id.xcoff_name() {
+            }
+
+            #[cfg(target_os = "aix")]
+            if let Some(name) = id.xcoff_name() {
```
workingjubilee, line 128: Mm, this means normal builds don't test if this part of the AIX path works tho'...

nyurik: reverted

workingjubilee (approved):

PR #650, opened 2024-08-02
Update `libloading` dev dep, run `cargo update`

nyurik:
Run

```shell
cargo +nightly -Z unstable-options update --breaking
cargo update
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -51,9 +51,12 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.0.99"
+version = "1.1.22"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "96c51067fd44124faa7f870b4b1c969379ad32b2ba805aa959430ceaa384f695"
+checksum = "9540e661f81799159abee814118cc139a2004b3a3aa3ea37724a1b66530b90e0"
+dependencies = [
+ "shlex",
+]
 
 [[package]]
 name = "cfg-if"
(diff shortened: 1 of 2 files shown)


crates/as-if-std/Cargo.toml
```diff
@@ -29,7 +29,7 @@ features = ['read_core', 'elf', 'macho', 'pe', 'xcoff', 'unaligned', 'archive']
 [build-dependencies]
 # Dependency of the `backtrace` crate
 cc = "1.0.97"
-windows-bindgen = "0.56"
+windows-bindgen = "0.58"
```
ChrisDenton, line 32: Updating `windows-bindgen` should also update the generated bindings, which involves running the `riddle` tool (`cargo install riddle`).

workingjubilee:
this doesn't mean we have added a third version of addr2line when the submodule is updated in std, does it?

workingjubilee:
can this be cut back to just the bindgen change?

workingjubilee:
merge conflicts

atouchet:
addr2line was updated in #638.

crates/as-if-std/Cargo.toml
nyurik: I removed this update in this PR, should probably be done separatelly

nyurik:
I revised this to only bump miniz_oxide and libloading crates.

ChrisDenton (approved):

nyurik:
I just rebased this PR - the only change now is the `libloading` 0.7 -> 0.8, plus some lock updates.  It builds fine locally, but CI is having some strange issues, esp with nightly?

PR #653, opened 2024-08-09
Update windows bindings and use windows-targets

ChrisDenton:
This synchronizes the API definitions with std. Trouble is, std uses its own definition of the `windows_targets::link!` macro so I've used a path hack to include it in this crate when used as a dependency of std. Otherwise the `windows-targets` crate is used.

Not sure if this will work but let's see what CI says.


ChrisDenton:
Oh, path hacks might not work because it looks like there's a difference between fake-std and real-std.

ChrisDenton:
Hmm... maybe I could make `windows_targets` crate in std's sysroot and share it that way.

ChrisDenton:
Ok, with the std changes this should be ready for review now so removing the draft thingy

workingjubilee:
merge conflicts

workingjubilee:
Remember that libstd does this for backtrace-rs:
```rust
    println!("cargo:rustc-cfg=backtrace_in_libstd");
```

Cargo.toml
```diff
@@ -38,6 +38,7 @@ serde = { version = "1.0", optional = true, features = ['derive'] }
 cpp_demangle = { default-features = false, version = "0.4.0", optional = true, features = [
   "alloc",
 ] }
+windows-targets = "0.52.6"
```
workingjubilee, line 41: shouldn't this be gated behind `cfg(windows)`?

ChrisDenton: Done!

workingjubilee (approved):
Thanks! Syncing with std is generally desirable so,

PR #656, opened 2024-08-23
remove build.rs for android

fengys1996:
Since https://github.com/rust-lang/rust/pull/120593 the minimum android API level is 21.

Related
https://github.com/android/ndk/wiki/Changelog-r26
https://github.com/rust-lang/rust/pull/129305

1. stop setting `dl_iterate_phdr` based on whether the android API level is greater than or equal to 21, but retain it for back compat.
2. remove `cc` build dependency.
3. upgrade ndk r25b to r26d in CI.


Cargo.toml
```diff
@@ -74,7 +68,6 @@ serialize-serde = ["serde"]
 # purposes. New code should use none of these features.
 coresymbolication = []
 dbghelp = []
-dl_iterate_phdr = []
```
bjorn3, line 77: This feature itself should probably remain for back compat reasons. It would just do nothing.

fengys1996: Thank you for your review, I think it makes sense.

fengys1996: Thanks for your review. 

`dl_iterate_phdr` is a dev dependency, and users will not explicitly specify it. I think it can be removed.

fengys1996:
Do we need to upgrade the `backtrace rs` version for update std?

workingjubilee:
Don't worry about the version bump, I'll take care of that in due time.

workingjubilee (approved):
Thanks!

workingjubilee:
Sorry, I don't think I saw this @MarijnS95, as I generally am busy keeping up with new PRs and don't look at PRs long after I review/merge them even if they get activity, until I eventually decide to do so for unrelated retrospective reasons.

Likewise, the MSRV changed to 1.82 for unrelated reasons afterwards.

"Invalid build"? I'm not sure, that term is vague. I think the result is a linkage error, as the code will attempt to invoke `dl_iterate_phdr` which must be available in some way. I believe it will happen at compiletime since libc will statically express the desire for the function, but it might happen during dynamic loading. It won't happen after start since we don't weakly link for it, so it won't risk calling a null symbol.

MarijnS95:
@workingjubilee it wasn't a specific request, just an open-ended question whether there is a scenario in which this could break ("invalid build" → I likely meant a build or startup error depending on how the linker is configured). If someone would search or bisect for it, they might end up here and hopefully have a use for my findings; but that fortunately never seemed to happen.

Note that I've also posted a more elaborate version of this change to https://github.com/gimli-rs/findshlibs/pull/82, whenever they bump their MSRV too.

Issue #659, opened 2024-08-29
`dl_iterate_phdr` assumes GNU/Linux


workingjubilee:
Only GNU/Linux guarantees:
- the name of the executable is `""`
- the executable is the first callback

This doesn't seem to be guaranteed to be true, even on other flavors of Linux, which may have a sense of humor, e.g. musl's choices in the static linkage case:
```c
        info.dlpi_addr  = base;
	info.dlpi_name  = "/proc/self/exe";
	info.dlpi_phdr  = (void *)aux[AT_PHDR];
	info.dlpi_phnum = aux[AT_PHNUM];
	info.dlpi_adds  = 0;
	info.dlpi_subs  = 0;
	if (tls_phdr) {
		info.dlpi_tls_modid = 1;
		info.dlpi_tls_data = __tls_get_addr((tls_mod_off_t[]){1,0});
	} else {
		info.dlpi_tls_modid = 0;
		info.dlpi_tls_data = 0;
	}
	return (callback)(&info, sizeof (info), data);
```

workingjubilee:
See also https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=199943

PR #660, opened 2024-08-29
Revise `dl_iterate_phdr` callback to be sound-ish

workingjubilee:
We
- no longer pretend the entire Unixverse is GNU/Linux
- no longer pretend static linking does not exist
- no longer dereference pointers without checking them, except for the info object itself, where it's libc's responsibility to be sound

This passed CI on locally running rustc's arm-android Docker image.

Fixes rust-lang/backtrace-rs#659.

### DIFF
--- a/src/symbolize/gimli/libs_dl_iterate_phdr.rs
+++ b/src/symbolize/gimli/libs_dl_iterate_phdr.rs
@@ -45,13 +45,10 @@ unsafe extern "C" fn callback(
     vec: *mut libc::c_void,
 ) -> libc::c_int {
     // SAFETY: We are guaranteed these fields:
-    let libc::dl_phdr_info {
-        dlpi_addr,
-        dlpi_name,
-        dlpi_phdr,
-        dlpi_phnum,
-        ..
-    } = unsafe { *info };
+    let dlpi_addr = unsafe { (*info).dlpi_addr };
+    let dlpi_name = unsafe { (*info).dlpi_name };
+    let dlpi_phdr = unsafe { (*info).dlpi_phdr };
+    let dlpi_phnum = unsafe { (*info).dlpi_phnum };
     // SAFETY: We assured this.
     let libs = unsafe { &mut *vec.cast::<Vec<Library>>() };
     // most implementations give us the main program first
--- a/src/symbolize/gimli/libs_dl_iterate_phdr.rs
+++ b/src/symbolize/gimli/libs_dl_iterate_phdr.rs
@@ -35,31 +35,53 @@ fn infer_current_exe(base_addr: usize) -> OsString {
     env::current_exe().map(|e| e.into()).unwrap_or_default()
 }
 
-// `info` should be a valid pointers.
-// `vec` should be a valid pointer to a `std::Vec`.
+/// # Safety
+/// `info` must be a valid pointer.
+/// `vec` must be a valid pointer to `Vec<Library>`
+#[forbid(unsafe_op_in_unsafe_fn)]
 unsafe extern "C" fn callback(
     info: *mut libc::dl_phdr_info,
     _size: libc::size_t,
     vec: *mut libc::c_void,
 ) -> libc::c_int {
-    let info = &*info;
-    let libs = &mut *vec.cast::<Vec<Library>>();
-    let is_main_prog = info.dlpi_name.is_null() || *info.dlpi_name == 0;
-    let name = if is_main_prog {
-        // The man page for dl_iterate_phdr says that the first object visited by
-        // callback is the main program; so the first time we encounter a
-        // nameless entry, we can assume its the main program and try to infer its path.
-        // After that, we cannot continue that assumption, and we use an empty string.
-        if libs.is_empty() {
-            infer_current_exe(info.dlpi_addr as usize)
-        } else {
-            OsString::new()
-        }
+    // SAFETY: We are guaranteed these fields:
+    let libc::dl_phdr_info {
+        dlpi_addr,
+        dlpi_name,
+        dlpi_phdr,
+        dlpi_phnum,
+        ..
+    } = unsafe { *info };
+    // SAFETY: We assured this.
+    let libs = unsafe { &mut *vec.cast::<Vec<Library>>() };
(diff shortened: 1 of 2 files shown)


src/symbolize/gimli/libs_dl_iterate_phdr.rs
```diff
@@ -35,31 +35,53 @@ fn infer_current_exe(base_addr: usize) -> OsString {
...
+    // sometimes we get a null or 0-len CStr, based on libc's whims, but these mean the same thing
+    let no_given_name = dlpi_name.is_null()
+        // SAFETY: we just checked for null
+        || unsafe { *dlpi_name == 0 };
+    let name = if is_static {
+        // don't try to look up our name from /proc/self/maps, it'll get silly
+        env::current_exe().unwrap_or_default().into_os_string()
+    } else if is_main && no_given_name {
+        infer_current_exe(dlpi_addr as usize)
+    } else {
```
ChrisDenton, line 70: Just making sure I understand but why don't we always use `dlpi_name` if it's available? I mean, if we have a name then looking it up elsewhere is redundant, no?

ChrisDenton:
I don't necessarily want to block this PR in particular since this is a comment that could apply to alot of code in this repo but... it feels like we should be making more use of `NonNull` or other abstraction for the `if ptr.is_null() { do stuff with ptr }` pattern. Because like there's nothing enforcing that the null check is done on the same pointer that's being used, Any mistakes should of course be caught by the reviewer but making it machine checked would be nicer.

src/symbolize/gimli/libs_dl_iterate_phdr.rs
```diff
@@ -35,31 +35,53 @@ fn infer_current_exe(base_addr: usize) -> OsString {
...
-    let is_main_prog = info.dlpi_name.is_null() || *info.dlpi_name == 0;
-    let name = if is_main_prog {
-        // The man page for dl_iterate_phdr says that the first object visited by
-        // callback is the main program; so the first time we encounter a
-        // nameless entry, we can assume its the main program and try to infer its path.
-        // After that, we cannot continue that assumption, and we use an empty string.
-        if libs.is_empty() {
-            infer_current_exe(info.dlpi_addr as usize)
-        } else {
+    // SAFETY: We are guaranteed these fields:
```
ChrisDenton, line 47: Hm, the unsafe bit is dereferencing the `info` pointer which requires it to point to a valid `dl_phdr_info` struct (as defined by `libc`). Which, like `vec` being `Vec<Library>` is already part of the function's safety contract.

The fact that we're using only the guaranteed fields is good to know but that particular part could as well be safe code.

workingjubilee: What safe code are you imagining, exactly? This signature must conform to the callback signature defined by `dl_iterate_phdr`.

workingjubilee: The "these fields are guaranteed" comment is because the struct is **variably-sized**, and only those fields are guaranteed to be **initialized**.

workingjubilee:
> I don't necessarily want to block this PR in particular since this is a comment that could apply to alot of code in this repo but... it feels like we should be making more use of NonNull or other abstraction for the if !ptr.is_null() { do stuff with ptr } pattern.

The particular goal here was to keep the differences with the original down to minimize the likelihood that merging this into rust-lang/rust will cause a regression that will have a very short time to root-cause and fix. I originally didn't even want to fix #659.

src/symbolize/gimli/libs_dl_iterate_phdr.rs
ChrisDenton: But the `*` dereferences the whole struct right? Which is the only unsafe part. The pattern matching is safe code after that.

workingjubilee: > dereferences

It creates a *place expression*.

But as I understand it, a pattern match only **loads** from the sites that it actually matches, thus, this:
```rust
let _ = unsafe { *info };
```
loads nothing and could actually be safe.

src/symbolize/gimli/libs_dl_iterate_phdr.rs
workingjubilee: because if it's `""` then we're probably on GNU/Linux and if we're the main binary we have to go look up the name somewhere else to make things make sense.

src/symbolize/gimli/libs_dl_iterate_phdr.rs
ChrisDenton: I want to believe but is there an example I can run in miri to confirm this? My quick attempt failed but I don't know if it's really representative https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=2c8fb800be3f138551cecda931fecec0

src/symbolize/gimli/libs_dl_iterate_phdr.rs
workingjubilee: in the `dlpi_addr == 0` case, the implementation for linux-musl at least gives you the funny "/proc/self/exe" which is basically Linux-ese for "go look it up yourself".

workingjubilee: on bionic they don't give you anything useful iirc.

workingjubilee: I was reading like 6 different libcs or libc docs last night, can you tell?

ChrisDenton: haha, yes. Fair enough then.

src/symbolize/gimli/libs_dl_iterate_phdr.rs
workingjubilee: hm, apparently it doesn't apply when pattern-matching, sigh.

workingjubilee: welp, loading four times instead i guess!

ChrisDenton (approved):
r=me when PR CI is done

Issue #661, opened 2024-09-03
Symbolication fails on Android with APK-embedded libraries


sudoBash418:
Modern Android tooling [defaults][1] to storing native libraries uncompressed and page-aligned within APKs (as opposed to the legacy behaviour where native libraries would be extracted upon app installation).
In turn, the dynamic linker [supports][2] loading these shared objects directly from an APK, which are identified by paths containing a `!/` delimiter between the name of the APK on disk and the name of the shared object within the APK (ex: `/data/app/com.example/base.apk!/lib/x86_64/mylib.so`).

[1]: https://developer.android.com/reference/tools/gradle-api/7.1/com/android/build/api/dsl/JniLibsPackagingOptions#uselegacypackaging
[2]: https://cs.android.com/search?q=open_library_in_zipfile&ss=android%2Fplatform%2Fsuperproject%2Fmain

Unfortunately, this behaviour breaks symbolication by `backtrace-rs`.

Here's the result of executing the `raw` example from an APK via the linker (supported since Android 10):
```shell
$ linker64 /data/local/tmp/android-test.apk!/embedded.so
frame #0  - 0x0000007d27046d44e9 - <no info>
frame #1  - 0x0000007d27046d4574 - <no info>
frame #2  - 0x0000007d27046d5420 - <no info>
frame #3  - 0x0000007d27046d53f6 - <no info>
frame #4  - 0x0000007d27046d53e6 - <no info>
frame #5  - 0x0000007d27046d53d6 - <no info>
frame #6  - 0x0000007d27046d53c6 - <no info>
frame #7  - 0x0000007d27046d644b - <no info>
frame #8  - 0x0000007d27046d414e - <no info>
frame #9  - 0x0000007d27046d53a1 - <no info>
frame #10 - 0x0000007d27047b3c19 - <no info>
frame #11 - 0x0000007d27046d537a - <no info>
frame #12 - 0x0000007d27046d544e - <no info>
frame #13 - 0x0000007d27012af9f0 - __libc_init
```

Executing the same binary outside an APK works as expected:

<details>
<summary>Show output</summary>

```shell
$ linker64 /data/local/tmp/embedded.so
frame #0  - 0x00000073ad9eea74e9 - backtrace::backtrace::libunwind::trace::h7981a4ba8bed542f
                                 @ /build/src/backtrace/libunwind.rs:116
                                 - backtrace::backtrace::trace_unsynchronized::h886ae3bd5307ab24
[...]

PR #567, opened 2023-09-21
Add GNU/Hurd support

sthibaul:

PR #662, opened 2024-09-03
Add support for symbolicating APK/ZIP-embedded libraries on Android

sudoBash418:
Fixes rust-lang/backtrace-rs#661

Add support for symbolicating APK/ZIP-embedded libraries by doing the following:

In `native_libraries`:
1. Each path is checked for the `!/` substring
2. If found, find the matching mapping in `/proc/self/maps` and use the `offset` field to determine where the library starts in the archive

In `Mapping::new_android` (non-embedded libraries use the same logic as before):
1. No attempts to find separate debug info files are made
2. Libraries are mapped starting from the offset from `native_libraries` and ending at the end of the APK.
(the cost of mmap-ing the unused remainder of the APK should be minimal)

---

Aside: my previous attempt used the `zip` crate to determine the exact size of the library and avoid mmap-ing the rest of the archive. But I figured it wasn't worth bringing in a new dependency and bloating the example binary by ~25% 😅

As for potential confusion between a file literally named `/path/to/apk!/lib.so` and a file named `lib.so` within `/path/to/apk`, I tried to follow the dynamic linker's behaviour: treat it as an embedded file first, and fallback to a literal path if mapping fails for any reason.

Tests have been skipped because there is no CI that runs the emulated binaries anyways.

### DIFF
--- a/src/symbolize/gimli.rs
+++ b/src/symbolize/gimli.rs
@@ -186,7 +186,7 @@ impl<'data> Context<'data> {
 fn mmap(path: &Path) -> Option<Mmap> {
     let file = File::open(path).ok()?;
     let len = file.metadata().ok()?.len().try_into().ok()?;
-    unsafe { Mmap::map(&file, len) }
+    unsafe { Mmap::map(&file, len, 0) }
 }
 
 cfg_if::cfg_if! {
@@ -268,6 +268,21 @@ struct Cache {
 
 struct Library {
     name: OsString,
+    #[cfg(target_os = "android")]
+    /// On Android, the dynamic linker [can map libraries directly from a
+    /// ZIP archive][ndk-linker-changes] (typically an `.apk`).
+    ///
+    /// The linker requires that these libraries are stored uncompressed
+    /// and page-aligned.
+    ///
+    /// These "embedded" libraries have filepaths of the form
+    /// `/path/to/my.apk!/lib/mylib.so` (where `/path/to/my.apk` is the archive
+    /// and `lib/mylib.so` is the name of the library within the archive).
+    ///
+    /// This mechanism is present on Android since API level 23.
+    ///
+    /// [ndk-linker-changes]: https://android.googlesource.com/platform/bionic/+/main/android-changes-for-ndk-developers.md#opening-shared-libraries-directly-from-an-apk
+    zip_offset: Option<u64>,
     #[cfg(target_os = "aix")]
     /// On AIX, the library mmapped can be a member of a big-archive file.
     /// For example, with a big-archive named libfoo.a containing libbar.so,
@@ -294,17 +309,31 @@ struct LibrarySegment {
     len: usize,
 }
 
-#[cfg(target_os = "aix")]
 fn create_mapping(lib: &Library) -> Option<Mapping> {
-    let name = &lib.name;
-    let member_name = &lib.member_name;
-    Mapping::new(name.as_ref(), member_name)
+    cfg_if::cfg_if! {
+        if #[cfg(target_os = "aix")] {
+            Mapping::new(lib.name.as_ref(), &lib.member_name)
+        } else if #[cfg(target_os = "android")] {
+            Mapping::new_android(lib.name.as_ref(), lib.zip_offset)
+        } else {
+            Mapping::new(lib.name.as_ref())
+        }
+    }
 }
 
-#[cfg(not(target_os = "aix"))]
-fn create_mapping(lib: &Library) -> Option<Mapping> {
-    let name = &lib.name;
-    Mapping::new(name.as_ref())
+/// Try to extract the archive path from an "embedded" library path
+/// (e.g. `/path/to/my.apk` from `/path/to/my.apk!/mylib.so`).
+///
+/// Returns `None` if the path does not contain a `!/` separator.
+#[cfg(target_os = "android")]
+fn extract_zip_path_android(path: &mystd::ffi::OsStr) -> Option<&mystd::ffi::OsStr> {
+    use mystd::os::unix::ffi::OsStrExt;
+
+    path.as_bytes()
+        .windows(2)
+        .enumerate()
+        .find(|(_, chunk)| chunk == b"!/")
+        .map(|(index, _)| mystd::ffi::OsStr::from_bytes(path.as_bytes().split_at(index).0))
 }
 
 // unsafe because this is required to be externally synchronized
--- a/src/symbolize/gimli/elf.rs
+++ b/src/symbolize/gimli/elf.rs
@@ -45,6 +45,47 @@ impl Mapping {
         })
     }
 
+    /// On Android, shared objects can be loaded directly from a ZIP archive
+    /// (see: [`super::Library::zip_offset`]).
+    ///
+    /// If `zip_offset` is not None, we interpret the `path` as an
+    /// "embedded" library path, and the value of `zip_offset` tells us where
+    /// in the ZIP archive the library data starts.
+    ///
+    /// We expect `zip_offset` to be page-aligned because the dynamic linker
+    /// requires this. Otherwise, loading the embedded library will fail.
+    ///
+    /// If we fail to load an embedded library for any reason, we fallback to
+    /// interpreting the path as a literal file on disk (same as calling [`Self::new`]).
+    #[cfg(target_os = "android")]
+    pub fn new_android(path: &Path, zip_offset: Option<u64>) -> Option<Mapping> {
+        fn map_embedded_library(path: &Path, zip_offset: u64) -> Option<Mapping> {
+            // get path of ZIP archive (delimited by `!/`)
+            let zip_path = Path::new(super::extract_zip_path_android(path.as_os_str())?);
+
+            let file = fs::File::open(zip_path).ok()?;
+            let len = file.metadata().ok()?.len();
+
+            // NOTE: we map the remainder of the entire archive instead of just the library so we don't have to determine its length
+            // NOTE: mmap will fail if `zip_offset` is not page-aligned
+            let map = unsafe {
+                super::mmap::Mmap::map(&file, usize::try_from(len - zip_offset).ok()?, zip_offset)
+            }?;
+
+            Mapping::mk(map, |map, stash| {
+                Context::new(stash, Object::parse(&map)?, None, None)
+            })
+        }
+
+        // if ZIP offset is given, try mapping as a ZIP-embedded library
+        // otherwise, fallback to mapping as a literal filepath
+        if let Some(zip_offset) = zip_offset {
+            map_embedded_library(path, zip_offset).or_else(|| Self::new(path))
+        } else {
+            Self::new(path)
+        }
+    }
+
     /// Load debuginfo from an external debug file.
     fn new_debug(original_path: &Path, path: PathBuf, crc: Option<u32>) -> Option<Mapping> {
         let map = super::mmap(&path)?;
--- a/src/symbolize/gimli/libs_dl_iterate_phdr.rs
+++ b/src/symbolize/gimli/libs_dl_iterate_phdr.rs
@@ -6,43 +6,54 @@ use super::mystd::borrow::ToOwned;
 use super::mystd::env;
 use super::mystd::ffi::{CStr, OsStr};
 use super::mystd::os::unix::prelude::*;
-use super::{Library, LibrarySegment, OsString, Vec};
+use super::{parse_running_mmaps, Library, LibrarySegment, OsString, Vec};
 use core::slice;
 
+struct CallbackData {
+    libs: Vec<Library>,
+    maps: Option<Vec<parse_running_mmaps::MapsEntry>>,
+}
 pub(super) fn native_libraries() -> Vec<Library> {
-    let mut ret = Vec::new();
+    let mut cb_data = CallbackData {
+        libs: Vec::new(),
+        #[cfg(not(target_os = "hurd"))]
+        maps: parse_running_mmaps::parse_maps().ok(),
+        #[cfg(target_os = "hurd")]
+        maps: None,
+    };
     unsafe {
-        libc::dl_iterate_phdr(Some(callback), core::ptr::addr_of_mut!(ret).cast());
+        libc::dl_iterate_phdr(Some(callback), core::ptr::addr_of_mut!(cb_data).cast());
     }
-    ret
+    cb_data.libs
 }
 
-fn infer_current_exe(base_addr: usize) -> OsString {
-    cfg_if::cfg_if! {
-        if #[cfg(not(target_os = "hurd"))] {
-                if let Ok(entries) = super::parse_running_mmaps::parse_maps() {
-                let opt_path = entries
-                    .iter()
-                    .find(|e| e.ip_matches(base_addr) && e.pathname().len() > 0)
-                    .map(|e| e.pathname())
-                    .cloned();
-                if let Some(path) = opt_path {
-                    return path;
-                }
-            }
+fn infer_current_exe(
+    maps: &Option<Vec<parse_running_mmaps::MapsEntry>>,
+    base_addr: usize,
+) -> OsString {
+    #[cfg(not(target_os = "hurd"))]
+    if let Some(entries) = maps {
+        let opt_path = entries
+            .iter()
+            .find(|e| e.ip_matches(base_addr) && e.pathname().len() > 0)
+            .map(|e| e.pathname())
+            .cloned();
+        if let Some(path) = opt_path {
+            return path;
         }
     }
+
     env::current_exe().map(|e| e.into()).unwrap_or_default()
 }
 
 /// # Safety
 /// `info` must be a valid pointer.
-/// `vec` must be a valid pointer to `Vec<Library>`
+/// `data` must be a valid pointer to `CallbackData`.
 #[forbid(unsafe_op_in_unsafe_fn)]
 unsafe extern "C" fn callback(
     info: *mut libc::dl_phdr_info,
     _size: libc::size_t,
-    vec: *mut libc::c_void,
+    data: *mut libc::c_void,
 ) -> libc::c_int {
     // SAFETY: We are guaranteed these fields:
     let dlpi_addr = unsafe { (*info).dlpi_addr };
@@ -50,7 +61,7 @@ unsafe extern "C" fn callback(
     let dlpi_phdr = unsafe { (*info).dlpi_phdr };
     let dlpi_phnum = unsafe { (*info).dlpi_phnum };
     // SAFETY: We assured this.
-    let libs = unsafe { &mut *vec.cast::<Vec<Library>>() };
+    let CallbackData { libs, maps } = unsafe { &mut *data.cast::<CallbackData>() };
     // most implementations give us the main program first
     let is_main = libs.is_empty();
     // we may be statically linked, which means we are main and mostly one big blob of code
@@ -63,7 +74,7 @@ unsafe extern "C" fn callback(
         // don't try to look up our name from /proc/self/maps, it'll get silly
         env::current_exe().unwrap_or_default().into_os_string()
     } else if is_main && no_given_name {
-        infer_current_exe(dlpi_addr as usize)
+        infer_current_exe(&maps, dlpi_addr as usize)
     } else {
         // this fallback works even if we are main, because some platforms give the name anyways
         if dlpi_name.is_null() {
@@ -73,6 +84,19 @@ unsafe extern "C" fn callback(
             OsStr::from_bytes(unsafe { CStr::from_ptr(dlpi_name) }.to_bytes()).to_owned()
         }
     };
+    #[cfg(target_os = "android")]
+    let zip_offset: Option<u64> = {
+        // only check for ZIP-embedded file if we have data from /proc/self/maps
+        maps.as_ref().and_then(|maps| {
+            // check if file is embedded within a ZIP archive by searching for `!/`
+            super::extract_zip_path_android(&name).and_then(|_| {
+                // find MapsEntry matching library's base address and get its file offset
+                maps.iter()
+                    .find(|m| m.ip_matches(dlpi_addr as usize))
+                    .map(|m| m.offset())
+            })
+        })
+    };
     let headers = if dlpi_phdr.is_null() || dlpi_phnum == 0 {
         &[]
     } else {
@@ -81,6 +105,8 @@ unsafe extern "C" fn callback(
     };
     libs.push(Library {
         name,
+        #[cfg(target_os = "android")]
+        zip_offset,
         segments: headers
             .iter()
             .map(|header| LibrarySegment {
--- a/src/symbolize/gimli/mmap_fake.rs
+++ b/src/symbolize/gimli/mmap_fake.rs
@@ -1,4 +1,5 @@
-use super::{mystd::io::Read, File};
+use super::mystd::io::{Read, Seek, SeekFrom};
+use super::File;
 use alloc::vec::Vec;
 use core::ops::Deref;
 
@@ -7,10 +8,11 @@ pub struct Mmap {
 }
 
 impl Mmap {
-    pub unsafe fn map(mut file: &File, len: usize) -> Option<Mmap> {
+    pub unsafe fn map(mut file: &File, len: usize, offset: u64) -> Option<Mmap> {
         let mut mmap = Mmap {
             vec: Vec::with_capacity(len),
         };
+        file.seek(SeekFrom::Start(offset));
         file.read_to_end(&mut mmap.vec).ok()?;
         Some(mmap)
     }
--- a/src/symbolize/gimli/mmap_unix.rs
+++ b/src/symbolize/gimli/mmap_unix.rs
@@ -15,14 +15,14 @@ pub struct Mmap {
 }
 
 impl Mmap {
-    pub unsafe fn map(file: &File, len: usize) -> Option<Mmap> {
+    pub unsafe fn map(file: &File, len: usize, offset: u64) -> Option<Mmap> {
         let ptr = mmap64(
             ptr::null_mut(),
             len,
             libc::PROT_READ,
             libc::MAP_PRIVATE,
             file.as_raw_fd(),
-            0,
+            offset.try_into().ok()?,
         );
         if ptr == libc::MAP_FAILED {
             return None;
--- a/src/symbolize/gimli/parse_running_mmaps_unix.rs
+++ b/src/symbolize/gimli/parse_running_mmaps_unix.rs
@@ -20,7 +20,7 @@ pub(super) struct MapsEntry {
     /// p = private (copy on write)
     perms: [char; 4],
     /// Offset into the file (or "whatever").
-    offset: usize,
+    offset: u64,
     /// device (major, minor)
     dev: (usize, usize),
     /// inode on the device. 0 indicates that no inode is associated with the memory region (e.g. uninitalized data aka BSS).
@@ -76,6 +76,11 @@ impl MapsEntry {
     pub(super) fn ip_matches(&self, ip: usize) -> bool {
         self.address.0 <= ip && ip < self.address.1
     }
+
+    #[cfg(target_os = "android")]
+    pub(super) fn offset(&self) -> u64 {
+        self.offset
+    }
 }
 
 impl FromStr for MapsEntry {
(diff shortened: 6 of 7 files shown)


sudoBash418:
The remaining macOS CI failure seems unrelated to this PR.

Not sure why the binary size check reports an increase on Ubuntu; AFAIK my changes should have minimal impact on non-Android targets.

ChrisDenton:
The macos failure is caused by a new lint in the nightly compiler. #664 fixes it.

src/symbolize/gimli/mmap_unix.rs
```diff
@@ -29,6 +29,22 @@ impl Mmap {
         }
         Some(Mmap { ptr, len })
     }
+
+    #[cfg(target_os = "android")]
+    pub unsafe fn map_with_offset(file: &File, len: usize, offset: usize) -> Option<Mmap> {
```
workingjubilee, line 34: Change `Mmap::map` instead. Alter callers to pass the 0 offset, and make the offset a 64-bit integer, not a usize, like [Seek](https://doc.rust-lang.org/std/io/trait.Seek.html) uses.

workingjubilee (changes requested):
Please.

src/symbolize/gimli/elf.rs
```diff
@@ -43,6 +43,44 @@ impl Mapping {
...
     }
 
+    /// On Android, shared objects can be loaded directly from a
+    /// ZIP archive. For example, an app may load a library from
+    /// `/data/app/com.example/base.apk!/lib/x86_64/mylib.so`
+    ///
+    /// For one of these "ZIP-embedded" libraries, `zip_offset` will be
+    /// non-zero (see [super::libs_dl_iterate_phdr]).
+    #[cfg(target_os = "android")]
+    pub fn new_android(path: &Path, zip_offset: usize) -> Option<Mapping> {
```
workingjubilee, line 53: Unless the zip offset is not a file offset, my comments about Mmap apply here as well. These are private types, despite the `pub` annotation.

sudoBash418: Just to clarify, do you want me to merge the two methods, or just change the type of `zip_offset`?

src/symbolize/gimli/mmap_unix.rs
sudoBash418: I can definitely do that; my original reason for creating a separate method was to avoid modifying the fake + Windows mmap implementations to match.

src/symbolize/gimli/elf.rs
workingjubilee: both. specifically it's not clear to me why we should have `new_android` instead of just "`Mapping::new` takes an offset now".

src/symbolize/gimli/mmap_unix.rs
workingjubilee: It's alright if you leave the extra parameter unused and leave a FIXME behind for those.

src/symbolize/gimli/elf.rs
sudoBash418: My reasoning was that the offset-handling code is Android-specific: the offset being non-zero signals that the file path should be treated as an archive on Android, and the offset wouldn't be used on any other platform.

If the methods are to be merged, I'm not sure what the expected behaviour should be if the offset is non-zero on other platforms - just ignore it?

workingjubilee: Oh...so zero is a magic number?

Is it literally impossible for us to want to map a zip file from offset 0?

workingjubilee: Especially considering the possibility of 16KiB pages?

sudoBash418: As far as I understand, not really? [zipalign](https://developer.android.com/tools/zipalign) will align these libraries such that the start of each file is page-aligned, and the dynamic linker will fail to load the library if it's not aligned correctly. And given that the ZIP header takes up some space at the start of the archive, I don't see how an embedded library on Android would have an offset of 0.

That said, I can wrap the offset in an `Option` if you'd prefer? (if for nothing else than clarity)

workingjubilee: Ah, zipalign was the missing piece of the puzzle for me, thank you. "Someone decided to shove their binary and its debuginfo inside a compressed archive" is just a Tuesday, but the specifics beyond that vary enough from case to case I didn't want to assume.

Yes, I would prefer the Option approach, and for Mapping it can remain under a `cfg` and leaving the other signatures intact.

I still would prefer Mmap was uniform though.

sudoBash418:
Rebased and added requested changes.

Also updated the `dl_iterate_phdr` safety doc + parameter name, which I missed earlier.

sudoBash418:
Looks like the binary size check is broken by the dependency bump I just rebased on, whoops.

src/symbolize/gimli/mmap_unix.rs
sudoBash418: I ended up implementing them anyway, didn't seem too difficult (besides Windows's more-than-page-size alignment requirements which tripped up my testing lol).

workingjubilee:
That's fine. It really ought to be fixed but whatever.

src/symbolize/gimli/elf.rs
```diff
@@ -43,6 +43,45 @@ impl Mapping {
...
+            let zip_path = raw_path
+                .windows(2)
+                .enumerate()
+                .find(|(_, chunk)| chunk == b"!/")
+                .map(|(index, _)| Path::new(OsStr::from_bytes(raw_path.split_at(index).0)))?;
+
+            let file = fs::File::open(zip_path).ok()?;
+            let len: usize = file.metadata().ok()?.len().try_into().ok()?;
+
+            // NOTE: we map the remainder of the entire archive instead of just the library so we don't have to determine its length
```
workingjubilee, line 66: and we're assuming this is okay since we don't prefault things, right?

src/symbolize/gimli/elf.rs
```diff
@@ -43,6 +43,45 @@ impl Mapping {
...
+            // NOTE: mmap will fail if `zip_offset` is not page-aligned
```
workingjubilee, line 67: invariants on mmap should be documented on the mmap function, no?

src/symbolize/gimli/elf.rs
```diff
@@ -43,6 +43,45 @@ impl Mapping {
         })
     }
 
+    /// On Android, shared objects can be loaded directly from a
+    /// ZIP archive. For example, an app may load a library from
+    /// `/data/app/com.example/base.apk!/lib/x86_64/mylib.so`
+    ///
+    /// `zip_offset` should be page-aligned; the dynamic linker
+    /// requires this when it loads libraries.
```
workingjubilee, line 51:
```suggestion
    /// `zip_offset` must be page-aligned if not None; the dynamic linker
    /// requires this when it loads libraries.
```

src/symbolize/gimli/elf.rs
```diff
@@ -43,6 +43,45 @@ impl Mapping {
...
+            // get path of ZIP archive (delimited by `!/`)
+            let raw_path = path.as_os_str().as_bytes();
+            let zip_path = raw_path
+                .windows(2)
+                .enumerate()
+                .find(|(_, chunk)| chunk == b"!/")
+                .map(|(index, _)| Path::new(OsStr::from_bytes(raw_path.split_at(index).0)))?;
+
+            let file = fs::File::open(zip_path).ok()?;
+            let len: usize = file.metadata().ok()?.len().try_into().ok()?;
```
workingjubilee, line 64: my last two brain cells: https://www.youtube.com/watch?v=h_D3VFfhvs4

src/symbolize/gimli/mmap_unix.rs
```diff
@@ -15,14 +15,14 @@ pub struct Mmap {
 }
 
 impl Mmap {
-    pub unsafe fn map(file: &File, len: usize) -> Option<Mmap> {
+    pub unsafe fn map(file: &File, len: usize, offset: i64) -> Option<Mmap> {
```
workingjubilee, line 18: hmm, maybe offset should be a u64? I'm not seeing how an negative offset is ever valid? I know it's an `off_t` in the API but like... we're not going to be passing a value greater than `i64::MAX` in.

src/symbolize/gimli/mmap_fake.rs
```diff
@@ -7,10 +7,11 @@ pub struct Mmap {
 }
 
 impl Mmap {
-    pub unsafe fn map(mut file: &File, len: usize) -> Option<Mmap> {
+    pub unsafe fn map(mut file: &File, len: usize, offset: i64) -> Option<Mmap> {
         let mut mmap = Mmap {
             vec: Vec::with_capacity(len),
         };
+        file.seek(SeekFrom::Start(offset.try_into().ok()?));
```
workingjubilee, line 14: mostly, using a u64 would remove this weird force

src/symbolize/gimli/elf.rs
```diff
@@ -43,6 +43,45 @@ impl Mapping {
...
+    #[cfg(target_os = "android")]
+    pub fn new_android(path: &Path, zip_offset: Option<i64>) -> Option<Mapping> {
+        fn map_embedded_library(path: &Path, zip_offset: i64) -> Option<Mapping> {
+            // get path of ZIP archive (delimited by `!/`)
+            let raw_path = path.as_os_str().as_bytes();
+            let zip_path = raw_path
+                .windows(2)
+                .enumerate()
+                .find(|(_, chunk)| chunk == b"!/")
+                .map(|(index, _)| Path::new(OsStr::from_bytes(raw_path.split_at(index).0)))?;
```
workingjubilee, line 61: ...so... does Android actually reject including `!` as the final character in a filename on the OS level? Is there no way to create such a file?

src/symbolize/gimli/libs_dl_iterate_phdr.rs
```diff
@@ -73,6 +86,22 @@ unsafe extern "C" fn callback(
...
         }
     };
+    #[cfg(target_os = "android")]
+    let zip_offset: Option<i64> = {
+        // only check for ZIP-embedded file if we have data from /proc/self/maps
+        maps.as_ref().and_then(|maps| {
+            // check if file is embedded within a ZIP archive by searching for `!/`
+            name.as_bytes()
+                .windows(2)
+                .find(|&chunk| chunk == b"!/")
```
workingjubilee, line 96: Hm. We obtain this information twice. First we locate it here, then we go through `create_mapping` and then into `new_android`. I suppose it's probably less of a hassle to simply compute it twice rather than try to cart around another usize offset into something, and potentially screw THAT up.

Still, I feel like, given questions like "can files have multiple ! in them on Android?", the question arises whether this sort of "split on the `!/`" shouldn't be extracted into its own logical function?

workingjubilee (changes requested):
One last(?) fussing about types, and most of the rest is just code and logic organization, and documenting the invariants correctly so as to make auditing and refactoring easy. This is an... oddly-organized library that has grown over quite a long time, before I even started programming, nevermind learned Rust. A significant amount of effort necessarily goes into just keeping this overgrown garden from finally bursting out of its walls, and the current organization doesn't really help you in that. You have my sympathies.

src/symbolize/gimli.rs
```diff
@@ -269,6 +269,8 @@ struct Cache {
 
 struct Library {
     name: OsString,
+    #[cfg(target_os = "android")]
+    zip_offset: Option<i64>,
```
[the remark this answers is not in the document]
sudoBash418: Yep will do, and I finally found [some useful documentation][1] of this behaviour, so I'll add a link to that as well

[1]: https://android.googlesource.com/platform/bionic/+/master/android-changes-for-ndk-developers.md#opening-shared-libraries-directly-from-an-apk

src/symbolize/gimli/parse_running_mmaps_unix.rs
```diff
@@ -76,6 +76,11 @@ impl MapsEntry {
     pub(super) fn ip_matches(&self, ip: usize) -> bool {
         self.address.0 <= ip && ip < self.address.1
     }
+
+    #[cfg(target_os = "android")]
+    pub(super) fn offset(&self) -> usize {
+        self.offset
+    }
```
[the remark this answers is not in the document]
sudoBash418: Good catch, thanks! I didn't think about it and just exposed a getter for the existing field, but it should probably use a `u64` instead, for exactly the scenario you described.

> can this entire PR only affect 64-bit Android?

Not without "artificially" leaving 32-bit Android unsupported; even the latest version of Android still supports 32-bit-only devices, and the embedded-library mechanism has been present in both 32-bit and 64-bit Android devices since Android 6 (API level 23), ~9 years ago.

src/symbolize/gimli/elf.rs
sudoBash418: Yes (AFAIK it should be safe regardless)

src/symbolize/gimli/elf.rs
sudoBash418: Fair point, though the exact requirements are platform-dependent.

Linux requires page-alignment (at least on the [architectures][linux-x86-64] I [looked at][linux-arm64]), [Windows][win-mmap] seems to require alignment to a larger number (64K on the 4K system I checked), and I have no idea what other platforms' requirements may be.

Honestly I assumed that callers were responsible for making sure their usage of `mmap` met any platform-specific requirements, but that's not explicitly documented anywhere.

I could move the Android-specific note to the function, or just add a statement to the function that "callers must ensure that any platform-specific requirements are upheld"?

---

I should also note that, at least on Android (and I *think* Linux in general), a call with an unaligned offset is defined behaviour: the call returns `-1`, which the `mmap` wrapper translates to `None`.

Hence my decision to just call mmap and let it fail if the offset is unaligned, rather than retrieve the system's page size and ensure alignment beforehand.

[linux-x86-64]: https://elixir.bootlin.com/linux/v6.10.8/source/arch/x86/kernel/sys_x86_64.c#L83
[linux-arm64]: https://elixir.bootlin.com/linux/v6.10.8/source/arch/arm64/kernel/sys.c#L25
[win-mmap]: https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-mapviewoffile

src/symbolize/gimli/elf.rs
sudoBash418: It does not prevent that - you can even have both paths existing simultaneously (e.g. `/path/to/my!/file` as a file within a directory at `/path/to/my!` *and* as a library in an archive named `/path/to/my`).

Maybe I should document this more clearly, but that's why the code first tries loading from an archive, and then falls back to trying to load the path literally.
The embedded path takes precedence to match [the behaviour of Android's dynamic linker](https://cs.android.com/android/platform/superproject/main/+/daa7e5ca6b9d2ae543e60be31502c302d08304e1:bionic/linker/linker.cpp;l=998), which has the same fallback behaviour.

src/symbolize/gimli/parse_running_mmaps_unix.rs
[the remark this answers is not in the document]
workingjubilee: hopefully no one is creating 5GB APKs but, well, I've met programmers before

src/symbolize/gimli/elf.rs
sudoBash418: Just to clarify: this is not a hard requirement for safety or avoiding a crash (ref: [my other comment](https://github.com/rust-lang/backtrace-rs/pull/662#discussion_r1749037840)).

If `zip_offset` is unaligned, the `mmap` call will fail and `new_android` will give up on loading the library from the archive.

src/symbolize/gimli/libs_dl_iterate_phdr.rs
sudoBash418: > probably less of a hassle to simply compute it twice rather than try to cart around another usize offset into something, and potentially screw THAT up

Yeah that was my logic.

> the question arises whether this sort of "split on the !/" shouldn't be extracted into its own logical function?

I don't have a problem with that, I'm just not sure where I'd put it. `src/symbolize/gimli.rs` maybe, or make a new sub-module?

src/symbolize/gimli/parse_running_mmaps_unix.rs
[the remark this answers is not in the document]
sudoBash418: (IIRC I read somewhere that APKs over 2GB are disallowed, but that's not really something we should rely on anyways)

src/symbolize/gimli/libs_dl_iterate_phdr.rs
```diff
@@ -73,6 +86,22 @@ unsafe extern "C" fn callback(
...
+                .and_then(|_| {
+                    // find MapsEntry matching library's base address
+                    maps.iter()
+                        .find(|m| m.ip_matches(dlpi_addr as usize))
+                        .and_then(|m| m.offset().try_into().ok())
```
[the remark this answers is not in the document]
sudoBash418: This is to find the "physical" offset of the library within the archive file.

`dl_iterate_phdr` gives us the base address of library in memory, and then we check `/proc/self/maps` for the mapping that matches that address, and the mapping tells us what the actual file offset is.

src/symbolize/gimli/mmap_unix.rs
sudoBash418: Yeah I agree

src/symbolize/gimli/elf.rs
sudoBash418: Just saw https://github.com/rust-lang/backtrace-rs/pull/667 - I suppose if the function will get some general documentation anyway, it might as well include a couple of notes to the effect of "on Android, (explain offset behaviour)" and "on other platforms, offset may have other requirements for safety or to avoid crashes"

src/symbolize/gimli/libs_dl_iterate_phdr.rs
```diff
@@ -37,20 +46,24 @@ fn infer_current_exe(base_addr: usize) -> OsString {
...
     let dlpi_name = unsafe { (*info).dlpi_name };
     let dlpi_phdr = unsafe { (*info).dlpi_phdr };
     let dlpi_phnum = unsafe { (*info).dlpi_phnum };
     // SAFETY: We assured this.
-    let libs = unsafe { &mut *vec.cast::<Vec<Library>>() };
+    let CallbackData {
+        ret: libs,
+        #[cfg(target_os = "android")]
+        maps,
+    } = unsafe { &mut *data.cast::<CallbackData>() };
```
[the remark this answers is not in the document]
sudoBash418: That makes sense to me; I'll try making it unconditional and we'll see if CI is happy.

src/symbolize/gimli/elf.rs
```diff
@@ -43,6 +43,45 @@ impl Mapping {
...
 
+    /// On Android, shared objects can be loaded directly from a
+    /// ZIP archive. For example, an app may load a library from
+    /// `/data/app/com.example/base.apk!/lib/x86_64/mylib.so`
+    ///
+    /// `zip_offset` should be page-aligned; the dynamic linker
+    /// requires this when it loads libraries.
+    #[cfg(target_os = "android")]
+    pub fn new_android(path: &Path, zip_offset: Option<i64>) -> Option<Mapping> {
+        fn map_embedded_library(path: &Path, zip_offset: i64) -> Option<Mapping> {
```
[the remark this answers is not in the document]
sudoBash418: Hm, `cargo fmt` doesn't like the blank line :/

src/symbolize/gimli/libs_dl_iterate_phdr.rs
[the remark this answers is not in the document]
sudoBash418: Unfortunately there's still the `cfg(target_os = "hurd")` case, [introduced](https://github.com/rust-lang/backtrace-rs/pull/567/files#diff-94c024636a34c29bad972edeabe708c9fb5d7118f6f34d77d5eed2a21aaf4540) by #567.

I kept the same behaviour of not parsing `/proc/self/maps` at all on Hurd, assuming that might be load bearing somehow.

src/symbolize/gimli/elf.rs
sudoBash418: I reworded the documentation to hopefully explain this better; let me know if it still needs changes

src/symbolize/gimli/libs_dl_iterate_phdr.rs
workingjubilee: Yeah, putting it in gimli.rs seems fine.

workingjubilee:
@sudoBash418 The reason that Windows demands you align the mapping to 64KiB is because that is the value of dwAllocationGranularity... a sort of virtual page size? Except not really. dwAllocationGranularity means that you lock down that much *address space*, but the actual *pages* are still faulted one by one within that region.

src/symbolize/gimli/elf.rs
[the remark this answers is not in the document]
workingjubilee: ...oh.

Well, okay, I'll deal.

src/symbolize/gimli/elf.rs
workingjubilee: Yes, that it's safe (though very easy to cause a *failure*, but a defined one) is a feature of [POSIX mmap](https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/functions/mmap.html).

It's just all the accesses after that which are racy as fuck.

src/symbolize/gimli/libs_dl_iterate_phdr.rs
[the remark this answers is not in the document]
workingjubilee: They return a memory map that doesn't conform to the pattern we are looking for.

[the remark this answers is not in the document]
workingjubilee: We had a remarkable argument about whether or not the path in a memory map must be an absolute path, or if some other OS might invent another schema. I argued that overcalibrating on how paths look on Linux is incorrect, because other Unix-y OS that otherwise copy Linux-like APIs inevitably will make up new definitions of what a "path" looks like.

I feel Android is pretty much proving me correct here (in that I cannot trust simply treating these things as "paths" to make sense).

src/symbolize/gimli.rs
```diff
@@ -295,17 +310,31 @@ struct LibrarySegment {
...
+///
+/// Returns `None` if the path does not contain a `!/` separator.
+#[cfg(target_os = "android")]
+fn extract_zip_path_android(path: &mystd::ffi::OsStr) -> Option<&mystd::ffi::OsStr> {
+    use mystd::os::unix::ffi::OsStrExt;
+
+    path.as_bytes()
+        .windows(2)
+        .enumerate()
+        .find(|(_, chunk)| chunk == b"!/")
```
workingjubilee, line 336: So we had a discussion about how the path can include multiple instances of `!/` due to both having an APK embedded .so and also just a directory with an `~/excited!` name.

...Because of this, it could even have... three.
However, if I understand the path structure and APK structure correctly, the `!/` that splits the embedded library must always be the *last* one... right?

...right?

...uh, if I'm right (I hope I am...) then we should probably use `.rev()` before we `.find()`?

sudoBash418: Actually I think it uses the first one: https://cs.android.com/android/platform/superproject/main/+/daa7e5ca6b9d2ae543e60be31502c302d08304e1:bionic/linker/linker.cpp;l=932

```cpp
  // Treat an '!/' separator inside a path as the separator between the name
  // of the zip file on disk and the subdirectory to search within it.
  // For example, if path is "foo.zip!/bar/bas/x.so", then we search for
  // "bar/bas/x.so" within "foo.zip".
  const char* const separator = strstr(path, kZipFileSeparator);
  if (separator == nullptr) {
    return -1;
  }
```

workingjubilee: *hmm*

workingjubilee: tempted to file the most contrived bug report

workingjubilee: I suppose we should match their bug with our own for the moment, but I'm still really curious if a `.so` can involve the `!/` sequence *inside* the APK zip?

sudoBash418: Seems to work fine, at least using the linker's CLI:

```
emu64xa:/data/local/tmp $ unzip -l android-test.apk                                                                                                                                                                                            
Archive:  android-test.apk
  Length      Date    Time    Name
---------  ---------- -----   ----
    52305  2024-09-08 02:07   suffix/3
    54461  2024-09-08 02:07   suffix/4
    52728  2024-09-08 02:07   prefix/1
    51323  2024-09-08 02:07   prefix/5
    21392  2024-09-08 02:07   prefix/4
    64481  2024-09-08 02:07   prefix/0
    11233  2024-09-08 02:07   suffix/8
      754  2024-09-08 02:07   suffix/7
    12335  2024-09-08 02:07   prefix/7
    50870  2024-09-08 02:07   prefix/6
      835  2024-09-08 02:07   suffix/2
    23008  2024-09-08 02:07   suffix/0
    21131  2024-09-08 02:07   suffix/9
 10560600  2024-09-08 02:05   somedir!/anotherone!/mylib.so
    11260  2024-09-08 02:07   prefix/3
    17046  2024-09-08 02:07   suffix/1
    43742  2024-09-08 02:07   prefix/8
    20599  2024-09-08 02:07   suffix/5
    54497  2024-09-08 02:07   prefix/2
     1370  2024-09-08 02:07   suffix/6
    26675  2024-09-08 02:07   prefix/9
---------                     -------
 11152645                     21 files
emu64xa:/data/local/tmp $ linker64 /data/local/tmp/android-test.apk!/somedir!/anotherone!/mylib.so
frame #0  - 0x00000073e9ee8d83bb - backtrace::backtrace::libunwind::trace::h89414344d67efe6b
                                 @ /build/src/backtrace/libunwind.rs:116
                                 - backtrace::backtrace::trace_unsynchronized::h78287cff64b72793
                                 @ /build/src/backtrace/mod.rs:66
[... clipped backtrace output ...]
frame #12 - 0x00000073e9ee8db1fe - main
frame #13 - 0x00000073e9ed3099f0 - __libc_init
```

workingjubilee: welp, there goes my hypothesis.

src/symbolize/gimli/elf.rs
sudoBash418: TIL it's part of POSIX (and that POSIX even publishes man pages for this kind of thing), thanks!

> It's just all the accesses after that which are racy as fuck.

Not sure I follow - you mean with other threads in the process?

workingjubilee: mmap is racy with every other process on the system.

src/symbolize/gimli/libs_dl_iterate_phdr.rs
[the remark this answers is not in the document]
workingjubilee: ...There's something that feels slightly off here, like we're relying implicitly on being mapped into certain addresses. Can you spell out the equation we're relying on?

src/symbolize/gimli/libs_dl_iterate_phdr.rs
```diff
@@ -73,6 +84,19 @@ unsafe extern "C" fn callback(
...
+    #[cfg(target_os = "android")]
+    let zip_offset: Option<u64> = {
+        // only check for ZIP-embedded file if we have data from /proc/self/maps
+        maps.as_ref().and_then(|maps| {
+            // check if file is embedded within a ZIP archive by searching for `!/`
+            super::extract_zip_path_android(&name).and_then(|_| {
+                // find MapsEntry matching library's base address and get its file offset
+                maps.iter()
+                    .find(|m| m.ip_matches(dlpi_addr as usize))
+                    .map(|m| m.offset())
```
workingjubilee, line 96: Note to self: reread the MapsEntry struct and its FromStr if you somehow develop another question about this.

workingjubilee:
Hm, this looks good but...

...ah, right.

Can you construct a test for specifically symbolicating a .so in a zip for Android? It likely has to be its own crate, like the other tests. Honestly, though, a bash script recorded from what you've been doing a zillion times I'm sure, and just plonked into the workflow yaml with a conditional to run it only on android, would also be fine.

With that, I can accept this!

workingjubilee:
...wait.
https://github.com/rust-lang/backtrace-rs/blob/df56954ca5cbfec623bd746009655697c36baad2/ci/docker/aarch64-linux-android/Dockerfile#L16

...nevermind, we're only testing linkage on Android apparently.

workingjubilee:
I believe I can accept this with a rebase.

workingjubilee (approved):

PR #667, opened 2024-09-07
Make mmap usage safe/sound

CraftSpider:
For those not aware - mmap is very difficult to use soundly, and relies on other programs in the file system not doing things like truncating the file beneath you. It's also not actually always faster. As such, it's often better to simply use an in-memory buffer.

Not sure if there's desire for this change, but it was easy enough, so figured I'd put it up for review.


bjorn3:
This doesn't help soundness at all. All files backtrace-rs mmap's are mmaped by the dynamic linker or the kernel anyway for the executable code of the process and thus modifying them would cause crashes or UB anyway.

CraftSpider:
Ah, I didn't consider that. That's good to know - instead of removing it, if you'd like, I can add actual soundness preconditions and a safety comment using that justification for it being okay.

workingjubilee:
Yeah, let's document this properly then.

workingjubilee:
some relevant sources:
- https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/functions/mmap.html
- https://man7.org/linux/man-pages/man2/mmap.2.html
- https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createfilemappinga
- https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-mapviewoffile

workingjubilee:
...actually, what bjorn3 said raises a larger issue, too...

PR #669, opened 2024-09-14
Update other GitHub actions

workingjubilee:
We started using download-artifact@v4.1.7 in https://github.com/rust-lang/backtrace-rs/pull/663 but the rest weren't so security critical so dependabot didn't bump them. Not having them "just" breaks a workflow or two. So bump them.

### DIFF
--- a/.github/workflows/check-binary-size.yml
+++ b/.github/workflows/check-binary-size.yml
@@ -126,7 +126,7 @@ jobs:
               { flag: "wx" },
             );
       - name: Upload size data
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4.1.7
         with:
           name: size-files
           path: ${{ env.SIZE_DATA_DIR }}/${{ env.SIZE_DATA_FILE }}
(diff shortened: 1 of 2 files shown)


workingjubilee:
It seems it doesn't want to use a version of this workflow that isn't the version on master.

.github/workflows/check-binary-size.yml
```diff
@@ -61,6 +61,7 @@ jobs:
         run: |
           # Bootstrap config
           cat <<EOF > config.toml
+          change-id = 9999999
```
ChrisDenton, line 64: This doesn't seem to have worked? Maybe it's not seeing the config.toml file for some reason?

workingjubilee: It's also using actions/checkout@v3. It won't use the workflow file changes until they're merged.

ChrisDenton: Oh I see! I get what you were saying now. This looks right to me so I guess we'll merge and see?

ChrisDenton (approved):

PR #670, opened 2024-09-14
Use fuzzy version for `{up,down}load-artifact`

workingjubilee:
I made a mistake in assuming there would be a twin for every version of the upload/download artifact changes. There are very not. Alternatively, we can use v4.1.8 for download-artifact and v4.4.0 for upload-artifact.

### DIFF
--- a/.github/workflows/check-binary-size.yml
+++ b/.github/workflows/check-binary-size.yml
@@ -127,7 +127,7 @@ jobs:
               { flag: "wx" },
             );
       - name: Upload size data
-        uses: actions/upload-artifact@v4.1.7
+        uses: actions/upload-artifact@v4
         with:
           name: size-files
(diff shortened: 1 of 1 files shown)


ChrisDenton (approved):
Looks good to me. There's no reason we need to pin to a specific minor version.

workingjubilee:
For future reference, these "fuzzy" versions are actually quite exact, no version resolution happens except via Git. Instead, GitHub simply uses a tag for each, e.g. these two tags are currently the same commit:
- https://github.com/actions/download-artifact/releases/tag/v4
- https://github.com/actions/download-artifact/releases/tag/v4.1.8

workingjubilee:
Ugh, fixing this has not been enough to restore the workflow to functionality.

PR #674, opened 2024-09-22
Document mmap safety

CraftSpider:
Supercedes #667

Document safety preconditions and reasoning for why they're followed. I'd like to push the unsafety down another level, or just remove the `mmap` function (make `Mmap` a common type with sys impls for the platform-specific bits), but those both are a lot more work.

### DIFF
--- a/src/symbolize/gimli.rs
+++ b/src/symbolize/gimli.rs
@@ -188,7 +188,7 @@ fn mmap(path: &Path) -> Option<Mmap> {
     let len = file.metadata().ok()?.len().try_into().ok()?;
     // SAFETY: All files we mmap are mmaped by the dynamic linker or the kernel already for the
     //         executable code of the process. Modifying them would cause crashes or UB anyways.
-    unsafe { Mmap::map(&file, len, 0) }
+    unsafe { Mmap::map(file, len, 0) }
 }
 
 cfg_if::cfg_if! {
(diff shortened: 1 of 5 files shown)


src/symbolize/gimli.rs
```diff
@@ -186,7 +186,9 @@ impl<'data> Context<'data> {
 fn mmap(path: &Path) -> Option<Mmap> {
     let file = File::open(path).ok()?;
     let len = file.metadata().ok()?.len().try_into().ok()?;
-    unsafe { Mmap::map(&file, len, 0) }
+    // SAFETY: All files we mmap are mmaped by the dynamic linker or the kernel already for the
+    //         executable code of the process. Modifying them would cause crashes or UB anyways.
```
philipc, line 190: This is true for some but not all uses of mmap in this crate. DWARF can be in a separate file that is not loaded by the dynamic linker or kernel (.dSYM on macOS, .debug/.dwz/.dwp on linux, or unpacked object files on both).

That doesn't mean we should use `fs::read` for those though. We don't need to read all of the DWARF, and doing so would affect performance and memory usage.

src/symbolize/gimli/elf.rs
```diff
@@ -69,9 +69,9 @@ impl Mapping {
             let len = file.metadata().ok()?.len();
 
             // NOTE: we map the remainder of the entire archive instead of just the library so we don't have to determine its length
-            // NOTE: mmap will fail if `zip_offset` is not page-aligned
+            // SAFETY: See `super::mmap` function
```
hkBst, line 72: The point of a safety comment for a use is to explain why that use is safe. I'm not sure what this is referring to and why it cannot explain safety directly instead.

hkBst (changes requested):

CraftSpider:
This PR is vastly out-of-date, and I don't have the time right now to update it. So I'm just going to close it for now.

PR #677, opened 2024-11-13
Recognize windows-sys signatures as "C" or "system" depending on cfg

### LINKED ISSUE #676: v0.3.74 fails to build on Windows
Here are the errors: [backtrace-errors.txt](https://github.com/user-attachments/files/17706575/backtrace-errors.txt)

We are building project with [buck2](https://github.com/facebook/buck2) but third party dependencies are resolved and vendored by cargo.

Is `windows-sys` dependency wrong?

v0.3.73 and below work fine. Other platforms seems to be ok as well.

dtolnay:
Closes #676.

This is not an ideal solution because it relies on `--cfg=windows_raw_dylib` to be passed to `backtrace` if and only if the same flag is passed to `windows-targets`. In Cargo this would always be the case and in Buck it's easy to arrange; not sure about different build systems.

I did not come up with a way to write a check that succeeds if `super::windows_sys::$name` is `unsafe extern "C" fn($($argty),*) -> $ret` OR `unsafe extern "system" fn($($argty),*) -> $ret` and fails for all other signatures.

### DIFF
--- a/src/dbghelp.rs
+++ b/src/dbghelp.rs
@@ -30,12 +30,6 @@ use core::mem;
 use core::ptr;
 use core::slice;
 
-// This is used when we're double-checking function signatures against windows-sys.
-#[inline(always)]
-fn assert_equal_types<T>(a: T, _b: T) -> T {
-    a
-}
-
 // This macro is used to define a `Dbghelp` structure which internally contains
 // all the function pointers that we might load.
 macro_rules! dbghelp {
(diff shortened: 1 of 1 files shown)


ChrisDenton:
This will need to exclude x86 because Windows x86 is the only target where the "system" is not the same as "C".

ChrisDenton (approved):
Thanks! 

CI is currently failing for unrelated reasons but the change is fairly straightforward and I've tested locally so I'm happy to merge this to unblock use of raw-dylib.

PR #679, opened 2024-11-13
Fix CI for lints and migrations

### LINKED ISSUE #678: Audit uses of `static mut`
The new warning is causing CI to fail. E.g:

```
error: creating a mutable reference to mutable static is discouraged
   --> src/symbolize/gimli.rs:368:11
    |
368 |         f(MAPPINGS_CACHE.get_or_insert_with(Cache::new))
    |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ mutable reference to mutable static
    |
    = note: for more information, see <https://doc.rust-lang.org/nightly/edition-guide/rust-2024/static-mut-references.html>
    = note: mutable references to mutable statics are dangerous; it's undefined behavior if any other pointer to the static is used or if any other reference is created for the static while the mutable reference lives
    = note: `-D static-mut-refs` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(static_mut_refs)]`

error: could not compile `backtrace` (lib) due to 1 previous error
```

Either we should allow the lint and add some comments or else actually fix the issue.

ChrisDenton:
This doesn't fix #678 but will hopefully get CI passing again.

Longer term, I would like to change our dbghelp library loading to be less eek.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -235,7 +235,7 @@ jobs:
       matrix:
         target:
           - wasm32-unknown-unknown
-          - wasm32-wasi
+          - wasm32-wasip1
           - x86_64-unknown-fuchsia
           - x86_64-fortanix-unknown-sgx
           - x86_64-unknown-illumos
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -92,7 +92,7 @@ jobs:
         if: contains(matrix.os, 'ubuntu')
         env:
           RUSTFLAGS: "-C link-arg=-Wl,--compress-debug-sections=zlib"
-      - run: cargo test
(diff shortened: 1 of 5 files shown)


ChrisDenton:
Oh, wasm target is still failing due to `wasm32-wasi` name change.

ChrisDenton:
I'm not sure why the binary size check is now failing. Unless maybe feature resolution has changed?

workingjubilee:
Oh, this is literally the same as 687 but missing the last commit, lol. oopsies.

workingjubilee:
The answer is we never made it a feature, actually.

workingjubilee (approved):

PR #681, opened 2024-11-20
Test Mac Catalyst in CI, and clean up unnecessary iOS env flags

madsmtm:
Mac Catalyst is a technology that allows macOS to emulate or closely mimic iOS in behaviour. Let's use that in CI, as a proxy for real testing on iOS.

Also, `rustc` will figure out the SDK itself nowadays, so calling `xcrun` should be unnecessary.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -162,20 +162,17 @@ jobs:
       matrix:
         include:
           - target: aarch64-apple-ios
-            sdk: iphoneos
           - target: x86_64-apple-ios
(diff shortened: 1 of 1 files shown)


workingjubilee:
> The workflow is not valid. .github/workflows/main.yml (Line: 173, Col: 14): Unexpected symbol: '('. Located at position 23 within expression: matrix.target.contains('macabi') && '' || '--no-run'

workingjubilee (approved):
Fixed the GHA syntax. Thanks!

PR #682, opened 2024-11-24
Enable libunwind for rtems

thesummer:
In order to switch the RTEMS port to `panic_unwind` these changes need to be added in `backtrace_rs`. 

Once this is merged and the submodule updated in rustc I will add another PR there to make the switch.


thesummer:
Hm, it seems that the CI needs the changes from PR https://github.com/rust-lang/rust/pull/130417 merged in order to use the `ruzstd` added in backtrace-rs previously. Any suggestions how to proceed? Do we wait until there is a solution to the issues discusses in `rustc`?

workingjubilee (approved):

PR #687, opened 2025-01-02
Fix CI again for lints and migrations

workingjubilee:


workingjubilee:
I do not care to fix our purely informational job.

tgross35:
> I do not care to fix our purely informational job.

GH has `continue-on-error: true`, you could set it for that step if you're looking to denoise the red CI

workingjubilee:
I fixed it anyways. Happy Holidays!

workingjubilee:
oh, lol, it won't actually take until after the merge because the base build is broken.

workingjubilee:
And I can't even test it because that workflow doesn't take PR diffs to it... anyway!

PR #688, opened 2025-01-03
Disable libunwind for NuttX

no1wudi:
NuttX is a resource-constrained system, libunwind is too large for it in terms of both code size and memory usage. So we disable libunwind for NuttX.

### DIFF
--- a/src/backtrace/mod.rs
+++ b/src/backtrace/mod.rs
@@ -174,7 +174,7 @@ cfg_if::cfg_if! {
         any(
             all(
                 unix,
-                not(target_os = "emscripten"),
+                not(any(target_os = "emscripten", target_os = "nuttx")),
                 not(all(target_os = "ios", target_arch = "arm")),
             ),
             all(


workingjubilee:
@no1wudi why are you building with unwinding enabled?

no1wudi:
@workingjubilee Previously, I have been using the RISC-V platform for development work related to Rust and NuttX. Recently, when I started testing on the ARM platform, I discovered that the ARM platform references _Unwind_GetIP, which is currently not supported in the NuttX environment. Considering that the size of the Rust code could be reduced by approximately 20% after disabling libunwind, we are currently considering turning it off for the time being.

workingjubilee:
@no1wudi it's... not supported?

...but this is Arm we're talking about...?

no1wudi:
> @no1wudi it's... not supported?
> 
> ...but this is Arm we're talking about...?

For ARM platform in NuttX, we are using unwind support from GCC (`libgcc.a` from arm-none-eabi-gcc),  it does not provide an implementation of _Unwind_GetIP.

This branch works, implement  _Unwind_GetIP on top of _Unwind_VRS_Get:

workingjubilee:
Isn't this what `panic_immediate_abort` is for?

no1wudi:
Get it, thank you !

I originally thought that setting `-Zbuild-std=std,panic_abort` would be sufficient, but I overlooked `-Zbuild-std-features=panic_immediate_abort`.

PR #689, opened 2025-01-04
Use a custom container for Cache's cache

workingjubilee:
The new Lru type, as a single-purpose "ArrayVec",
- removes a level of indirection
- omits a needless Vec allocation entirely
- internalizes invariants we attempted to maintain in open-coded form

It costs a little `unsafe` to handle Drop, which we unfortunately need as addr2line uses Arc inside Context.

### DIFF
--- a/src/symbolize/gimli/lru.rs
+++ b/src/symbolize/gimli/lru.rs
@@ -46,7 +46,7 @@ impl<T, const N: usize> Lru<T, N> {
         let len_to_init = self.len + 1;
         let mut last = MaybeUninit::new(value);
         for elem in self.arr[0..len_to_init].iter_mut() {
-            last = mem::replace(elem, last);
+            mem::swap(elem, &mut last);
         }
         self.len = len_to_init;
 
@@ -63,7 +63,7 @@ impl<T, const N: usize> Lru<T, N> {
         // so it is permissible to allow the len invariant to decay, as we always restore it
         let mut last = mem::replace(elem, MaybeUninit::uninit());
         for elem in self.arr[0..=idx].iter_mut() {
-            last = mem::replace(elem, last);
+            mem::swap(elem, &mut last);
         }
         self.arr
             .first_mut()
--- a/src/symbolize/gimli/lru.rs
+++ b/src/symbolize/gimli/lru.rs
@@ -46,7 +46,8 @@ impl<T, const N: usize> Lru<T, N> {
         let len_to_init = self.len + 1;
         let mut last = MaybeUninit::new(value);
         for elem in self.arr[0..len_to_init].iter_mut() {
-            mem::swap(elem, &mut last);
+            // OPT(size): using `mem::swap` allows surprising size regressions
+            last = mem::replace(elem, last);
         }
         self.len = len_to_init;
 
@@ -63,7 +64,8 @@ impl<T, const N: usize> Lru<T, N> {
         // so it is permissible to allow the len invariant to decay, as we always restore it
         let mut last = mem::replace(elem, MaybeUninit::uninit());
         for elem in self.arr[0..=idx].iter_mut() {
-            mem::swap(elem, &mut last);
+            // OPT(size): using `mem::swap` allows surprising size regressions
+            last = mem::replace(elem, last);
         }
         self.arr
             .first_mut()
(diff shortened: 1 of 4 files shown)


src/symbolize/gimli/lru.rs
```diff
@@ -0,0 +1,73 @@
...
+    // SAFETY: len <= initialized values
+    len: usize,
+    arr: [MaybeUninit<T>; N],
+}
+
+impl<T, const N: usize> Default for Lru<T, N> {
+    fn default() -> Self {
+        Lru {
+            len: 0,
+            arr: [const { MaybeUninit::uninit() }; N],
```
workingjubilee, line 15: this requires an MSRV bump, *or* to use the infamously sus-looking-but-actually-okay-for-this-ONE-instance:
```suggestion
            arr: unsafe { MaybeUninit::uninit().assume_init() },
```

workingjubilee (commented):
hell yeah!

src/symbolize/gimli/lru.rs
ChrisDenton: On the one hand I'm not against bumping the MSRV for this crate but on the other hand it is only one line. But on the third hand it is a very sussy line that would at least need a comment saying "no really, this is ok!"

What would the new MSRV be?

workingjubilee: if https://github.com/rust-lang/backtrace-rs/pull/689/commits/318dd91a3dc077bd7c3710ea66c5bc664f61590a passes then the answer is 1.79

ChrisDenton: Ah 1.73 to 1.79? That seems fine to me. We should probably also update the `rust-version` in `Cargo.toml` as that seems to have gone out of the sink.

src/symbolize/gimli/lru.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+}
+
+impl<T, const N: usize> Lru<T, N> {
+    #[inline]
+    pub fn clear(&mut self) {
+        let len = self.len;
+        self.len = 0;
+        // SAFETY: we can't touch these values again due to setting self.len = 0
+        unsafe { ptr::drop_in_place(ptr::addr_of_mut!(self.arr[0..len]) as *mut [T]) }
```
ChrisDenton, line 26: `cast` still requires sized, shame.

src/symbolize/gimli/lru.rs
```diff
@@ -0,0 +1,73 @@
...
+            return None;
+        } else if self.len == N {
+            self.len = N - 1;
+            // SAFETY: we maintain len invariant and bail on N == 0
+            unsafe { ptr::drop_in_place(self.arr.as_mut_ptr().cast::<T>().add(N - 1)) };
+        };
+        let len_to_init = self.len + 1;
+        let mut last = MaybeUninit::new(value);
+        for elem in self.arr[0..len_to_init].iter_mut() {
+            last = mem::replace(elem, last);
```
ChrisDenton, line 49: Isn't this just `mem::swap`?

workingjubilee: what we do is assign the `last` to an index, and then take out the old value. it isn't really `mem::swap` because we're instead moving everything a step over, and that isn't really expressible that way (and there's no `windows_mut`, besides).

ChrisDenton: Maybe I'm having a brain fart but I mean literally `last = mem::replace(elem, last);` looks suspiciously like `mem::swap(elem, &mut last);` no?

workingjubilee: oh.

workingjubilee: yes.

ChrisDenton: Hm, that had a big effect on code size???

workingjubilee:
...
@ChrisDenton apparently it optimizes worse?

ChrisDenton:
Yeah ok, no idea what that's about but revering back to `replace` makes sense. Maybe with a comment saying why.

ChrisDenton (approved):
This is such a nice improvement, thanks!

PR #692, opened 2025-01-19
libunwind: Use builtin _Unwind_GetIP for NuttX on ARM platform

no1wudi:
Like on other platforms (Linux/RTEMS etc), `_Unwind_GetIP` is not available on NuttX with ARM.


ChrisDenton (approved):
Thanks!

PR #693, opened 2025-01-21
Parse short backtraces out of debuginfo

jyn514:
This implements the runtime side of https://github.com/rust-lang/compiler-team/issues/818.

- Extract a helper out of `find_frames` for iterating over a `LookupContinuation`
- Make the type signature for `search_object_map` consistent across all platforms. Backtraces are inherently platform specific, but let's not make it any harder on ourselves than we have to.
- Add a new `pub fn short_backtraces() -> enum { ThisFrameOnly, Start, End }` API
- Use the new [`gimli::UnitRef::shared_attrs`](https://github.com/gimli-rs/gimli/pull/756) API to determine whether a given frame has a short backtrace. This, for now, requires a git dependency on gimli.

Note that this currently does not work on MacOS. I do not have a Mac to test this; someone lent me theirs and I tracked it down to `cx.find_dwarf_and_unit` returning `None`, but I have not had the time or resources to narrow it down further than that.

### DIFF
--- a/src/symbolize/gimli.rs
+++ b/src/symbolize/gimli.rs
@@ -8,8 +8,10 @@ use self::mmap::Mmap;
 use self::stash::Stash;
 use super::BytesOrWideString;
 use super::ResolveWhat;
+use super::ShortBacktrace;
 use super::SymbolName;
 use addr2line::gimli;
+use addr2line::{LookupContinuation, LookupResult};
 use core::convert::TryInto;
 use core::mem;
 use core::u32;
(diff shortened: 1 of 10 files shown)


src/symbolize/gimli.rs
```diff
@@ -409,6 +426,43 @@ impl Cache {
...
+            1 => ShortBacktrace::Start,
+            2 => ShortBacktrace::End,
+            _ => return None,
+        };
+        Some(this)
+    }
+}
+
+#[allow(non_upper_case_globals)]
+const DW_AT_short_backtrace: gimli::DwAt = gimli::DwAt(0x3c00);
```
jyn514, line 442: if this changes upstream it will also need to change here: https://github.com/llvm/llvm-project/pull/123683#discussion_r1923086452

PR #697, opened 2025-01-23
Cleanup Windows trace modules

ChrisDenton:
Windows 64-bit doesn't use dbghelp for tracing, so the name `dbghelp64.rs` was inaccurate and confusing. Rename it to `win64.rs`.

Also I moved a stray comment from `dbghelp32.rs` to `win64.rs` and removed `#![allow(bad_style)]` from both modules.


workingjubilee:
Thanks.

Is there a specific reason to keep dbghelp32 as dbghelp32 instead of win32?

ChrisDenton:
Not really, other than it didn't occur to me. I can totally do that though.

ChrisDenton:
Done!

ChrisDenton:
The error is caused by <https://github.com/rust-lang/rust/issues/136096> which in turn is caused by old mingw emitting a bunch of warnings in stderr. These aren't new but erroring on them is. Considering the lint is going to be [downgraded to allow](https://github.com/rust-lang/rust/pull/136098) until it is less noisy, maybe we should just wait a day or two for that to be merged.

workingjubilee:
Yeah I'm fine with just waiting a bit.

jyn514:
(the lint has been downgraded)

workingjubilee (approved):

PR #704, opened 2025-03-28
Add cygwin support

Berrysoft:
There's no `OsString::from_wide` on cygwin because it is treated as unix. ~~I choose `WideCharToMultiByte`.~~

EDIT: Now I use `cygwin_conv_path`.


Berrysoft:
@alexcrichton Could you review this PR?

Berrysoft:
@workingjubilee Could you review this PR, then...?

Berrysoft:
@ChrisDenton Could you review this PR?

Berrysoft:
Seems that everyone is going on holidays:(

mati865:
I'm not sure if this is the correct way to solve the problem. This will print Windows style paths, right?
Using APIs provided by Cygwin should be always preferred over calling Win32 APIs.

Is there no way to hook up implementation that other UNIXes use?

Berrysoft:
The path style depends on the compiler. If it's cross-compiled, the path is windows style. Otherwise it's unix style.

Berrysoft:
> Using APIs provided by Cygwin should be always preferred over calling Win32 APIs.

So where is cygwin api to get the dependency dlls' names? Here we use winapi to get the names only, not the paths.

mati865:
> The path style depends on the compiler. If it's cross-compiled, the path is windows style. Otherwise it's unix style.

If you call Win32 APIs, they'll return Windows style paths for files.
I think you are mixing host and target paths. Windows host compiler will use Windows style paths, but binaries built for Cygwin (either host or target) will use Cygwin style paths.

> So where is cygwin api to get the dependency dlls' names?

I don't know Cygwin APIs, maybe @jeremyd2019 could answer that?

> Here we use winapi to get the names only, not the paths.

Ah, that wouldn't be a big deal then, just a small divergence from the Cygwin way.

Can you share a backtrace from binary built for Cygwin that panics?

jeremyd2019:
I don't understand exactly what you're asking.  Cygwin has dladdr which will give you a (Cygwin) path to a DLL given an address within it.  I don't know of any Cygwin way to enumerate loaded modules though... It has that information, but I don't see an API to access it.

mati865:
> I don't know of any Cygwin way to enumerate loaded modules though... It has that information, but I don't see an API to access it.

IIUC that's what we need. We can stick to Win32 APIs then.

jeremyd2019:
> > Here we use winapi to get the names only, not the paths.

I think you're wrong there.  szExePath is converted, that's the full path.  szModuleName is just the name.  Also, that path is then passed to `mmap`, presumably the Cygwin `mmap`, so it should probably be getting the Cygwin path.  Ideas:
1. `cygwin_conv_path(CCP_WIN_W_TO_POSIX, ...`  Paths are not 100% guaranteed to be straight converted to valid UTF-8.  There are degenerate cases where something else happens.
2. Pass the modBaseAddr or hModule into `dlinfo` and let that handle the conversion.  That might be a little more wasteful as it would get the module file name again when that's already in the MODULEENTRY32W struct, but maybe easier to get at dlinfo than cygwin_conv_path since that's a more standard function.
3. look at `/proc/self/maps` ?  The unix side does that, but in combination with what seems to be an ELF-specific method of enumerating modules (which is probably why Cygwin doesn't provide it)

jeremyd2019:
https://inbox.sourceware.org/cygwin/6a5d6737-ebba-9d10-7e9f-3aae3ea3989a@jdrake.com/T/#u

Berrysoft:
I didn't find dlinfo method in cygwin. I will use cygwin_conv_path since it's straightforward. It's OK to be not UTF-8 because we are using OsString.

jeremyd2019:
> I didn't find dlinfo method in cygwin.

https://github.com/cygwin/cygwin/blob/d9e9405becbac3c8bc0c9f23f822fb2339990fab/winsup/cygwin/include/dlfcn.h#L49-L61

Berrysoft:
I'm not sure if `dladdr` accepts a HMODULE. Anyway it also uses `cygwin_conv_path`: https://github.com/cygwin/cygwin/blob/d9e9405becbac3c8bc0c9f23f822fb2339990fab/winsup/cygwin/dlfcn.cc#L430-L431

And I even think the code of cygwin above is a little buggy, considering that a UTF-16 path with length `MAX_PATH` might be converted to a POSIX path that is longer than `MAX_PATH` (e.g., the UTF-16 path is full of CJK characters)...

jeremyd2019:
> I'm not sure if `dladdr` accepts a HMODULE. Anyway it also uses `cygwin_conv_path`: https://github.com/cygwin/cygwin/blob/d9e9405becbac3c8bc0c9f23f822fb2339990fab/winsup/cygwin/dlfcn.cc#L430-L431

An `HMODULE` is the base address of the module.  `dladdr` takes an address within the module, the `HMODULE` of the module is the lowest such address.  But I only suggested it because I wasn't sure you had an easy way to call `cygwin_conv_path` directly from rust (it rather unsafely even for C takes different types depending on the enum value given, so for a "safe" language like rust it's probably a real headache).  Since you apparently can call `cygwin_conv_path`, that's best.
 
> And I even think the code of cygwin above is a little buggy, considering that a UTF-16 path with length `MAX_PATH` might be converted to a POSIX path that is longer than `MAX_PATH` (e.g., the UTF-16 path is full of CJK characters)...

Yeah, I see, the array in the DL_info struct is actually `PATH_MAX` not `MAX_PATH`, it should probably be passing that to `cygwin_conv_path` instead, and that should be plenty big to hold a `MAX_PATH` long UTF-16 string in any supported multibyte encoding.  I'll look at sending Cygwin a patch for that.

Berrysoft:
Thanks for review from @mati865 & @jeremyd2019 ! However it still needs a maintainer to review...

workingjubilee:
Apparently I'm reviewing backtrace-rs PRs today, a moment.

workingjubilee:
oh dear.

workingjubilee:
> An `HMODULE` is the base address of the module.  `dladdr` takes an address within the module, the `HMODULE` of the module is the lowest such address.  But I only suggested it because I wasn't sure you had an easy way to call `cygwin_conv_path` directly from rust (it rather unsafely even for C takes different types depending on the enum value given, so for a "safe" language like rust it's probably a real headache).  Since you apparently can call `cygwin_conv_path`, that's best.
 
Safety in Rust is strictly about memory safety, the stuff C also bans: out of bounds accesses and data races. So "safety" is not "a good API". You can write a bad API in Rust too that generates illogical garbage or is easy to misuse. The goal is just to make sure the program dumps that into the buffer you intended, instead of turning the entire address space into illogical garbage.

The most annoying thing about that function, actually, is that enums in Rust don't follow the "bitflags" rules that `cygwin_conv_path_t` implicitly relies on. So for any C enum like that, you can't directly use `enum` and have to write `const CONST: c_int = ..;` a bunch of times. Otherwise all calls to C are just `unsafe {}` because the Rust compiler doesn't have special knowledge of which C functions are safe to call, if any.

For a library like `backtrace`, which deals with grubbing through the address space and poking at random pointers to headers and pointers in headers... well, you get used to seeing `unsafe {}`. A lot.

src/symbolize/gimli/libs_windows.rs
```diff
@@ -43,13 +42,67 @@ unsafe fn add_loaded_images(ret: &mut Vec<Library>) {
...
+    const CCP_WIN_W_TO_POSIX: libc::c_uint = 3;
+
+    let name_len = unsafe {
+        cygwin_conv_path(
+            CCP_WIN_W_TO_POSIX,
+            long_path.as_ptr().cast(),
+            core::ptr::null_mut(),
+            0,
+        )
+    };
```
workingjubilee, line 67: Please add a comment documenting how `cygwin_conv_path` returns the length you must allocate if you call it with null pointers. And possibly a link to https://cygwin.com/cygwin-api/func-cygwin-conv-path.html as well?

src/symbolize/gimli/libs_windows.rs
```diff
@@ -43,13 +42,67 @@ unsafe fn add_loaded_images(ret: &mut Vec<Library>) {
...
+unsafe fn get_posix_path(long_path: &[u16]) -> Option<OsString> {
+    use super::mystd::os::unix::ffi::OsStringExt;
+
+    unsafe extern "C" {
+        fn cygwin_conv_path(
+            what: libc::c_uint,
+            from: *const libc::c_void,
+            to: *mut libc::c_void,
+            size: libc::size_t,
+        ) -> libc::ssize_t;
```
workingjubilee, line 56: As long as we have the ear of appropriate folks, then the safety conditions of this should be documented. When does this function trigger undefined behavior if you call it? Is it about certain arguments or does it just alternate every Tuesday? ...Probably the former.

src/symbolize/gimli/libs_windows.rs
```diff
@@ -43,13 +42,67 @@ unsafe fn add_loaded_images(ret: &mut Vec<Library>) {
...
+            CCP_WIN_W_TO_POSIX,
+            long_path.as_ptr().cast(),
+            core::ptr::null_mut(),
+            0,
+        )
+    };
+    // Expect at least 1 for null terminator.
+    if name_len < 1 {
+        return None;
+    }
```
workingjubilee, line 71: If it's -1 then it's an error condition. Should we check errno and react to that?

workingjubilee: If the answer is "no" then we should probably comment that we are explicitly saying "lol" to any error.

src/symbolize/gimli/libs_windows.rs
jeremyd2019: I don't know what exactly you are asking.  `from` and `to` need to be valid pointers, unless `size` is zero then `to` isn't used.  `from` needs to be either a nul-terminated `char *`, or a nul-terminated `wchar_t *` (if `what` is `CCP_WIN_W_TO_POSIX`).  `to` must have enough room to write `size` bytes.  If there's not enough space the function will error, it will not result in a truncated or non-nul-terminated result in `to`.  The function does try to trap invalid memory accesses and set errno to `EFAULT` but that's obviously not a good idea to rely on (there's always the chance that your bad pointer is accidentally somebody else's valid memory).

jeremyd2019: oh, `to` will be filled with a nul-terminated `char *` unless `what` is `CCP_POSIX_TO_WIN_W` in which case it will be a nul-terminated `wchar_t *`

workingjubilee: > I don't know what exactly you are asking.  `from` and `to` need to be valid pointers, unless `size` is zero then `to` isn't used.  `from` needs to be either a nul-terminated `char *`, or a nul-terminated `wchar_t *` (if `what` is `CCP_WIN_W_TO_POSIX`).  `to` must have enough room to write `size` bytes.  If there's not enough space the function will error, it will not result in a truncated or non-nul-terminated result in `to`. 

That is _exactly_ the answer I was looking for! The documentation was not explicit that `size == 0` allows `to == NULL`. It's a reasonable inference, but, well, until [a very recent decision by WG14](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3322.pdf), [`memcpy(NULL, NULL, 0)` was technically UB](https://developers.redhat.com/articles/2024/12/11/making-memcpynull-null-0-well-defined). So...!

Berrysoft: > The documentation was not explicit that `size == 0` allows `to == NULL`.

I think it's the documentation's problem:) It should say that if `size == 0`, `to` is not used. I can confirm that from the source code.

jeremyd2019: I was trying to do that too, but found a case where that might not be true https://inbox.sourceware.org/cygwin/9c8107b6-a040-d956-2590-ede1517bc2ad@jdrake.com/T/#u

workingjubilee: > oh, `to` will be filled with a nul-terminated `char *` unless `what` is `CCP_POSIX_TO_WIN_W` in which case it will be a nul-terminated `wchar_t *`

@jeremyd2019 Hm. That suggests that `to` needs to be a 2-byte aligned in that case, if `to` is written in the semi-obvious way of
```c
*((wchar_t *)to)[i] = wchar_val;
```

I suppose it's irrelevant for our usage since we're always going to call it in the other direction, but

workingjubilee: Aha!

I am glad that I asked silly questions instead of guessing. :^)

jeremyd2019: x86 doesn't enforce alignment, but sure.  You'd be giving it a wchar_t (or u16 I guess) array in that case anyway.

src/symbolize/gimli/libs_windows.rs
Berrysoft: I checked the documentation and found that it's not likely to return error here. I think just ignoring it is fine.

src/symbolize/gimli/libs_windows.rs
Berrysoft: Comments are added. The documentation didn't say that `to` is allowed to be null if `size` is 0. The source code of cygwin, however, allows it.

Do you think I need to change `null_mut()` to `1 as *mut c_void`?

src/symbolize/gimli/libs_windows.rs
Berrysoft: Here we only use `CCP_WIN_W_TO_POSIX` so `to` should be 1-byte aligned, I think.

Berrysoft: I passed `8 as *mut c_void`. It should be enough anyways:)

jeremyd2019: is that `(void *)8` ?  No.  Don't do that.  Just pass `NULL`.

Berrysoft: I thought @workingjubilee was suggesting passing a non-null pointer...

Berrysoft: Oh, I see. The problem is not related to our usage. I'll reset this commit.

src/symbolize/gimli/libs_windows.rs
```diff
@@ -43,13 +42,79 @@ unsafe fn add_loaded_images(ret: &mut Vec<Library>) {
...
+        // Doc: https://cygwin.com/cygwin-api/func-cygwin-conv-path.html
+        // Src: https://github.com/cygwin/cygwin/blob/718a15ba50e0d01c79800bd658c2477f9a603540/winsup/cygwin/path.cc#L3902
+        // Safety:
+        // * `what` should be `CCP_WIN_W_TO_POSIX` here
+        // * `from` is `*const u16`, null-terminated, UTF-16 path
+        // * `to` is `*mut u8` buffer, the buffer size is `size`.
+        fn cygwin_conv_path(
+            what: libc::c_uint,
+            from: *const libc::c_void,
+            to: *mut libc::c_void,
```
workingjubilee, line 60: So now, our only question is left as "why not just declare `from` as `*const u16` and `to` as `*mut u8`*?" Those types are ABI compatible and it would prevent errors from casts.

Aside from that, this looks fine.

src/symbolize/gimli/libs_windows.rs
```diff
@@ -43,13 +42,79 @@ unsafe fn add_loaded_images(ret: &mut Vec<Library>) {
...
+            name_buffer.len(),
+        )
+    };
+    // It's not likely to return error here.
+    if res != 0 {
+        return None;
+    }
+    // Ignore null terminator.
+    unsafe { name_buffer.set_len(name_len - 1) };
+    let name = OsString::from_vec(name_buffer);
```
workingjubilee, line 99: ...I wonder...

I don't believe an OsStr actually has to not include `'\0'` in its span? So we could just not bother updating the len here, if we want.

Berrysoft: OK.

Berrysoft: Well, well... it seems that I *should* set the length. If I don't, the code doesn't work.

Berrysoft: `mmap` opens the file, while the path passed to `File::open` could not contain null bytes.
```
Error { kind: InvalidInput, message: "file name contained an unexpected NUL byte" }
```

workingjubilee:
@Berrysoft oh lol whoops, thanks for noticing that!

workingjubilee:
Cool. This now looks good at a glance, and I will approve and merge when the part of me that feels confident in that statement (or would have the gall to disagree? one of those) actually wakes up.

src/symbolize/gimli/libs_windows.rs
jeremyd2019: > > The documentation was not explicit that `size == 0` allows `to == NULL`.
> 
> I think it's the documentation's problem:) It should say that if `size == 0`, `to` is not used. I can confirm that from the source code.

https://github.com/cygwin/cygwin/commit/06ee311c337aa71152b57936dbb0d6f83100e123

workingjubilee: Thank you for updating the documentation there! Lovely.

workingjubilee:
@jeremyd2019 Thank you for all your help in doublechecking the edge-cases (and even hammering a few out upstream)!

workingjubilee (approved):

Djihads80:
yay

PR #705, opened 2025-04-01
Add `optimize(size)` to some particularly large functions

clubby789:
When built as part of libstd, this saves a few kb in hello world binaries.

The comments are formatted like they are so that if this feature is stabilized they can be identified and used unconditionally.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -137,8 +137,11 @@ jobs:
 
       # Test that including as a submodule will still work, both with and without
       # the `backtrace` feature enabled.
+      # Building as if part of std requires nightly features to be available
       - run: cargo build --manifest-path crates/as-if-std/Cargo.toml
+        if: matrix.rust == 'nightly'
       - run: cargo build --manifest-path crates/as-if-std/Cargo.toml --no-default-features
+        if: matrix.rust == 'nightly'
 
   windows_arm64:
     name: Windows AArch64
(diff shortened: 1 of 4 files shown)


workingjubilee:
I mean, I think that's kind of the point, no?

workingjubilee:
This does seem more than slightly annoying though, to be clear.

clubby789:
Changed the `as-if-std` jobs to only run when using a nightly toolchain. The binary size job failure seems unrelated

workingjubilee:
It is indeed. I am of half a mind to give that CI job the knife, but it certainly won't block this. It _is_ supposed to help answering questions like "is this exact kind of PR an improvement?" but I can manually validate things.

Small questions first before I go looking: Did you evaluate these functions for size based on their source size or the actual compiled code size? Did the reduction happen only in those functions or was the entire compiled libstd smaller? I'm mostly wondering to make sure we didn't just hit some weird inlining thresholds or anti-thresholds.

workingjubilee:
> I'm mostly wondering to make sure we didn't just hit some weird inlining thresholds or anti-thresholds.

I think those are called "limits", Jubilee.

clubby789:
I identified these functions with `cargo bloat` iirc, then recompiled stage1 rustc with these changes, then a hello world binary with that compiler, and compared the size of those binaries

workingjubilee:
Neat!

workingjubilee (approved):
Reproduces!

clubby789:
https://github.com/rust-lang/rust/pull/140735#issuecomment-2858776389 (includes backtrace update with these changes)
Some small but nice binary size wins 🙂

PR #706, opened 2025-04-01
Re-enable `download-ci-llvm = true` for CI

clubby789:
https://github.com/rust-lang/rust/pull/138704 disabled this when running in GHA, but this is not relevant for backtrace.

The patched used here is copied from `compiler/rustc_codegen_cranelift/scripts/setup_rust_fork.sh`


Kobzol:
Obsoleted by https://github.com/rust-lang/backtrace-rs/pull/708. Thank you!

PR #707, opened 2025-04-28
Set up release-plz

tgross35:
Add a release-plz workflow and ensure that all unneeded packages are marked `publish = false`. We should get both the release PRs and the automatic publish on version changes.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,52 @@
+name: Release-plz
+
+permissions:
+  pull-requests: write
+  contents: write
+
+on:
+  push: { branches: [master] }
+
+jobs:
+  # Release unpublished packages.
+  release-plz-release:
+    name: Release-plz release
+    runs-on: ubuntu-24.04
+    if: ${{ github.repository_owner == 'rust-lang' }}
+    steps:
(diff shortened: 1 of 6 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Install Rust (rustup)
+        run: rustup update nightly --no-self-update && rustup default nightly
+      - name: Run release-plz
+        uses: MarcoIeni/release-plz-action@v0.5
```
workingjubilee, line 24: this seems to now be at https://github.com/release-plz/action

tgross35: Indeed - I updated that, and also split the jobs as seems to be recommended now from https://release-plz.ieni.dev/docs/github/quickstart#3-setup-the-workflow

workingjubilee (approved):

tgross35:
0.3.75 now exists, yay 🎉 https://crates.io/crates/backtrace/versions

PR #718, opened 2025-06-10
Upgrade `ruzstd`, `object`, and `addr2line` to the latest versions

tgross35:
`object` released a breaking change with adjustments to `elf::R_RISCV_GNU_*` and `pe::IMAGE_WEAK_EXTERN_*` constants, as well as a changed API for Mach-O dyld. None of these are relevant to our use.

`ruzstd` moved some API around to account for compression support, the only required update for us is to change import paths.

`addr2line` only bumped the gimli version, which includes an update that will help us drop the crates.io compiler-builtins dependency from the standard library.

Object changelog: https://github.com/gimli-rs/object/blob/master/CHANGELOG.md#0370
Ruzstd changelog: https://github.com/KillingSpark/zstd-rs/releases/tag/v0.8.0
Addr2line changelog: https://github.com/gimli-rs/addr2line/blob/master/CHANGELOG.md#0250-20250611

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.24.1"
+version = "0.25.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f5fb1d8e4442bd405fdfd1dacb42792696b0cf9cb15882e5d097b742a676d375"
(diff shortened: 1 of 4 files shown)


khuey:
Could we get a new backtrace release at some point with this (so I can have only one copy of object/gimli/etc in my binaries)?

PR #723, opened 2025-07-19
Remove lazily initialized Mutex

Mark-Simulacrum:
Mutex::new is const since 1.63, which is below the 1.82 MSRV, so this should be fine to rely on. I wouldn't expect this to practically matter ~ever, but spotted this while poking around and it seems like a nice bit of cleanup (also removes a bit of unsafe and a leaked allocation).


Mark-Simulacrum:
Alright, that fixed CI. Should be ready for review.

PR #725, opened 2025-08-28
Fix missing alignment check for ELF note header in fuchsia support

philipc:
The transmute required that the alignment of bytes was correct, but nothing checked this. Use `object::pod::from_bytes` which does the transmute and includes this check.

### DIFF
--- a/src/print/fuchsia.rs
+++ b/src/print/fuchsia.rs
@@ -1,7 +1,7 @@
 use core::fmt::{self, Write};
-use core::mem::{size_of, transmute};
 use core::slice::from_raw_parts;
 use libc::c_char;
+use object::NativeEndian as NE;
 
 unsafe extern "C" {
     // dl_iterate_phdr takes a callback that will receive a dl_phdr_info pointer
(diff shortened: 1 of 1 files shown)


src/print/fuchsia.rs
```diff
@@ -176,19 +171,14 @@ fn take_bytes_align4<'a>(num: usize, bytes: &mut &'a [u8]) -> Option<&'a [u8]> {
     Some(out)
 }
 
-// This function has no real invariants the caller must uphold other than
-// perhaps that 'bytes' should be aligned for performance (and on some
-// architectures correctness). The values in the Elf_Nhdr fields might
-// be nonsense but this function ensures no such thing.
+// This function has no invariants the caller must uphold, but it will return `None`
+// if 'bytes' is not correctly aligned.  The values in the Elf_Nhdr fields might be
```
workingjubilee, line 175: As long as we're here we might as well be exhaustive and make these doc comments.
```suggestion
/// This function has no invariants the caller must uphold, but
/// it will return `None`, without mutating, if `bytes` has insufficient size or alignment.
/// If this returns `Some(nhdr)`, then `bytes` was and remains 4-byte-aligned.
/// The values in the Elf_Nhdr fields might be nonsense.
```

PR #694, opened 2025-01-21
build(core): allow `windows-targets` 0.53.0

ErichDonGubler:

PR #727, opened 2025-09-21
Switch from `windows-targets` to `windows-link`

a1phyr:
Supersedes #694


a1phyr:
I'm pretty sure sure that the MacOS failure isn't related to this... Maybe an environment change?

PR #729, opened 2025-09-23
Optimize proc maps parsing code size

Noratrieb:
The current code size is really wastefully large. Originally, it was 1500 lines of assembly in Godbolt, now I reduced it to just under 800. The effect of `.text` size in hello world is from 297028 to 295453 (measured with `-Clto=fat -Copt-level=s` with a normal sysroot) which is small but not completely irrelevant. It's just a small fish in the bigger pond of DWARF parsing, but it's better than nothing.

I extracted the parsing of each component into a separate function to allow for better sharing. I replaced the string methods with manual iteration since that results in simpler code because it has to handle fewer cases. I also had to use unsafe because the bounds checks were sadly not optimized out and were really large.

I also made the parser less resilient against whitespace, now it no longer handles Unicode whitespace (an obvious simplification) but also no longer handles any whitespace except the normal SP. I think this is fine, it seems highly unlikely that a system would suddenly use another type of whitespace (but I guess not impossible?).

Another simplification was simply removing the parsing of unused fields that were not needed.

I can split it into separate commits if that helps review and maybe some of these changes are more of a hassle than it's worth (while some others like the field removals are obviously good), but I'll let you choose.


src/symbolize/gimli/parse_running_mmaps_unix.rs
```diff
@@ -44,13 +44,14 @@ pub(super) struct MapsEntry {
...
 pub(super) fn parse_maps() -> Result<Vec<MapsEntry>, &'static str> {
-    let mut v = Vec::new();
     let mut proc_self_maps =
         File::open("/proc/self/maps").map_err(|_| "Couldn't open /proc/self/maps")?;
     let mut buf = String::new();
     let _bytes_read = proc_self_maps
         .read_to_string(&mut buf)
         .map_err(|_| "Couldn't read /proc/self/maps")?;
+
+    let mut v = Vec::new();
```
workingjubilee, line 54: kinda insane that this matters that much

src/symbolize/gimli/parse_running_mmaps_unix.rs
```diff
@@ -84,41 +85,50 @@ impl FromStr for MapsEntry {
     // Note that paths may contain spaces, so we can't use `str::split` for parsing (until
     // Split::remainder is stabilized #77998).
     fn from_str(s: &str) -> Result<Self, Self::Err> {
+       fn error(msg: &str) -> &str {
+            if cfg!(debug_assertions) {
+                msg
+            } else {
+                "invalid map entry"
+            }
+        }
```
workingjubilee: yeah this makes sense.

src/symbolize/gimli/parse_running_mmaps_unix.rs
```diff
@@ -81,46 +90,73 @@ impl FromStr for MapsEntry {
...
+                Some(match_idx) => {
+                    // SAFETY: match_index comes from .bytes().position() of an ASCII character,
+                    //         so it's both in bounds and a UTF-8 boundary
+                    let result = unsafe { state.get_unchecked(..match_idx) };
+                    // SAFETY: Since match_idx is the ' ', there must be at least the end after it.
+                    *state = unsafe { state.get_unchecked((match_idx + 1)..) };
+                    result
+                }
+            }
+        }
```
workingjubilee, line 123: @hkBst This is an example of the kind of microoptimization that is both very hard to do in the compiler and also can reap significant benefits since here we care mostly about code size, even in the dead code, as this code size is essentially multiplied by *all* Rust binaries.

src/symbolize/gimli/parse_running_mmaps_unix.rs
Noratrieb: it's only a tiny improvement, so _that much_ is a bit of an overstatement :). actually i don't have numbers on this one, but I did see the call to `drop_in_place` before but not after.

src/symbolize/gimli/parse_running_mmaps_unix.rs
hkBst: @workingjubilee Thanks for thinking of me, although I'm not quite sure why you did.

PR #731, opened 2025-09-23
edition 2024 and raise MSRV to 1.88

workingjubilee:
With 1.90 being released, the stable - 2 policy now means 1.88 is fair game and we can use let-chains in the code. As we have lots of parsing code, this is a significant improvement. However, first, we have to also update to edition 2024. I have left out let-chain improvements from this PR to make it tractable to review.

Due to the somewhat adventuresome nature of updating the edition, this also includes cleaning up the win32 implementation a fair deal. Mostly this is the obvious unsafe-wrapping. That itself has been mostly split out to rust-lang/backtrace-rs#735, modulo one unsafe block that turned out to be required after rust-lang/backtrace-rs#737.

### DIFF
--- a/src/symbolize/dbghelp.rs
+++ b/src/symbolize/dbghelp.rs
@@ -245,8 +245,8 @@ unsafe fn do_resolve(
     // the real value.
     // SAFETY: We assume NameLen has been initialized by SymFromAddrW, and we initialized MaxNameLen
     let name_len = unsafe { cmp::min((*info).NameLen as usize, (*info).MaxNameLen as usize - 1) };
-    // Name must be initialized by SymFromAddrW, but we only interact with it as a pointer anyways.
-    let name_ptr = (&raw const (*info).Name).cast::<u16>();
+    // SAFETY: Name must be initialized by SymFromAddrW, but we only interact with it as a pointer anyways.
+    let name_ptr = unsafe { (&raw const (*info).Name).cast::<u16>() };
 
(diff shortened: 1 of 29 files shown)


ChrisDenton:
Could you rebase on master? Github's diff seems to be confused.

ChrisDenton (approved):
Thanks!

src/symbolize/dbghelp.rs
```diff
@@ -245,8 +245,8 @@ unsafe fn do_resolve(
     // the real value.
     // SAFETY: We assume NameLen has been initialized by SymFromAddrW, and we initialized MaxNameLen
     let name_len = unsafe { cmp::min((*info).NameLen as usize, (*info).MaxNameLen as usize - 1) };
-    // Name must be initialized by SymFromAddrW, but we only interact with it as a pointer anyways.
-    let name_ptr = (&raw const (*info).Name).cast::<u16>();
+    // SAFETY: Name must be initialized by SymFromAddrW, but we only interact with it as a pointer anyways.
+    let name_ptr = unsafe { (&raw const (*info).Name).cast::<u16>() };
```
workingjubilee, line 249: I removed this unsafe block during my post-push editing of #737 because I thought the compiler said it was superfluous, but now it is required for reasons that aren't clear to me.

ChrisDenton: Yeah I wondered about that but I trusted the compiler to complain if it's not needed, ha.

PR #732, opened 2025-09-24
Upgrade `cpp_demangle` to the latest version.

khuey:
cpp_demangle 0.5.0 drops the `Display` impl on the `Symbol` object because it can subtly violate the `Display` contract. The recommended replacement is `Symbol::demangle`.

Changelog: https://github.com/gimli-rs/cpp_demangle/blob/0.5.0/CHANGELOG.md#050


workingjubilee (approved):
all green! great!

PR #736, opened 2025-09-25
cleanup: simplify scan for xcoff .text section

workingjubilee:
As s_name is an 8-byte array, and we check only for ASCII characters, checking for UTF-8 validity is pure overhead.

Small cleanup opportunity I noticed while poking through the code looking for old-fashioned if-let-nesting to dedent with let-chains.


ChrisDenton (approved):
Yeah, `from_utf8` is completely unnecessary.

PR #750, opened 2026-04-12
Swap `WaitForSingleObjectEx` with `WaitForSingleObject` and check result

### LINKED ISSUE #749: Not checking the result of `WaitForSingleObjectEx` on Windows
In `src/dbghelp.rs:294`. `WaitForSingleObjectEx` is only checked with `debug_assert_eq!`. It also doesn't check `WAIT_ABANDONED`. Are we just assuming it can never fail?

Zoxc:
This swaps `WaitForSingleObjectEx` with the simpler `WaitForSingleObject` and also checks the result for failure.

Fixes https://github.com/rust-lang/backtrace-rs/issues/749


ChrisDenton:
Testing with CI issues now fixed...

ChrisDenton (approved):

PR #752, opened 2026-04-12
Make optimize_attribute conditional on libstd

ChrisDenton:
This should hopefully fix CI.

### DIFF
--- a/crates/as-if-std/src/lib.rs
+++ b/crates/as-if-std/src/lib.rs
@@ -2,7 +2,17 @@
 // submodule into the standard library. We try to set this crate up similarly
 // to the standard library itself to minimize the likelihood of issues when
 // updating the `backtrace` crate.
-#![feature(optimize_attribute)]
+#![cfg_attr(
+    // Only used by the gimli backend, see src/symbolize/mod.rs
+    all(
+        backtrace_in_libstd,
+        feature = "backtrace",
+        any(unix, all(windows, target_env = "gnu")),
+        not(target_vendor = "uwp"),
+        not(target_os = "emscripten"),
+    ),
+    feature(optimize_attribute)
+)]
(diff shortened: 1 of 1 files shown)


ChrisDenton:
cc @workingjubilee, I'm trying to get CI working again. Does this look ok to you? The alternative is to just allow unused features but it seems like a useful lint. Then again, it's not like I'd expect we'd use many of them.

workingjubilee:
This seems fine, yes! Will doublecheck something and then merge.

workingjubilee:
...What were you gonna check, Jubilee?

crates/as-if-std/src/lib.rs
```diff
@@ -2,7 +2,17 @@
...
 // to the standard library itself to minimize the likelihood of issues when
 // updating the `backtrace` crate.
-#![feature(optimize_attribute)]
+#![cfg_attr(
+    // Only used by the gimli backend, see src/symbolize/mod.rs
+    all(
+        backtrace_in_libstd,
+        feature = "backtrace",
+        any(unix, all(windows, target_env = "gnu")),
+        not(target_vendor = "uwp"),
```
workingjubilee, line 11: ...is windows with gnu *and* UWP a coherent *concept*?

anyway the `not` obtains so we're fine regardless.

workingjubilee (approved):

crates/as-if-std/src/lib.rs
ChrisDenton: Maybe. I mean, hypothetically at least. We have an `x86_64-uwp-windows-gnu` target but idk how well all these targets are maintained. Or even if they make any sense. I know the msvc one is used by at least one person because they complain when I break it (though not immediately).

PR #755, opened 2026-04-20
Add support for `aarch64-unknown-linux-pauthtest`

jchlanda:
This PR adds support for aarch64-unknown-linux-pauthtest, a target that enables Pointer Authentication Code (PAC) support in Rust on AArch64 ELF based Linux systems. It uses the aarch64-unknown-linux-pauthtest LLVM target and a pointer-authentication-enabled sysroot with a custom [musl](https://github.com/access-softek/musl) as a reference libc implementation. Dynamic linking is required, with a dynamic linker acting as the ELF interpreter that can resolve pauth relocations and enforce pointer authentication constraints.

Please consult a rust-lang PR for the details on the target: https://github.com/rust-lang/rust/pull/155722

### DIFF
--- a/src/backtrace/libunwind.rs
+++ b/src/backtrace/libunwind.rs
@@ -79,6 +79,15 @@ impl Frame {
         // clause, and if this is fixed that test in theory can be run on macOS!
         if cfg!(target_vendor = "apple") {
             self.ip()
+        } else if cfg!(target_abi = "pauthtest") {
+            // NOTE: `_Unwind_FindEnclosingFunction` creates a fresh unwind
+            // cursor and, on the pointer-authentication-enabled AArch64
+            // reference ABI, authenticates/re-signs the supplied IP using that
+            // cursor's SP. The original frame's SP is not available through
+            // this API, so the current libunwind implementation cannot be used
+            // here: it would attempt to authenticate the IP using the wrong SP.
+            // Return the raw instruction pointer instead.
+            self.ip()
         } else {
             unsafe { uw::_Unwind_FindEnclosingFunction(self.ip()) }
         }
--- a/tests/skip_inner_frames.rs
+++ b/tests/skip_inner_frames.rs
@@ -10,6 +10,12 @@ const ENABLED: bool = cfg!(all(
     target_os = "linux",
     // On ARM finding the enclosing function is simply returning the ip itself.
     not(target_arch = "arm"),
+    // On `aarch64-unknown-linux-pauthtest` `_Unwind_FindEnclosingFunction`
+    // cannot be used safely, because instruction pointers are not in a form
+    // suitable for authentication/resigning, so `symbol_address()` returns the
+    // raw IP instead. In this case the returned address may be inside the
+    // function body rather than at its entry.
+    not(target_env = "pauthtest"),
 ));
 
 #[test]
--- a/tests/smoke.rs
+++ b/tests/smoke.rs
@@ -25,6 +25,13 @@ fn get_actual_fn_pointer(fp: *mut c_void) -> *mut c_void {
 }
 
 #[test]
+// This test relies on recovering precise symbol addresses from instruction
+// pointers. On `aarch64-unknown-linux-pauthtest`, we cannot safely use
+// `_Unwind_FindEnclosingFunction`, and instead fall back to using raw
+// instruction pointers. As a result, symbol resolution cannot reliably
+// recover a canonical function entry address, and `sym.addr()` will often be
+// `None`. Therefore this test is disabled.
+#[cfg_attr(target_env = "pauthtest", ignore)]
 // FIXME: shouldn't ignore this test on i686-msvc, unsure why it's failing
 #[cfg_attr(all(target_arch = "x86", target_env = "msvc"), ignore)]
 #[inline(never)]
@@ -310,7 +317,10 @@ fn sp_smoke_test() {
                 let r = refs.pop().unwrap();
                 eprintln!("ref = {:p}", r);
                 if sp as usize != 0 {
-                    assert!(r > sp);
+                    // Stack grows down, however stack slots can be reused and
+                    // frame pointers ommited, `r == sp` should be a valid edge
+                    // case.
+                    assert!(r >= sp);
                     if let Some(child_ref) = child_ref {
                         assert!(sp >= child_ref);
                     }


src/backtrace/libunwind.rs
```diff
@@ -79,6 +79,18 @@ impl Frame {
...
+            // must not use uw::_Unwind_FindEnclosingFunction. This is because,
+            // for pauthtest toolchain, libunwind will try to authenticate and
+            // resign it. Signing here (apart from risking creating a signing
+            // oracle) is not possible. According to the schema the value must
+            // be signed using SP as the discriminator - which is the problem.
+            // SP obtained here would not match the SP at the auth-resign time,
+            // as uw::_Unwind_FindEnclosingFunction creates a new context so
+            // the SP used for signing here would belong to a different frame
+            // that the one used for auth-resign. Hence return a raw value.
+            self.ip()
```
bjorn3, line 93: Why is `ip` not signed? It is directly the result of `_Unwind_GetIP`. It should be possible to pass that to `_Unwind_FindEnclosingFunction`. How else would you even be able to use `_Unwind_FindEnclosingFunction` otherwise?

src/backtrace/libunwind.rs
```diff
@@ -79,6 +79,18 @@ impl Frame {
         // clause, and if this is fixed that test in theory can be run on macOS!
         if cfg!(target_vendor = "apple") {
             self.ip()
+        } else if cfg!(target_env = "pauthtest") {
```
bjorn3, line 82: Should this not check for the respective target feature being enabled? Not every target with pointer authentication enabled will use that target env, right?

src/backtrace/libunwind.rs
jchlanda: > Why is ip not signed? It is directly the result of _Unwind_GetIP.

A raw `ip` is expected. `_Unwind_GetIP` authenticates the pointer and returns a stripped version (via `ptrauth_auth_data` [here](https://github.com/llvm/llvm-project/blob/main/libunwind/src/UnwindLevel1.c#L624)).

> It should be possible to pass that to `_Unwind_FindEnclosingFunction`. How
> else would you even be able to use `_Unwind_FindEnclosingFunction` otherwise?

This is the crux of the issue. `_Unwind_FindEnclosingFunction` is fundamentally incompatible with the pointer authentication model. The problem is that `_Unwind_FindEnclosingFunction` does not follow PAC semantics, it blindly creates a fresh unwind cursor and then resets the instruction pointer [here](https://github.com/llvm/llvm-project/blob/main/libunwind/src/UnwindLevel1-gcc-ext.c#L122): `__unw_set_reg(cursor, UNW_REG_IP, pc);`. As if to say: "treat this `pc` as if it belonged to this new cursor's frame".

`__unw_set_reg` does apply PAC logic internally ([authenticates](https://github.com/llvm/llvm-project/blob/main/libunwind/src/libunwind.cpp#L153)) under the assumption that the `ip` behaves like a return address for the current frame.

This cannot be fixed on the Rust side. `_Unwind_FindEnclosingFunction` constructs a new cursor (and therefore a new `sp`), so there is no way to correctly sign pointer: any signing we would perform would use the wrong context and would still fail authentication inside libunwind.

src/backtrace/libunwind.rs
jchlanda: You are right.
We're in the middle of preparing an upstream patch that introduces a pointer authentication aware target. It just so happens that in there I ended up using (incorrectly) the target environment as the gating mechanism. 

We have a follow up task to resolve the target to a set of features, just like it's done in clang [where it solves platform/environment specifics on the driver](https://github.com/llvm/llvm-project/blob/main/clang/lib/Driver/ToolChains/Linux.cpp#L552) level and later on everything is based on [language flags](https://github.com/llvm/llvm-project/blob/main/clang/lib/Driver/ToolChains/Linux.cpp#L501) that are there regardless of the platform. And finally are resolved to a [concrete signing schema](https://github.com/llvm/llvm-project/blob/main/clang/lib/Frontend/CompilerInvocation.cpp#L1445).

This will become a bit more clear once the PR lands and I submit a ticket.

src/backtrace/libunwind.rs
bjorn3: That sounds like a bug in libunwind to me. What purpose does `_Unwind_FindEnclosingFunction` have at all if you can't pass the result of `_Unwind_GetIP` to it?

jchlanda: I'm not sure, the documentation seems to be pretty scarce, and one I [found](https://refspecs.linuxbase.org/LSB_4.1.0/LSB-Core-generic/LSB-Core-generic/baselib--unwind-findenclosingfunction.html) is rather quite lax about possible outcomes of the call, explicitly allowing failure.

I personally don't read it as a bug. To me it is more that `_Unwind_FindEnclosingFunction` was designed before pointer authentication existed, and its API implicitly assumes that instruction pointers are context-free values. But in PAC-aware code PC is only meaningful in the context of the (original) SP.

bjorn3: Why does `_Unwind_FindEnclosingFunction` need an authenticated pointer? It doesn't dereference the pointer, only looks it up in a side table with the result not having any security relevance. The only useful thing you can do with it afaik is looking up debuginfo for the function. Due to inlining and outlining you can't guarantee that it points to any particular function.

jchlanda: Let me re-phrase the problem here.

> Why does `_Unwind_FindEnclosingFunction` need an authenticated pointer?

It does not, perhaps this was an oversimplification on my side. The real problem is about what invariant `__unw_set_reg` is enforcing under PAC - and since `_Unwind_FindEnclosingFunction` calls set reg, that's why the simplification crept into the discussion.

You are right that `_Unwind_GetIP` returns a usable pointer. It authenticates a signed pointer with the *original* `sp` and gives you a stripped version.
The problem shows up in `_Unwind_FindEnclosingFunction`, because it creates a fresh cursor (ergo a new `sp`) and then calls:
```
__unw_set_reg(&cursor, UNW_REG_IP, pc);
```

On PAC aware platforms this will go through an auth/resign sequence:
```
        unw_word_t opaque_value = (uint64_t)ptrauth_auth_and_resign(
            (void *)value, ptrauth_key_return_address, sp,
            ptrauth_key_return_address, &authenticated_value);
```
What that code does is it enforces that `value` (`ip` from Rust) is a valid for *this cursor's* `sp` - so `ip` *must* be a return address for the current frame.

However, as you noted: `ip` is stripped by `_Unwind_GetIP` and there is no way for us to sign it. So there is no way to make it pass auth/resign check. We can’t re-sign it correctly because we don’t have the original `sp`, and using the new `sp` will fail authentication.

So the issue isn’t just "signed vs unsigned". It’s that `_Unwind_FindEnclosingFunction` relies on being able to treat a pointer as if it were a return address for an arbitrary frame, while `__unw_set_reg` (correctly) enforces that return addresses are `sp`-bound under PAC. And that is what makes `_Unwind_FindEnclosingFunction` fundamentally incompatible.

As a side note a notion of lack of support for `symbol_address` is already present in the [test suite](https://github.com/rust-lang/backtrace-rs/blob/28ec93b503bf0410745bc3d571bf3dc1caac3019/tests/skip_inner_frames.rs#L4).

jchlanda:
Although the parent change ([introduction](https://github.com/rust-lang/rust/pull/155722) of aarch64-unknown-linux-pauthtest) has not landed yet, it seems this should not prevent this PR from being marked as ready for review.

jchlanda:
@davidtwco, @folkertdev, @tgross35, @madsmtm FWI this is a follow up to https://github.com/rust-lang/rust/pull/155722

workingjubilee:
Hello. That PR needs to land first for this to be accepted, but it can be reviewed now? Noted. Drawing up my queue again.

jchlanda:
> Hello. That PR needs to land first for this to be accepted, but it can be reviewed now? Noted. Drawing up my queue again.

Yeap, it's ready for review. If you feel like going on a review spree, there is plenty more pointer auth stuff:
* `pauthtest` introduction: https://github.com/rust-lang/rust/pull/155722
* library support follow up: https://github.com/rust-lang/rust/pull/156548
* config follow up: https://github.com/rust-lang/rust/pull/156712
And the crates
* libc: https://github.com/rust-lang/libc/pull/5065
* backtrace: https://github.com/rust-lang/backtrace-rs/pull/755

Thank you!

src/backtrace/libunwind.rs
jchlanda: Please let me know if my description of why `uw::_Unwind_FindEnclosingFunction` is incompatible with pointer authentication is insufficient. Alternatively, if you have any ideas on how it should be addressed, I would be very keen to hear them.

src/backtrace/libunwind.rs
jchlanda: It was decided that using `target_env = "musl"` and `target_abi = "pauthtest"`is the way forward, I've updated the check to be `target_abi = "pauthtest"`.

The rust-lang commit introducing the change: https://github.com/rust-lang/rust/pull/155722/changes/229e0f54c1c0e1c183900e2fe77390a5063acdfd

src/backtrace/libunwind.rs
kovdan01: @bjorn3 If you have any questions related to how pointer authentication is working in terms of libunwind, I would be happy to answer - I was previously working on that from Linux perspective, while most work was done by folks from Apple.

I confirm @jchlanda 's reasoning in the thread above and I think that it should be good enough to just return `self.ip()` for pauthtest target.

Also note that the signing/auehtntication behavior from libunwind is mostly shared between Apple's arm64e and Linux's pauthtest (when not platform-specific, e.g. personality pointers are stored and signed differently on Darwin and Linux). So, while I do not believe that we have any bug in libunwind's behavior and API design, even if we do - it would require a big discussion involving folks from Apple. And at this point, it's probably better just to conform to status-quo here in backtrace-rs.

So, we would very much appreciate your feedback on this

jchlanda:
Hi all (@davidtwco, @folkertdev, @tgross35, @madsmtm, @bjorn3, @workingjubilee),

The PR [introducing the pauthtest](https://github.com/rust-lang/rust/pull/155722) target has been merged into main. I'd be very grateful if we could move this one forward as well.

Thank you!

tgross35:
No need to ping me on pauthtest events - you already have reviewers on your PRs, I'm not following pauthtest anything special, and I don't really do reviews for backtrace or cc. I believe that applies to most of the people in that list.

jchlanda:
> No need to ping me on pauthtest events - you already have reviewers on your PRs, I'm not following pauthtest anything special, and I don't really do reviews for backtrace or cc. I believe that applies to most of the people in that list.

Feel free to ignore it then, apologies for the noise.

workingjubilee:
And my time is uncompensated, as it is for many reviewers. This is a problem that Access Softek can address if it wants.

jchlanda:
@bjorn3 the other PAC related rust-lang PRs have now been earmarked for merge, could we please move this one forward?

src/backtrace/libunwind.rs
bjorn3: I think using `self.ip()` here will break the internal frame omission for backtraces:

https://github.com/rust-lang/backtrace-rs/blob/06b2cc573703c98df4b543422c8339694420bb02/src/capture.rs#L300-L303

And in tokio: https://github.com/tokio-rs/tokio/blob/dac81bf8c8de0a3e35f1626643674ba9faf9569c/tokio/src/runtime/task/trace/trace_impl.rs#L16

I think it would also effectively break the scoped-trace crate: https://github.com/jswrenn/scoped-trace/blob/e299c46905a2063957414c34d76264281852461b/src/lib.rs#L126

Is there really no way to retain this support? Even if just stripping the pointer signature? `_Unwind_FindEnclosingFunction` can be fixed to accept arbitrary pointer signatures without compromising security, right?

jchlanda: I will need to dig into it again to provide a more detailed answer, but for now, a couple of quick observations:

* returning `self.ip()` is exactly what happens on Apple, so the precedence is already there;
* I'm not familiar with neither Tokio nor scoped-trace, so I can't comment on their relative importance in the wider ecosystem. What I can say with confidence, though, is that the current implementation (`uw::_Unwind_FindEnclosingFunction(self.ip())`) causes every program compiled for `aarch64-unknown-linux-pauthtest` that relies on backtrace-rs to encounter a pointer authentication failure. As a first approximation, that's every program that panics.

jchlanda: @bjorn3 I spent some time looking into this issue again.

To sum up, there are two stages:

1. backtrace-rs obtains the instruction pointer.
   This is done with `_Unwind_GetIP(context)`. On PAC-enabled AArch64, libunwind authenticates the stored PC and returns an instruction pointer value associated with the original frame context (including the frame's SP-based PAC discriminator). That value is what we keep in `Frame::ip()`.

```cpp
uint64_t Registers_arm64::getIP() const {
  uint64_t value = _registers.__pc;

  value = (uint64_t)ptrauth_auth_and_resign(
      (void *)_registers.__pc,
      ptrauth_key_return_address,
      &_registers.__pc,
      ptrauth_key_return_address,
      getSP());

  return value;
}
```

2. backtrace-rs calls `_Unwind_FindEnclosingFunction`.
   The expectation is that libunwind returns the start address of the enclosing function. This call is [implemented as](https://github.com/llvm/llvm-project/blob/main/libunwind/src/UnwindLevel1-gcc-ext.c#L117):

```cpp
    unw_cursor_t cursor;
    unw_context_t uc;
    unw_proc_info_t info;
    __unw_getcontext(&uc);
    __unw_init_local(&cursor, &uc);
    __unw_set_reg(&cursor, UNW_REG_IP, (unw_word_t)(intptr_t)pc);
    if (__unw_get_proc_info(&cursor, &info) == UNW_ESUCCESS)
        return (void *)(intptr_t) info.start_ip;
```

The root cause of the issue is in this snippet: the cursor is **not** the original frame that produced `pc`. It is a brand new unwind cursor initialized from the current thread context.



On PAC-enabled AArch64, the instruction pointer returned by `_Unwind_GetIP()` is not simply an arbitrary integer address. Although it is suitable for use as a code address by callers, it is still a value whose PAC validity depends on the original frame context (in particular, the SP discriminator used by the PAC scheme). When `_Unwind_FindEnclosingFunction()` imports this IP into its newly created cursor, it goes through `__unw_set_reg()`, which calls the AArch64 implementation of [`Registers_arm64::setIP(value)`](https://github.com/llvm/llvm-project/blob/main/libunwind/src/Registers.hpp#L1908):

```cpp
value = (uint64_t)ptrauth_auth_and_resign(
    (void *)value,
    ptrauth_key_return_address,
    getSP(),
    ptrauth_key_return_address,
    &_registers.__pc);

_registers.__pc = value;
```

`getSP()` here returns the SP stored in the newly created cursor. It is not the SP belonging to the frame from which `_Unwind_GetIP()` obtained the instruction pointer. Therefore, `setIP()` attempts to authenticate an IP that belongs to the original frame using the SP of a different frame. The authentication step fails because the PAC discriminator does not match; the issue is not that the IP lacks a signature, but that it is being reintroduced into a different unwind context.

> Is there really no way to retain this support?

Not from backtrace-rs, as far as I can see. By the time we call `_Unwind_FindEnclosingFunction`, all we have is the instruction pointer returned by `_Unwind_GetIP()`. `_Unwind_FindEnclosingFunction` immediately constructs its own unwind cursor and, as we know now, this must fail authentication - since the original SP is no longer available. And because that cursor is created entirely inside libunwind, backtrace-rs has no opportunity to provide the original SP that was associated with the instruction pointer returned by `_Unwind_GetIP()`. From our side, there doesn't appear to be any way to make this call succeed.

> Even if just stripping the pointer signature?

`_Unwind_GetIP()` already returns the public instruction pointer representation expected by callers. The problem is not that backtrace-rs passes an unstripped/signed pointer incorrectly; the problem is that `_Unwind_FindEnclosingFunction()` tries to import that pointer into a new unwind context with a different SP, eventually calls `Registers_arm64::setIP()`, which authenticates/re-signs the supplied value using the SP of the newly-created cursor.

> `_Unwind_FindEnclosingFunction` can be fixed to accept arbitrary pointer
> signatures without compromising security, right?

I think it should be possible. I earlier said that `_Unwind_FindEnclosingFunction` is fundamentally incompatible with PAC enabled arch, that is probably not fully correct, I should have said: the current implementation in LLVM's libunwind doesn't work for this use case.

For example, if libunwind implemented `_Unwind_FindEnclosingFunction` without synthesizing a fresh unwind cursor, or otherwise avoided routing through `Registers_arm64::setIP()`, then the problem would not manifest. I am not in position to say if either of those is possible to implement.

And finally, from the perspective of backtrace-rs, the current implementation aborts on `aarch64-unknown-linux-pauthtest`, and we don't have enough information (most notably the original frame's SP) to make this call succeed with the current API.

Therefore, the only implementation that works reliably for pauthtest today is to use the instruction pointer already returned by `_Unwind_GetIP()` (`self.ip()`), which is also what backtrace-rs already does on Apple targets.

jchlanda:
@ChrisDenton, would you be able to help with a review? We've had a somewhat lengthy discussion about the merits of this change, and I believe the last message in the thread (https://github.com/rust-lang/backtrace-rs/pull/755#discussion_r3604258396) provides a fairly comprehensive justification for the patch.

tests/smoke.rs
```diff
@@ -310,7 +317,10 @@ fn sp_smoke_test() {
                 let r = refs.pop().unwrap();
                 eprintln!("ref = {:p}", r);
                 if sp as usize != 0 {
-                    assert!(r > sp);
+                    // Stack grows down, however stack slots can be reused and
+                    // frame pointers ommited, `r == sp` should be a valid edge
+                    // case.
+                    assert!(r >= sp);
```
ChrisDenton, line 323: This change doesn't seem specifically related to `pauthtest`?

ChrisDenton (commented):
I would need to read up on all the context first. Which might take me a while atm.

tests/smoke.rs
jchlanda: That's right. It just came up during testing on our side, and I bundled it into this PR since both changes were small. If you prefer, I can split it into a separate PR.

jchlanda:
> I would need to read up on all the context first. Which might take me a while atm.

If you need any help with the `pauthtest` target or have any questions, I'd be happy to help.

For reference, the document introducing `aarch64-unknown-linux-pauthtest` can be found here: [aarch64-unknown-linux-pauthtest.md](https://github.com/rust-lang/rust/blob/main/src/doc/rustc/src/platform-support/aarch64-unknown-linux-pauthtest.md)

I also keep the [tracking issue](https://github.com/rust-lang/rust-project-goals/issues/618) up to date with work related to the target.

jchlanda:
Friendly ping @ChrisDenton. Thank you.

src/backtrace/libunwind.rs
ChrisDenton: cc @bjorn3 would you agree that this is the best option available to us?

ChrisDenton:
Sorry, just getting back to this. My fear here is that backtrace-rs, as currently designed, isn't going to be a great fit for this. But just to be clear on what we can currently support, with the standard library's usage, we can get accurate backtrace right? And symbolization of that backtrace works?

jchlanda:
> Sorry, just getting back to this. My fear here is that backtrace-rs, as currently designed, isn't going to be a great fit for this. But just to be clear on what we can currently support, with the standard library's usage, we can get accurate backtrace right? And symbolization of that backtrace works?

Yes, both work on this target.

For example:

```rust
use std::panic;

fn main() {
    let _ = panic::catch_unwind(|| {
        panic!("BOOM");
    });
}
```

Compiled for `aarch64-unknown-linux-pauthtest` and run with `RUST_BACKTRACE=full`, the standard library produces a backtrace with correctly symbolized function names and source locations (shortened):

```
16: ... - panic[... ]::main::{closure#0}
    at .../local_tests/panic/panic.rs:5:9
17: ... - std[... ]::panicking::catch_unwind::do_call<...>
    at .../library/std/src/panicking.rs:573:43
20: ... - panic[... ]::main
    at .../local_tests/panic/panic.rs:4:13
33: ... - main
```
<details>
<summary>Full log</summary>

```
$ RUST_BACKTRACE=full ./panic_O3_g

thread 'main' (1851724) panicked at panic.rs:5:9:
BOOM
stack backtrace:
   0:     0xaeeb874fae0c - trace
                               at /home/Work/pauth/rust/library/std/src/../../backtrace/src/backtrace/libunwind.rs:208:9
   1:     0xaeeb874fae0c - trace_unsynchronized<std::sys::backtrace::_print_fmt::{closure_env#1}>
                               at /home/Work/pauth/rust/library/std/src/../../backtrace/src/backtrace/mod.rs:66:14
   2:     0xaeeb874fae0c - _print_fmt
                               at /home/Work/pauth/rust/library/std/src/sys/backtrace.rs:74:9
   3:     0xaeeb874fae0c - fmt
                               at /home/Work/pauth/rust/library/std/src/sys/backtrace.rs:44:26
   4:     0xaeeb8750f0e0 - fmt
                               at /home/Work/pauth/rust/library/core/src/fmt/rt.rs:152:76
   5:     0xaeeb8750f0e0 - write
   6:     0xaeeb874fea10 - default_write_fmt<std::sys::stdio::unix::Stderr>
                               at /home/Work/pauth/rust/library/core/src/io/write.rs:402:11
   7:     0xaeeb874fea10 - write_fmt<std::sys::stdio::unix::Stderr>
                               at /home/Work/pauth/rust/library/core/src/io/write.rs:335:13
   8:     0xaeeb874e4c3c - print
                               at /home/Work/pauth/rust/library/std/src/sys/backtrace.rs:47:9
   9:     0xaeeb874e4c3c - {closure#0}
                               at /home/Work/pauth/rust/library/std/src/panicking.rs:289:27
  10:     0xaeeb874f6aac - default_hook
                               at /home/Work/pauth/rust/library/std/src/panicking.rs:316:9
  11:     0xaeeb874f6ce0 - panic_with_hook
                               at /home/Work/pauth/rust/library/std/src/panicking.rs:811:13
  12:     0xaeeb874e4d60 - {closure#0}
                               at /home/Work/pauth/rust/library/std/src/panicking.rs:678:13
  13:     0xaeeb874db160 - std[9a20e64c0bb2485b]::sys::backtrace::__rust_end_short_backtrace::<std[9a20e64c0bb2485b]::panicking::panic_handler::{closure#0}, !>
                               at /home/Work/pauth/rust/library/std/src/sys/backtrace.rs:182:18
  14:     0xaeeb874e58b4 - panic_handler
                               at /home/Work/pauth/rust/library/std/src/panicking.rs:676:5
  15:     0xaeeb8750fb14 - panic_fmt
                               at /home/Work/pauth/rust/library/core/src/panicking.rs:80:14
  16:     0xaeeb874cf280 - panic[50ad4060d4f2f1db]::main::{closure#0}
                               at /home/Work/pauth/rust/local_tests/panic/panic.rs:5:9
  17:     0xaeeb874cf1b8 - std[9a20e64c0bb2485b]::panicking::catch_unwind::do_call::<panic[50ad4060d4f2f1db]::main::{closure#0}, !>
                               at /home/Work/pauth/rust/library/std/src/panicking.rs:573:43
  18:     0xaeeb874cf1b8 - std[9a20e64c0bb2485b]::panicking::catch_unwind::<!, panic[50ad4060d4f2f1db]::main::{closure#0}>
                               at /home/Work/pauth/rust/library/std/src/panicking.rs:541:19
  19:     0xaeeb874cf2b8 - std[9a20e64c0bb2485b]::panic::catch_unwind::<panic[50ad4060d4f2f1db]::main::{closure#0}, !>
                               at /home/Work/pauth/rust/library/std/src/panic.rs:359:14
  20:     0xaeeb874cf2b8 - panic[50ad4060d4f2f1db]::main
                               at /home/Work/pauth/rust/local_tests/panic/panic.rs:4:13
  21:     0xaeeb874cf1e8 - <fn() as core[c3c78aaece9448f7]::ops::function::FnOnce<()>>::call_once
                               at /home/Work/pauth/rust/library/core/src/ops/function.rs:250:5
  22:     0xaeeb874cf1e8 - std[9a20e64c0bb2485b]::sys::backtrace::__rust_begin_short_backtrace::<fn(), ()>
                               at /home/Work/pauth/rust/library/std/src/sys/backtrace.rs:166:18
  23:     0xaeeb874cf250 - std[9a20e64c0bb2485b]::rt::lang_start::<()>::{closure#0}
                               at /home/Work/pauth/rust/library/std/src/rt.rs:206:18
  24:     0xaeeb874f5f38 - call_once<(), (dyn core::ops::function::Fn<(), Output=i32> + core::marker::Sync + core::panic::unwind_safe::RefUnwindSafe)>
                               at /home/Work/pauth/rust/library/core/src/ops/function.rs:287:21
  25:     0xaeeb874f5f38 - do_call<&(dyn core::ops::function::Fn<(), Output=i32> + core::marker::Sync + core::panic::unwind_safe::RefUnwindSafe), i32>
                               at /home/Work/pauth/rust/library/std/src/panicking.rs:573:43
  26:     0xaeeb874f5f38 - catch_unwind<i32, &(dyn core::ops::function::Fn<(), Output=i32> + core::marker::Sync + core::panic::unwind_safe::RefUnwindSafe)>
                               at /home/Work/pauth/rust/library/std/src/panicking.rs:541:19
  27:     0xaeeb874f5f38 - catch_unwind<&(dyn core::ops::function::Fn<(), Output=i32> + core::marker::Sync + core::panic::unwind_safe::RefUnwindSafe), i32>
                               at /home/Work/pauth/rust/library/std/src/panic.rs:359:14
  28:     0xaeeb874f5f38 - {closure#0}
                               at /home/Work/pauth/rust/library/std/src/rt.rs:175:24
  29:     0xaeeb874f5f38 - do_call<std::rt::lang_start_internal::{closure_env#0}, isize>
                               at /home/Work/pauth/rust/library/std/src/panicking.rs:573:43
  30:     0xaeeb874f5f38 - catch_unwind<isize, std::rt::lang_start_internal::{closure_env#0}>
                               at /home/Work/pauth/rust/library/std/src/panicking.rs:541:19
  31:     0xaeeb874f5f38 - catch_unwind<std::rt::lang_start_internal::{closure_env#0}, isize>
                               at /home/Work/pauth/rust/library/std/src/panic.rs:359:14
  32:     0xaeeb874f5f38 - lang_start_internal
                               at /home/Work/pauth/rust/library/std/src/rt.rs:171:5
  33:     0xaeeb874cf370 - main

```

</details>

jchlanda:
> > Sorry, just getting back to this. My fear here is that backtrace-rs, as currently designed, isn't going to be a great fit for this. But just to be clear on what we can currently support, with the standard library's usage, we can get accurate backtrace right? And symbolization of that backtrace works?
> 
> Yes, both work on this target.
> 
> For example:
> 
> ```rust
> use std::panic;
> 
> fn main() {
>     let _ = panic::catch_unwind(|| {
>         panic!("BOOM");
>     });
> }
> ```
> 
> Compiled for `aarch64-unknown-linux-pauthtest` and run with `RUST_BACKTRACE=full`, the standard library produces a backtrace with correctly symbolized function names and source locations (shortened):
> 
> ```
> 16: ... - panic[... ]::main::{closure#0}
>     at .../local_tests/panic/panic.rs:5:9
> 17: ... - std[... ]::panicking::catch_unwind::do_call<...>
>     at .../library/std/src/panicking.rs:573:43
> 20: ... - panic[... ]::main
>     at .../local_tests/panic/panic.rs:4:13
> 33: ... - main
> ```
> 
> Full log

Does the above answer your question @ChrisDenton ?

tests/skip_inner_frames.rs
```diff
@@ -10,6 +10,12 @@ const ENABLED: bool = cfg!(all(
     target_os = "linux",
     // On ARM finding the enclosing function is simply returning the ip itself.
     not(target_arch = "arm"),
+    // On `aarch64-unknown-linux-pauthtest` `_Unwind_FindEnclosingFunction`
+    // cannot be used safely, because instruction pointers are not in a form
+    // suitable for authentication/resigning, so `symbol_address()` returns the
+    // raw IP instead. In this case the returned address may be inside the
+    // function body rather than at its entry.
+    not(target_env = "pauthtest"),
```
philipc:
```suggestion
    not(target_abi = "pauthtest"),
```

tests/smoke.rs
```diff
@@ -25,6 +25,13 @@ fn get_actual_fn_pointer(fp: *mut c_void) -> *mut c_void {
 }
 
 #[test]
+// This test relies on recovering precise symbol addresses from instruction
+// pointers. On `aarch64-unknown-linux-pauthtest`, we cannot safely use
+// `_Unwind_FindEnclosingFunction`, and instead fall back to using raw
+// instruction pointers. As a result, symbol resolution cannot reliably
+// recover a canonical function entry address, and `sym.addr()` will often be
+// `None`. Therefore this test is disabled.
+#[cfg_attr(target_env = "pauthtest", ignore)]
```
philipc:
```suggestion
#[cfg_attr(target_abi = "pauthtest", ignore)]
```

tests/skip_inner_frames.rs
jchlanda: Done, thank you.

src/backtrace/libunwind.rs
philipc: The changes and comments added in this PR read as though this is just the way it has to be for pauthtest.

However, it seems to me that `_Unwind_FindEnclosingFunction` is implemented wrong, and if it were fixed then the changes in this PR would not be needed (but I don't know much about libunwind and even less about pointer authentication).

If that's the case, then this PR is adding a workaround that is the same as what we do for Apple, so that seems fine, but maybe it would be better if the comments indicated it is only a workaround. For example, instead of long explanations in the tests, it would be better as short note that it is a workaround for `_Unwind_FindEnclosingFunction` with a link to an issue.

jchlanda: I've created an upstream libunwind issue here: https://github.com/llvm/llvm-project/issues/223698
I'm not certain that it is genuinely a libunwind bug though. There was somewhat similar discussion on an unrelated PR: https://github.com/llvm/llvm-project/pull/202772/changes#r3633237542

One could argue that the reasoning behind that comment:

> If you are fabricating a new sp that means you are taking on responsibility for constructing that new sp value correctly. Removing that requirement means that the stack pointer returned by __unw_get_reg would necessarily have lost its PAC signature.

Should be applied here as well.

jchlanda: @philipc / @ChrisDenton is there anything else you'd like me to do for this PR?
Thank you.

PR #756, opened 2026-04-23
Set no-op backtrace for all `wasm` family targets

arjunr2:
Particularly needed to support `wasm32-wali-linux-musl` appropriately

### DIFF
--- a/src/backtrace/mod.rs
+++ b/src/backtrace/mod.rs
@@ -174,6 +174,7 @@ cfg_if::cfg_if! {
         any(
             all(
                 unix,
+                not(target_family = "wasm"),
                 not(target_os = "emscripten"),
                 not(all(target_os = "ios", target_arch = "arm")),
             ),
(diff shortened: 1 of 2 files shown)


workingjubilee:
it is not clear to me that backtraces should be no-op for *all* wasm-family targets|

arjunr2:
Backtraces aren't supported on any wasm target besides those that use emscripten right now; I figured this should be reflected. Would you rather I gate it on just `not(all(target_arch = wasm32, target_os = linux))`?

workingjubilee:
My *understanding*, and it could be wrong, is the current environs is that one or two wasm targets are moving towards unwinding support, where it seems better to have noisier failures when they try to build and this code starts mattering than a silent failure to produce traces.

And that everyone else who isn't, is just building with `panic="abort"` so it doesn't matter to them?

arjunr2:
That makes sense, will dclose this for now until unwinding support is complete then. Thanks

PR #757, opened 2026-05-26
[AIX] Update backtrace to account for libc changes and fix XCOFF parsing

amy-kwan:
This patch synchronizes AIX backtrace support with newer libc updates and fix type correctness issues:

- Bump libc to 0.2.175 to pick up latest AIX-related fixes.
- Update libc::loadquery buffer argument to use *mut c_void instead of *mut c_char, matching the current libc API.
- Fix XCOFF section name comparison to handle Option<&[u8]>.

### DIFF
--- a/src/symbolize/gimli/libs_aix.rs
+++ b/src/symbolize/gimli/libs_aix.rs
@@ -30,7 +30,7 @@ pub(super) fn native_libraries() -> Vec<Library> {
         loop {
             if libc::loadquery(
                 libc::L_GETINFO,
-                buffer.as_mut_ptr().cast::<libc::c_char>(),
+                buffer.as_mut_ptr().cast::<libc::c_void>(),
                 (mem::size_of::<libc::ld_info>() * buffer.len()) as u32,
             ) != -1
             {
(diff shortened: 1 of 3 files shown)


workingjubilee:
Thanks! Will try to take a look soon.

workingjubilee:
rustfmt failure

amy-kwan:
Ah, I missed that. Thanks a lot! Just fixed them.

workingjubilee:
Yay, can merge this then.

src/symbolize/gimli/libs_aix.rs
```diff
@@ -30,7 +30,7 @@ pub(super) fn native_libraries() -> Vec<Library> {
         loop {
             if libc::loadquery(
                 libc::L_GETINFO,
-                buffer.as_mut_ptr().cast::<libc::c_char>(),
+                buffer.as_mut_ptr().cast::<libc::c_void>(),
```
workingjubilee, line 33: though is there a reason this isn't `core::ffi::c_void`?

should be fine either way I s'pose.

workingjubilee (approved):

src/symbolize/gimli/libs_aix.rs
amy-kwan: Oops, good point. Honestly, no strong reason, I used `libc::c_void` since it matched the `loadquery()` signature.

workingjubilee: That *should* just be a reexport from core so hopefully it doesn't matter (it's just nice to reduce indirection when viable).

Issue #758, opened 2026-06-24
Soundness: UAF/concurrent access on the MAPPINGS_CACHE


Manishearth:
> [!NOTE]
> This finding was identified during an agentic unsafe Rust code review performed by Gemini AI, followed by human review and verification.

### The Issue

In `backtrace` v0_3, the symbolication backend using `gimli` maintains a process-wide global cache of DWARF mappings (`static mut MAPPINGS_CACHE`) at [src/symbolize/gimli.rs#L17-L26](https://github.com/rust-lang/backtrace-rs/blob/d902726a1dcdc1e1c66f73d1162181b5423c645b/src/symbolize/gimli.rs#L17-L26), protected by a re-entrant mutex (`crate::lock::lock()`) in `backtrace::resolve` at

https://github.com/rust-lang/backtrace-rs/blob/d902726a1dcdc1e1c66f73d1162181b5423c645b/src/symbolize/mod.rs#L98-L105

When resolving symbols, `Cache::with_global` borrows `MAPPINGS_CACHE` mutably and obtains a DWARF context `cx` and stash `stash` from an active cached mapping. It then executes the caller's closure `cb(&Symbol)` while this mutable cache borrow remains live.

https://github.com/rust-lang/backtrace-rs/blob/d902726a1dcdc1e1c66f73d1162181b5423c645b/src/symbolize/gimli.rs#L130-L142

Because the global lock is re-entrant and `cb` runs synchronously inside the live borrow of `MAPPINGS_CACHE`, re-entrant calls in safe code trigger memory safety violations:
1. **Use-After-Free via `clear_symbol_cache()`**: If `cb` calls `backtrace::clear_symbol_cache()`, it creates a second concurrent `&mut Cache` reference to `MAPPINGS_CACHE` (Aliasing Violation) and clears `cache.mappings`. This drops all cached `Mapping`s and unmaps the DWARF file (`Mmap`), deallocating `cx` and `stash`. When `cb` returns, `resolve` continues iterating `frames.next()` on dangling pointers (Use-After-Free).
2. **Use-After-Free via LRU Cache Eviction**: The LRU cache has a fixed capacity of 4 (`MAPPINGS_CACHE_SIZE = 4`). If `cb` recursively resolves symbols across 4+ libraries, new mappings evict and drop the oldest mapping actively in use by the outer `resolve` frame (Use-After-Free).

<details><summary>Minimal Reproduction (Segfault)</summary>

```rust
fn main() {
[...]

PR #759, opened 2026-06-28
Store evicted cache entries to avoid dropping in-use mappings

ChrisDenton:
This is one approach to workaround #758.

A reference count is kept with each mapping so we know if it's still in use. If it is then we never drop it from the cache. This will work for recursion up to `MAPPINGS_CACHE_SIZE` (aka 4). After that it will fail to add new mappings. Which means it will effectively act as though loading the debug information failed (i.e. it won't resolve any symbols).

The disadvantage of this approach is it does potentially break programs that worked before. If you only use a symbol before you recurse (and never afterward) then you could previously do so infinitely*. I'm not sure that's a useful property though and tbh is a bit dubious soundness wise anyway (but probably works in practice).

Other possible approaches include making some kind of linked list of lru caches or redesigning the API so that `Symbol` can properly hold (and drop!) a reference to its cache entry.

### DIFF
--- a/src/symbolize/gimli.rs
+++ b/src/symbolize/gimli.rs
@@ -414,12 +414,19 @@ impl Cache {
             .next()
     }
 
-    fn mapping_for_lib<'a>(&'a mut self, lib: usize) -> Option<(&'a mut Context<'a>, &'a Stash)> {
+    fn mapping_for_lib<'a>(
+        &'a mut self,
+        lib: usize,
+    ) -> Option<(&'a mut Context<'a>, &'a Stash, Option<(usize, Mapping)>)> {
+        let mut evicted = None;
         let cache_idx = self.mappings.iter().position(|(lib_id, _)| *lib_id == lib);
 
         let cache_entry = if let Some(idx) = cache_idx {
             self.mappings.move_to_front(idx)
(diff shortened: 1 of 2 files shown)


workingjubilee:
So the immediate thought I had when the issue came up was that maybe we should just wholesale duplicate the cache with every call to `resolve` and then restore it after.

ChrisDenton:
Hm, duplicating everything would be a bit much I think? Debug info can be quite large. That being so, I think we need a reference count *somewhere* to avoid dropping memory that's in use. But it could be on the `Mapping` directly. Though I admit I may be missing something.

ChrisDenton:
Or, hm, maybe we can store the library id that is in use and restore them by pushing to the cache normally.

Something like this: https://github.com/ChrisDenton/backtrace-rs/commit/8bcfb5025cce8fba3b61f4357a7160aa2c0ae312?

workingjubilee:
I did mean something trying to preserve identity without like redoing mmaps, yes.

ChrisDenton:
Ok, I've taken another go at it. This time it stores then restores evicted cache entries.

(btw the macos runner is still intermittently playing up)

PR #763, opened 2026-07-02
use cfg_select instead of cfg_if

hkBst:
Now that cfg_select is stable since 1.95, we can replace cfg_if. One difference is that cfg_select needs one branch to match. Therefore an empty default branch is sometimes needed.

### DIFF
--- a/src/symbolize/gimli.rs
+++ b/src/symbolize/gimli.rs
@@ -25,14 +25,13 @@ mod mystd {
 #[cfg(not(backtrace_in_libstd))]
 extern crate std as mystd;
 
-cfg_if::cfg_if! {
-    if #[cfg(windows)] {
+cfg_select! {
+    windows => {
         #[path = "gimli/mmap_windows.rs"]
         mod mmap;
-    } else if #[cfg(target_vendor = "apple")] {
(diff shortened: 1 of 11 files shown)


src/symbolize/gimli.rs
```diff
@@ -233,22 +239,26 @@ cfg_if::cfg_if! {
...
-    } else if #[cfg(target_os = "aix")] {
+    }
+    target_os = "aix" => {
         mod libs_aix;
         use libs_aix::native_libraries;
-    } else {
-        // Everything else should doesn't know how to load native libraries.
+    }
+    _ => {
+        // Everything else doesn't know how to load native libraries.
```
hkBst, line 261: small typo fix here ^

hkBst:
I don't see how I broke (macos-latest, nightly)... maybe the object point release will help...

workingjubilee (changes requested):
Not your problem exactly but this should not land before we fix rust-lang/backtrace-rs#758 and cut a release as I do not want a further MSRV change to become a precondition for a safety-related update. Thus I am filing this comment so that in later triage it becomes obvious there is something to address (...outside this PR).

PR #766, opened 2026-08-01
Update `object` to 0.40.0

philipc:
The Mach-O change also means it handles file offsets larger than 32-bit, which can happen for large dSYM files.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -124,9 +124,9 @@ dependencies = [
 
 [[package]]
 name = "object"
-version = "0.39.1"
+version = "0.40.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


ChrisDenton:
I found the test failures confusing until I remembered that Cargo recently changed its directory layout so it's now looking for the DLL in the wrong place. In any case the failures are unrelated to this PR.

philipc:
Yes I had to do a [fix in addr2line](https://github.com/gimli-rs/addr2line/pull/372/changes/dbb8e987fcfadc45bbbf0b64ba585e04a197c031) to handle the directory layout, but I don't know how to fix it here yet.

djc:
Would be nice to get this merged.

ChrisDenton (approved):

PR #767, opened 2026-08-12
Fix accuracy test with new cargo build layout

philipc:
This fixes the test on nightly rust.

Also remove unneeded `dbginfo = "collapsible"` because `collapse_debuginfo` is stable for our MSRV. This allows the test to run on stable rust again.

Also use DLL_PREFIX and DLL_SUFFIX from std::env::consts while here.

Not an ideal fix, but the only environment variable I found was CARGO_CDYLIB_FILE_DYLIB_DEP_dylib_dep with a Cargo.toml change and -Z bindeps.


ChrisDenton (approved):
Thanks!

PR #768, opened 2026-08-13
Fix apt-get install failures in CI

philipc:


ChrisDenton:
Sorry, just getting back to this. I think I'm going to need this and #767 in the same PR so they can be merged successfully,

philipc:
Ok, I added this to #767.

PR #765, opened 2026-07-14
Upgrade addr2line, libloading, miniz_oxide

cuviper:
* `addr2line v0.25 -> v0.27`
  * `gimli v0.32 -> v0.34`
* `libloading v0.8 -> v0.9`
* `miniz_oxide v0.8 -> v0.9`

No code changes required.

PR #769, opened 2026-08-17
cleanup: bump addr2line dep to 0.27.1

durin42:
Backtrace was the last user of 0.25 at work, which is how we noticed.


durin42:
Other recent PRs are failing the same as mine, so I'm pretty confident my change didn't break the CI checks. :/

durin42:
Ah, just saw #765 - closing in favor of that.

PR #771, opened 2026-09-11
Deny unsafe operations in unsafe functions

### LINKED ISSUE #643: `deny(unsafe_op_in_unsafe_fn)` in backtrace-rs
Because backtrace-rs is included as a submodule and has platform-specific code, this lint is currently allowed inside std. I intend to remove this `allow` in std for this crate, and deny this lint inside backtrace.

There is platform-specific code in this crate. For most platforms, most of it is fairly limited, and limited to safe functions, so that there is no concern with this lint. There is, however, the divide between DWARF (or "Unix") and PDB (or "Windows"), on which pretty much everything else about this crate hinges. This should be... comparatively easy? to fix?

akashchamp:
Fixes #643

Enable `unsafe_op_in_unsafe_fn` at the crate root so unsafe operations inside
unsafe functions remain explicitly marked.

Tests:

- `cargo +1.90.0 fmt --all -- --check`
- `RUSTFLAGS=-Dwarnings cargo +1.90.0 test`
- `RUSTFLAGS=-Dwarnings cargo +1.90.0 test --features serialize-serde`
- `RUSTFLAGS=-Dwarnings cargo +1.90.0 test --features cpp_demangle`
- `RUSTFLAGS=-Dwarnings cargo +1.90.0 test --no-default-features`
- `RUSTFLAGS=-Dwarnings cargo +1.90.0 test --no-default-features --features std`
- `RUSTFLAGS=-Dunsafe_op_in_unsafe_fn cargo +1.90.0 check -p backtrace --all-features`


akashchamp:
The two red i686-msvc jobs die in the rustup install step on the new windows-2025 runner image ("toolchain 'stable-i686-msvc' may not be able to run on this system") before any code is built, and #755 fails the same way, so it looks like a CI image regression rather than something from this change. I checked i686-pc-windows-msvc, x86_64-pc-windows-msvc, i686-pc-windows-gnu and aarch64-pc-windows-msvc locally with RUSTFLAGS=-Dwarnings and they all compile clean with the lint denied.

